PR #8, opened 2015-04-06
adding basic account creation/deletion (no network support)

vtnerd:


src/account.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+    }
+    return sealed2_result && sealed1_result && signed_result;
+}
+
+#[test]
+fn generating_new_account() {
+    let username = "James".to_string();
+    let password = "Bond".as_bytes();
+    let pin = 500u32;
```
dirvine: cannot believe this is not 007 :-)

PR #11, opened 2015-04-13
Update README.md

vtnerd:
Need some help on expectation specifics here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,32 @@
     - [ ] test
   - [ ] Password Change
     - [ ] test
-- [ ] Implement Storage API:
-    - [ ] test
+- [ ] Implement Storage API (think about all of this as one unit when designing!)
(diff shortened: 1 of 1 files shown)


dirvine:
Thanks chap, I think you may need to merge with master and push again. 

It's probably worth adding all you can think of to take this lib right up to the c++ level (posix and RESTful interface). We can make most of this post 0.1.0 but it's good to collate it all whilst its fresh in our heads. Likely we will get a team to cut over very soon so it will help target resources as well.

vtnerd:
I started to write more for the Client-API (blob / container), but that started to get massive, so I left it open ended. Do you want me to continue for that too?

dirvine:
Yes please Lee it will give us the target we need. don't worry how big it gets. thx

vtnerd:
Updated.

dirvine:
Excellent, cheers.

PR #13, opened 2015-04-20
make maidsafe_client catch-up with the changes in routing

maqi:


maqi:
Hi David,
the failing CI is due to can not find the routing version 0.1.1.
could you make it another run manually once get routing 0.1.1 published?
Thank you.

dirvine:
Yes will do Qi, thanks

src/account.rs
```diff
@@ -236,15 +237,15 @@ fn slice_eq(left : &[u8], right : &[u8]) -> bool {
 #[allow(dead_code)]
 fn account_eq(left : &Account, right : &Account) -> bool {
     let data = "blah blah blah".to_string().into_bytes();
```
dirvine: poss better if this is a random string we create from a test perspective.

PR #30, opened 2015-05-25
MAID - 1080_Storage data types

krishnaIndia:
Implemented data tyoes fior NFS. Serialisation and unit test added for testing serialisation

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_client/30)


maidsafe-highfive:
r? @ustulation

src/nfs/types/container_id.rs
```diff
@@ -0,0 +1,184 @@
...
+    }};
+}
+
+impl ContainerId {
+    pub fn new() -> ContainerId {
+        let mut vec = Vec::with_capacity(ID_LEN);
+        for _ in (0..ID_LEN) {
+            vec.push(random::<u8>());
+        }
+        ContainerId(container_of_u8_to_array!(vec, ID_LEN).unwrap())
```
ustulation: panic! should be better than unwrap as it will point to line number in our source code

src/nfs/types/container_id.rs
```diff
@@ -0,0 +1,184 @@
...
+// bound by the terms of the MaidSafe Contributor Agreement, version 1.0.  This, along with the
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+const ID_LEN: usize = 512;
```
ustulation: length is 512 bits. However this variable is being used as bytes. So change to 64 maybe.

src/nfs/types/metadata.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::str;
+use std::fmt;
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct Metadata {
+    name: String,
+    size: i64,
+    created_time:  time::Tm,
+    modified_time: time::Tm,
+    user_metadata: String
```
ustulation: this should be a Vec<u8> .. It is serialised content of what the upper layers think about MetaData. Type should be unknown to Client library.

src/nfs/types/container_id.rs
```diff
@@ -0,0 +1,184 @@
...
+            Some(id_arr) => Ok(ContainerId(id_arr)),
+            None => Err(d.error("Bad NameType size"))
+        }
+    }
+}
+
+fn id_to_vec(id: &[u8;ID_LEN]) -> Vec<u8>{
+    let mut vec = Vec::with_capacity(ID_LEN);
+    for i in &id[..] {
+        vec.push(*i);
```
ustulation: would `id[..].into_iter().collect()` into a vector work? if so then this just reduces to a one liner and does not need a separate function.

src/nfs/types/file.rs
```diff
@@ -0,0 +1,57 @@
...
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+use super::metadata::Metadata;
+use self_encryption::datamap::DataMap;
+use std::fmt;
+
+#[derive(RustcEncodable, RustcDecodable, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct File(Metadata, DataMap);
```
ustulation: better to have struct with named fields than tuple-struct.

src/nfs/types/container_info.rs
```diff
@@ -0,0 +1,56 @@
...
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+use super::metadata::Metadata;
+use super::container_id::ContainerId;
+use std::fmt;
+
+#[derive(RustcEncodable, RustcDecodable, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct ContainerInfo(Metadata, ContainerId);
```
ustulation: better to have named field instead of tuple-struct

krishnaIndia:
Suggestions have been resolved. But could not build and test locally, self_encrytion is not building locally or on jenkins too. Do not merge,

krishnaIndia:
The self_encryption in crates.io is not building(0.1.1). Thus using self_encryption version(0.1.2) from github master branch. @ustulation the suggested refactoring is completed :)

PR #125, opened 2015-08-04
Channels approach

ustulation:


maidsafe-highfive:
r? @krishnaIndia

PR #128, opened 2015-08-09
reduce network-delay simulation to run test cases faster

ustulation:


maidsafe-highfive:
r? @krishnaIndia

PR #129, opened 2015-08-12
Changed Travis file to include testing on OS X and use QA scripts.

Fraser999:


maidsafe-highfive:
r? @krishnaIndia

PR #135, opened 2015-08-21
removing CborTagToEncode and reducing derivable parameter to Response…

ustulation:
…Getter


maidsafe-highfive:
r? @krishnaIndia

PR #136, opened 2015-08-21
Revert "removing CborTagToEncode and reducing derivable parameter to …

krishnaIndia:
Reverts maidsafe/safe_client#135


maidsafe-highfive:
r? @ustulation

PR #140, opened 2015-08-25
integrate with routing in crates.io

ustulation:


maidsafe-highfive:
r? @krishnaIndia

Cargo.toml
```diff
@@ -11,7 +11,7 @@ homepage      = "http://maidsafe.net"
 
 [dependencies]
 # ---- MaidSafe -----
-#routing         = "0.2.*"
+routing         = "0.3.4"
```
benjaminbollen: You don't trust routing the upcoming routing 0.3.5 ;-)

ustulation: doesn't `= "0.3.4"` actually mean `>= "0.3.4"` as opposed to `= "=0.3.4"` which would mean `exactly and only == 0.3.4` ? I'm confused myself .. @Fraser999 do you know ? ^^^

krishnaIndia: Can't we go with `0.3.*`?

benjaminbollen: oh no sorry @ustulation as most of the time, you are right; it does mean the minimal version

Fraser999: The rules vary for how far past `x.y.z` it considers.  For `0.3.4` it will consider anything >= 0.3.4 and < 0.4.0

ustulation: i think it's the same, atleast for `"0.3"`.. but seems better so will do that in a moment

Fraser999: Having said that, I think we have a policy of specifying `*` for all non-MaidSafe libs and `x.y.*` for all MaidSafe libs (not sure why).

PR #141, opened 2015-08-26
Remove allow failures from x86

rossmuir:


maidsafe-highfive:
r? @krishnaIndia

PR #144, opened 2015-08-26
allow x86 build to fail - This is to ignore `rust-crypto` build failures on Windows-32bit just now

ustulation:


maidsafe-highfive:
r? @krishnaIndia

ustulation:
This is to ignore `rust-crypto` build failures on Windows just now

PR #147, opened 2015-08-28
<1> credentials as utf-8 and non-restrictive .. restriction to be put…

ustulation:
… by the frontend apps

<2> allow build failures on x86 for libsodium


maidsafe-highfive:
r? @krishnaIndia

PR #156, opened 2015-09-08
Created framework for Linux safe_client installers - duplicated self_…

rossmuir:
…authentication example just now as POC until safe_client binary completed


maidsafe-highfive:
r? @krishnaIndia

PR #165, opened 2015-09-16
MAID-1300

ustulation:
MAID-1304


maidsafe-highfive:
r? @krishnaIndia

PR #166, opened 2015-09-17
<1> MAID-1299

ustulation:
<2> Adding version cache test case


maidsafe-highfive:
r? @krishnaIndia

PR #198, opened 2016-01-26
client send mpid message

maqi:


maidsafe-highfive:
r? @krishnaIndia

PR #200, opened 2016-02-02
Client messaging operations

maqi:


maidsafe-highfive:
r? @ustulation

ustulation:
hey @brian-js - the pr looks fine too me - there isn't much to review from my side (no test cases etc.) so If you are fine then i'll merge this.

brian-js:
If we need to add anything further we can do it in a separate PR, so I'd go for the merge thanks @ustulation.

PR #201, opened 2016-02-05
using mpid_messaging v0.1.0

maqi:


maidsafe-highfive:
r? @krishnaIndia

PR #202, opened 2016-02-05
messaging getters return ResponseGetter

maqi:


maidsafe-highfive:
r? @krishnaIndia

src/client/mod.rs
```diff
@@ -400,6 +400,28 @@ impl Client {
...
+
+        let mut msg_queue = unwrap_result!(self.message_queue.lock());
+        if msg_queue.local_cache_check(&msg_name) {
+            return Ok(ResponseGetter::new(None, self.message_queue.clone(), msg_request));
+        }
+
+        let request = MpidMessageWrapper::GetMessage(mpid_header.clone());
+        let serialised_request = unwrap_result!(serialise(&request));
+        let data = Data::PlainData(PlainData::new(msg_name.clone(), serialised_request));
+        try!(self.post(data, Some(Authority::ClientManager(mpid_header.sender().clone()))));
```
ustulation: Do posts go to ClientManagers or is it different for messaging ?

maqi: yes, it goes to ClientManager, not MaidManager but MpidManager, different persona

PR #249, opened 2016-04-27
chore/update: use immutable_data_operations when storing structured data

brian-js:
Update un/versioned SD to use immutable_data_operations and fix corresponding tests.
Use `<Arc<Mutex<..>>` consistently.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/249)


maidsafe-highfive:
r? @ustulation

PR #259, opened 2016-06-07
feat/account_info: add support for providing client account information

Fraser999:
This implements the client-side portion of the new `GetAccountInfo` Client request.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/259)


maidsafe-highfive:
r? @ustulation

PR #262, opened 2016-06-20
fix/dns: enable registered client to fetch public dns file

krishnaIndia:
Previously only registered owner and unregistered client could fetch the public dns file. A registered client could not fetch a public file that had been created by another owner. This fix aims to resolve that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/262)


maidsafe-highfive:
r? @ustulation

PR #264, opened 2016-06-23
Use Box::{from,into}_raw instead of transmute

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/264)

### DIFF
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -51,7 +51,7 @@ use std::sync::{Arc, Mutex, mpsc};
 use rustc_serialize::Decoder;
 use core::client::Client;
 use rustc_serialize::Decodable;
-use libc::{self, c_void, int32_t, c_char};
+use libc::{int32_t, c_char};
 use std::mem;
 use rustc_serialize::base64::FromBase64;
 use maidsafe_utilities::serialisation::{serialise, deserialise};
@@ -107,7 +107,8 @@ pub trait Action {
     fn execute(&mut self, params: ParameterPacket) -> ResponseType;
 }
 
-struct FfiHandle {
+/// A handle, passed through the FFI.
+pub struct FfiHandle {
     client: Arc<Mutex<Client>>,
     network_thread_terminator: Option<Sender<NetworkEvent>>,
     raii_joiner: Option<RaiiThreadJoiner>,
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @krishnaIndia

ustulation:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/ffi/mod.rs, line 137 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_core/264#-KKwly811LWJrGPrMeAU:-KKwly82o5-qM4rk1WxC:257005714) ([raw file](https://github.com/maidsafe/safe_core/blob/7e3f819a426211f90e1dd4c2015bbfa719f8035c/src/ffi/mod.rs#L137)):*
> ```Rust
> #[no_mangle]
> #[allow(unsafe_code)]
> pub extern "C" fn create_unregistered_client(ffi_handle: *mut *const c_void) -> int32_t {
> ```

maybe a touch up here too to make it `*mut *mut` instead of `*mut *const`

---

*[src/ffi/mod.rs, line 288 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_core/264#-KKwmd_FzHVay-ZxdSS4:-KKwmd_GNh-bVt_DyZbr:-693319286) ([raw file](https://github.com/maidsafe/safe_core/blob/7e3f819a426211f90e1dd4c2015bbfa719f8035c/src/ffi/mod.rs#L288)):*
> ```Rust
> pub extern "C" fn drop_client(client_handle: *mut c_void) {
>     unsafe {
>         let client_handle = client_handle as *mut Arc<Mutex<Client>>;
> ```

This seems wrong - need to cast it to `FfiHandle` and drop that instead because we do not send `Arc<Mutex<Client>>` across FFI so we are only supposed to get `FfiHandle` back. This was wrong in the previous code too. But do scan the file to see if what i said makes sense.

---

*[src/ffi/mod.rs, line 364 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_core/264#-KKwnC1rfFV-ga9AyzJD:-KKwnC1rfFV-ga9AyzJE:2018171809) ([raw file](https://github.com/maidsafe/safe_core/blob/7e3f819a426211f90e1dd4c2015bbfa719f8035c/src/ffi/mod.rs#L364)):*
> ```Rust
> #[allow(unsafe_code)]
> /// Drop the null pointer returned as error from the execute_for_content fn
> pub fn drop_null_ptr(ptr: *mut u8) {
> ```

This one does not belong to the change set you did, but I have no idea why this function was ever coded. Deleting a null pointer is a waste - it's a NOP in c & c++. So while at it could you just scan and remove this too ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/264)*

ustulation:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/264)*

PR #269, opened 2016-06-27
fix:ffi_api/fix nfs move api issue

krishnaIndia:
nfs move directory api error fixed and test case updated for move and copy.
Additional ffi api added for fetching file metadata from dns and nfs in the ffi interface.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/269)


maidsafe-highfive:
r? @ustulation

PR #279, opened 2016-06-28
fix/ffi_api: resolve move dir api fix

krishnaIndia:
Fixed issue with move directry api in ffi nfs module
Added new api for fetching file metadata from nfs and dns

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/279)


maidsafe-highfive:
r? @ustulation

PR #280, opened 2016-06-28
perf/recovery: avoid recovery on low balance

ustulation:
Low balace cannot be recovered so bail out of recovery process if so instead of wasting cycles and bandwidth

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/280)


maidsafe-highfive:
r? @krishnaIndia

PR #296, opened 2016-07-05
fix/unwinding: correct order of destruction

ustulation:
Routing needs to be destroyed first for mpsc::Sender given to it to be destroyed which will trigger the exit of thread running MessageQueue and thus cause raii_joiner to exit. This order is respected in Client struct but in the stack frame of constructors (create/login functions) it was not. This fix resolves that as things that come later in the frame will be destroyed first so routing needs to come after MessageQueue because it needs to be destroyed before MessageQueue for raii_joiner to exit.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/296)

### DIFF
--- a/src/core/client/mod.rs
+++ b/src/core/client/mod.rs
@@ -78,9 +78,9 @@ impl Client {
         let (routing_sender, routing_receiver) = mpsc::channel();
         let (network_event_sender, network_event_receiver) = mpsc::channel();
 
-        let routing = try!(Routing::new(routing_sender, None));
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @krishnaIndia

Viv-Rajkumar:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/296)*

PR #300, opened 2016-07-11
Upgrade to routing 0.24.0

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/300)


maidsafe-highfive:
r? @ustulation

PR #312, opened 2016-08-05
fix/ffi_move: fetch copy or moved directory

krishnaIndia:
Create the directory moved or copied in the network. Previously the metadata was updated, and when the directory was fetched it returned error.
Similarly in copy file the copied file must generate a new id, instead of using the same id from the source

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/312)


maidsafe-highfive:
r? @ustulation

ustulation:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/312)*

PR #328, opened 2016-09-13
refactor/ffi: refactor and improve ObjectCache

madadam:
- use lazy_static.
- expose `object_cache_reset` FFI function.
- `object_cache()` now calls `lock()` internally and returns `LockResult<MutexGuard<ObjectCache>>`.
- use the convenience functions (get_data_id(), etc...) in more places.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/328)


maidsafe-highfive:
r? @ustulation

ustulation:
*[src/ffi/low_level_api/immut_data.rs, line 132 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXtO1wC47G9XOWahfP:-KRXtO1wC47G9XOWahfQ:brvn1m9) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/immut_data.rs#L132)):*
> ```Rust
> 
>         let mut obj_cache = unwrap!(object_cache());
>         let handle = obj_cache.insert_data_id(data_id);
> ```

Since we don't need the mutex guard to persist it should be better practice to combine this in single line with above - it is less dangerous to have MutexGuard rendered out of scope in the same expression than keep it lingering around (else refactors or otherwise might lead to subtle deadlocks).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

ustulation:
*[src/ffi/low_level_api/immut_data.rs, line 181 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXtrizIp3InIKVIvMG:-KRXtrizIp3InIKVIvMH:b9jtd7d) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/immut_data.rs#L181)):*
> ```Rust
>         let mut obj_cache = unwrap!(object_cache());
>         let handle = obj_cache.new_handle();
>         if obj_cache.se_reader.insert(handle, se_wrapper).is_some() {
> ```

Have convenience functions for these as well (in addition to data_id/appendable data etc.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

ustulation:
Review status: 0 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/ffi/low_level_api/object_cache.rs, line 38 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXuCl9OFtbFC5zxLue:-KRXuCl9OFtbFC5zxLuf:b-5t33yo) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/object_cache.rs#L38)):*
> ```Rust
> 
> pub fn object_cache() -> LockResult<MutexGuard<'static, ObjectCache>> {
>     OBJECT_CACHE.lock()
> ```

Might be a good idea to `unwrap!()` right at source here ? That way (apart from typing it again everywhere) if we change this `unwarp` to `try` (which i think we will some day) it will be only at one place (though of-course the return type would be different but the main semantic change will be only at one place).

---

*[src/ffi/low_level_api/object_cache.rs, line 147 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXujlCq2sVGJwRnM67:-KRXujlDsb_vUio4Chgt:b-i72pbc) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/object_cache.rs#L147)):*
> ```Rust
> /// Reset the object cache (drop all objects stored in it). This will invalidate
> /// all currently held object handles.
> pub extern "C" fn object_cache_reset() {
> ```

Should this function be taken to `misc` module instead ? It looks pretty hidden here and since this module is not otherwise an ffi module we needed to expose this one function in `low_level_api::mod.rs`. Instead if we take it to `misc` then it becomes easier for ppl to track (we can tell the users all ffi functions are in these `a,b,c..`  modules)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

madadam:
Review status: 0 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/ffi/low_level_api/object_cache.rs, line 38 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXuCl9OFtbFC5zxLue:-KRXwu5w8p7luUBiejIL:bsnfw9h) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/object_cache.rs#L38)):*
<details><summary><i>Previously, ustulation (Spandan Sharma) wrote…</i></summary>
> Might be a good idea to `unwrap!()` right at source here ? That way (apart from typing it again everywhere) if we change this `unwarp` to `try` (which i think we will some day) it will be only at one place (though of-course the return type would be different but the main semantic change will be only at one place).
</details>
I was considering it. But in the end I decided to leave the decision whether to `unwrap` or do something else to the caller. I don't, however, feel strongly about it. If you do, I'll change it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

ustulation:
*[src/ffi/low_level_api/object_cache.rs, line 38 at r1](https://reviewable.io:443/reviews/maidsafe/safe_core/328#-KRXuCl9OFtbFC5zxLue:-KRYAlliaKO1vV_TbcpX:b-k6myb3) ([raw file](https://github.com/maidsafe/safe_core/blob/ab654655bf8d896c9465a6cf0b67be72f85affa5/src/ffi/low_level_api/object_cache.rs#L38)):*
<details><summary><i>Previously, madadam (Adam Cigánek) wrote…</i></summary>
> I was considering it. But in the end I decided to leave the decision whether to `unwrap` or do something else to the caller. I don't, however, feel strongly about it. If you do, I'll change it.
</details>
Ok let's leave it then

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

ustulation:
Reviewed 4 of 10 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/328)*

PR #330, opened 2016-09-13
refactor/mock-routing: refactor mock routing

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_core/330)

### DIFF
--- a/src/core/client/non_networking_test_framework/mod.rs
+++ b/src/core/client/non_networking_test_framework/mod.rs
@@ -17,10 +17,9 @@
 
 #![cfg_attr(feature="clippy", allow(map_entry))]  // TODO: Look to enable this lint check.
 
+mod storage;
 
-use bincode::SizeLimit;
-use maidsafe_utilities::serialisation::{deserialise, deserialise_with_limit, serialise,
-                                        serialise_with_limit};
+use maidsafe_utilities::serialisation::serialise;
 use maidsafe_utilities::thread;
 use rand;
 use routing::{AppendWrapper, Authority, Data, DataIdentifier, Event, FullId, InterfaceError,
@@ -30,17 +29,13 @@ use routing::client_errors::{GetError, MutationError};
 use rust_sodium::crypto::hash::sha256;
 use rust_sodium::crypto::sign;
 use rustc_serialize::Encodable;
+use self::storage::{ClientAccount, Storage};
 use std;
 use std::cell::Cell;
-use std::collections::HashMap;
-use std::env;
-use std::fs::File;
-use std::io::{Read, Write};
-use std::sync::{Arc, Mutex, ONCE_INIT, Once};
+use std::sync::Mutex;
 use std::sync::mpsc::Sender;
 use std::time::Duration;
 
-const STORAGE_FILE_NAME: &'static str = "VaultStorageSimulation";
 const NETWORK_CONNECT_DELAY_SIMULATION_THREAD: &'static str = "NetworkConnectDelaySimulation";
 
 // Activating these (ie., non-zero values) will require an update to all test cases. Once activated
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @krishnaIndia

ustulation:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core/client/non_networking_test_framework/mod.rs, line 451 at r2](https://reviewable.io:443/reviews/maidsafe/safe_core/330#-KRhXUtPdYCZFIAPh5m4:-KRhXUtPdYCZFIAPh5m5:b-5u4m2y) ([raw file](https://github.com/maidsafe/safe_core/blob/e79a25c8098bae1d99ac89189fdd78581c966f60/src/core/client/non_networking_test_framework/mod.rs#L451)):*
> ```Rust
>         let err = if storage.contains_data(&data_name) {
>             if self.network_limits_reached() {
>                 info!("Mock POST: {:?} {:?} [0]", data_id, msg_id);
> ```

This should say `APPEND` instead of `POST`

---

*[src/core/client/non_networking_test_framework/storage.rs, line 51 at r2](https://reviewable.io:443/reviews/maidsafe/safe_core/330#-KRhXg5FsLT1zrdqK5qM:-KRhXg5FsLT1zrdqK5qN:b-vkvfzy) ([raw file](https://github.com/maidsafe/safe_core/blob/e79a25c8098bae1d99ac89189fdd78581c966f60/src/core/client/non_networking_test_framework/storage.rs#L51)):*
> ```Rust
>     pub fn get_data(&self, name: &XorName) -> Result<Data, StorageError> {
>         match self.data_store.get(name) {
>             Some(data) => deserialise(data).map_err(|_| StorageError::SerialisationError),
> ```

Maybe use `map_err(StorageError::SerialisationError)` and later in `From<StorageError>` do something like
```
e @ StorageError::SerialisationError(_) => {
                MutationError::NetworkOther(format!("{:?}", e))
            }
```
That will give more info to debug with should it be ever triggered

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/330)*

ustulation:
*[src/core/client/non_networking_test_framework/mod.rs, line 51 at r2](https://reviewable.io:443/reviews/maidsafe/safe_core/330#-KRh_27zz8gEW8CZNegM:-KRh_27zz8gEW8CZNegN:b-fpzh8v) ([raw file](https://github.com/maidsafe/safe_core/blob/e79a25c8098bae1d99ac89189fdd78581c966f60/src/core/client/non_networking_test_framework/mod.rs#L51)):*
> ```Rust
> const SIMULATED_NETWORK_DELAY_PUTS_DELETS_MS: u64 = 2 * SIMULATED_NETWORK_DELAY_GETS_POSTS_MS;
> 
> lazy_static! {
> ```

I am thinking of only selectively using this for non-tests. So something along the lines:
```
#[not(cfg(test))]
lazy_static! {
```
and
```
pub struct RoutingMock {
#[cfg(test)]
    storage: Storage,
    sender: Sender<Event>,
    client_auth: Authority,
    max_ops_countdown: Option<Cell<u64>>,
}

#[cfg(test)]
fn storage(&self) -> &mut Storage { &mut self.storage }
#[not(cfg(test))]
fn storage(&self) -> &mut Storage { &mut unwrap!(*STORAGE) }
```

This will make the test cases faster and less memory intensive ? (As they won't be holding permanent memories and will release with the end of test-case or end of RoutingMock whichever is first).

What do you think @madadam ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/330)*

madadam:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/core/client/non_networking_test_framework/mod.rs, line 51 at r2](https://reviewable.io:443/reviews/maidsafe/safe_core/330#-KRh_27zz8gEW8CZNegM:-KRhlA3XbTJntf2mLPQq:b-is9fzf) ([raw file](https://github.com/maidsafe/safe_core/blob/e79a25c8098bae1d99ac89189fdd78581c966f60/src/core/client/non_networking_test_framework/mod.rs#L51)):*
<details><summary><i>Previously, ustulation (Spandan Sharma) wrote…</i></summary>
> I am thinking of only selectively using this for non-tests. So something along the lines:
> ```
> #[not(cfg(test))]
> lazy_static! {
> ```
> and
> ```
> pub struct RoutingMock {
> #[cfg(test)]
>     storage: Storage,
>     sender: Sender<Event>,
>     client_auth: Authority,
>     max_ops_countdown: Option<Cell<u64>>,
> }
> 
> #[cfg(test)]
> fn storage(&self) -> &mut Storage { &mut self.storage }
> #[not(cfg(test))]
> fn storage(&self) -> &mut Storage { &mut unwrap!(*STORAGE) }
> ```
> 
> This will make the test cases faster and less memory intensive ? (As they won't be holding permanent memories and will release with the end of test-case or end of RoutingMock whichever is first).
> 
> What do you think @madadam ?
</details>
That would break tests which create more than one session (and thus routing client) as they wouldn't be able to access each others data. Besides, the memory usage and speed benefits are questionable in my opinion. I'd say, it isn't worth the increased code complexity. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/330)*

ustulation:
*[src/core/client/non_networking_test_framework/mod.rs, line 51 at r2](https://reviewable.io:443/reviews/maidsafe/safe_core/330#-KRh_27zz8gEW8CZNegM:-KRhmeoGKx9lAbOfB5Rr:b-daxtu) ([raw file](https://github.com/maidsafe/safe_core/blob/e79a25c8098bae1d99ac89189fdd78581c966f60/src/core/client/non_networking_test_framework/mod.rs#L51)):*
<details><summary><i>Previously, madadam (Adam Cigánek) wrote…</i></summary>
> That would break tests which create more than one session (and thus routing client) as they wouldn't be able to access each others data. Besides, the memory usage and speed benefits are questionable in my opinion. I'd say, it isn't worth the increased code complexity. 
</details>
@madadam : Ah you are right about multisession tests - like create-account followed by login etc.

But speed benefits i still believe would be there - tests are run in parallel by cargo and a global mutex _will_ slow things down (they are held almost for entire of PUT/POST etc functions). With over 100 tests for `safe_core` it might soon become a bothersome waiting for all to pass on local machine as well as CI (it already take around 20-25 min per windows CI (thus over 40 min in total)) and that is in release.

Can we not design something better here ? Like say `RoutingMock` constructor takes extra storage parameter in case of tests.  Since there are only 3 entry points to the api: `create_account`, `login` and `create_unregistered_account`, those could take extra parameter for tests to pass it to RoutingMock. For multisession tests then we can have:
```
let storage: Arc<Mutex<Storage>>; // And pass storage.clone() to above. So same storage for the duration of the test
```
And other tests are not affected by any global mutex so run as parallel as cargo makes them.

Maybe roughly record the difference in timings on running full test suite in release with current PR and something like above mentioned. Then if the difference is not significant we can drop the idea.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/330)*

ustulation:
Reviewed 1 of 2 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_core/330)*

PR #339, opened 2016-09-21
feat/ffi: add low level API to get SD/AD versions

nbaksalyar:
In addition, this patch optimises versions storage in versioned StructuredData. Now it stores the current version and total versions count along with a XorName of the versions list.

### DIFF
--- a/src/core/structured_data_operations/versioned.rs
+++ b/src/core/structured_data_operations/versioned.rs
@@ -24,6 +24,17 @@ use routing::{Data, DataIdentifier, ImmutableData, StructuredData, XorName};
 use rust_sodium::crypto::sign;
 use std::sync::{Arc, Mutex};
 
+/// All fields updated whenever a version is appended/removed
+#[derive(RustcEncodable, RustcDecodable, Debug)]
+struct VersionsInfo {
+    /// Contains an ID for Vec<XorName> of versions
+    ptr_to_versions: XorName,
+    /// Total number of available versions
+    total_versions: u64,
+    /// Contains an ID for the current version
+    ptr_to_current_version: XorName,
+}
+
 /// Create the StructuredData to manage versioned data.
 #[cfg_attr(feature="clippy", allow(too_many_arguments))]
 pub fn create(client: Arc<Mutex<Client>>,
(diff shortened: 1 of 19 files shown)


maidsafe-highfive:
r? @ustulation

src/ffi/low_level_api/appendable_data.rs
```diff
@@ -666,6 +666,23 @@ pub unsafe extern "C" fn appendable_data_append(app: *const App,
     })
 }
 
+/// Get the current version of AppendableData by its handle
+#[no_mangle]
+pub extern "C" fn appendable_data_version(handle: AppendableDataHandle,
```
ustulation: This is an `unsafe` function as opposed to `unsafe` block used within implying it's a safe function. This is because we are accepting a raw pointer in the parameter. We assume it is non-null and does infact point to memory which can be interpreted as `u64`. We cannot verify these (could have been a pointer to a completely different memory layout) and will be U.B if caller wasn't careful. Thus function itself being `unsafe`.

src/ffi/low_level_api/struct_data.rs
```diff
@@ -456,6 +447,17 @@ fn struct_data_delete_impl(client: Arc<Mutex<Client>>,
     Ok(new_sd)
 }
 
+/// Get the current version of StructuredData by its handle
+#[no_mangle]
+pub extern "C" fn struct_data_version(handle: StructDataHandle, o_version: *mut u64) -> i32 {
```
ustulation: This function should be marked `unsafe` instead of using `unsafe` blocks inside, as explained in comment on `appendable_data_version(...)`.

src/core/structured_data_operations/versioned.rs
```diff
@@ -24,6 +24,17 @@ use routing::{Data, DataIdentifier, ImmutableData, StructuredData, XorName};
...
+    ptr_to_versions: XorName,
+    /// Total number of available versions
+    total_versions: u64,
+    /// Contains an ID for the current version
+    ptr_to_current_version: XorName,
+}
+
 /// Create the StructuredData to manage versioned data.
 #[cfg_attr(feature="clippy", allow(too_many_arguments))]
 pub fn create(client: Arc<Mutex<Client>>,
```
ustulation: In this and in other places (unversioned_sd.rs/dir_helper.rs etc.) can we also get rid of the type-tag parameter as we now know that it has to be of a fixed value to be called versioned or unversioned - so scan the create api's in nfs/dns to make them take something like `is_versioned` as a parameter and convert that to the fixed tags we reserved.

PR #346, opened 2016-10-03
fix/ffi: add missing no_mangle on an FFI function

nbaksalyar:
misc_deserialise_sign_key name is missing #[no_mangle]


maidsafe-highfive:
r? @krishnaIndia

PR #347, opened 2016-10-03
Version change to 0.21.2

ustulation:
API for getting number of filter keys
API for getting nth filter key
Remove deprecated lints
Rustfmt changes


maidsafe-highfive:
r? @krishnaIndia

PR #348, opened 2016-10-06
Conform to latest utilities and routing

ustulation:
Re-run CI once routing 0.27.0 is published


maidsafe-highfive:
r? @krishnaIndia

ustulation:
Routing is not going to be published to crates just now. Needs to be re-raised when routing gets a version publish.

PR #350, opened 2016-10-11
Foundation work  for async safe_core - Futures paradigm

ustulation:
1. Using tokio-core as the reactor event loop
2. Using futures for async communication.


maidsafe-highfive:
r? @krishnaIndia

PR #351, opened 2016-10-11
refactor/core: make ok! and err! macros use boxed values

nbaksalyar:
With the added bonus of formatting altered by rustfmt

### DIFF
--- a/src/core/futures.rs
+++ b/src/core/futures.rs
@@ -43,7 +43,7 @@ macro_rules! ok {
 /// construct the return type equivalent of `Result::Err` in futures paradigm.
 macro_rules! err {
     ($elt:expr) => {
-        futures::done(Err($elt)).into_box()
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @ustulation

src/core/futures.rs
```diff
@@ -35,27 +35,27 @@ macro_rules! fry {
...
+        futures::done(Ok($elt)).into_box()
     }
 }
 
 /// This is the equivalent of `Result::Ok()` adapted to deal with futures. This should be used to
 /// construct the return type equivalent of `Result::Err` in futures paradigm.
 macro_rules! err {
     ($elt:expr) => {
-        futures::done(Err($elt))
+        futures::done(Err($elt)).into_box()
```
ustulation: Will it be better if it were `futures::done(Err(From::from($elt))).into_box()`

PR #352, opened 2016-10-11
fix/client: persist user and config root modifications

madadam:


maidsafe-highfive:
r? @krishnaIndia

ustulation:
addressed in https://github.com/maidsafe/safe_core/pull/354

PR #354, opened 2016-10-12
Client handle

madadam:


maidsafe-highfive:
r? @ustulation

PR #355, opened 2016-10-12
refactor/core: implement immutable data operations

madadam:


maidsafe-highfive:
r? @ustulation

src/core/immutable_data.rs
```diff
@@ -0,0 +1,287 @@
...
+enum DataTypeEncoding {
+    Serialised(Vec<u8>),
+    DataMap(DataMap),
+}
+
+/// Create and obtain immutable data out of the given raw bytes. The API will encrypt the right
+/// content if the keys are provided and will ensure the maximum immutable data chunk size is
+/// respected.
+pub fn create(client: &Client, value: Vec<u8>,
+              encryption_keys: Option<(PublicKey, SecretKey, Nonce)>)
```
ustulation: These need to be symmetric encryption - so maybe should only accept `secretbox::Key` and generate nonce internally ?

src/core/immutable_data.rs
```diff
@@ -0,0 +1,289 @@
...
+    match fry!(deserialise(data.value())) {
+        DataTypeEncoding::Serialised(value) => ok!(value),
+        DataTypeEncoding::DataMap(data_map) => {
+            let storage = SelfEncryptionStorage::new(client.clone());
+            let self_encryptor = fry!(SelfEncryptor::new(storage, data_map));
+            let length = self_encryptor.len();
+            self_encryptor.read(0, length)
+                .map_err(From::from)
+                .and_then(|serialised_data| {
+                    Ok(try!(deserialise(&serialised_data)))
```
ustulation: same as previous comment on multiple combinators.

src/core/immutable_data.rs
```diff
@@ -0,0 +1,289 @@
...
+    let data = ImmutableData::new(value);
+    let serialised_data = fry!(serialise(&data));
+
+    if serialised_data.len() > MAX_IMMUTABLE_DATA_SIZE_IN_BYTES {
+        let storage = SelfEncryptionStorage::new(client.clone());
+        let self_encryptor = fry!(SelfEncryptor::new(storage, DataMap::None));
+        self_encryptor.write(&serialised_data, 0).and_then(move |_| {
+            self_encryptor.close()
+        }).map_err(From::from).and_then(|(data_map, _)| {
+            Ok(try!(serialise(&DataTypeEncoding::DataMap(data_map))))
```
ustulation: why require two `and_then`'s here - serialise is not something that would finish immediately as other normal functions. So this and the next `and_then` can be combined, no ?

ustulation: why require two and_then's here - serialise is not something that would finish immediately as other normal functions. So this and the next and_then can be combined, no ?

src/core/immutable_data.rs
```diff
@@ -0,0 +1,289 @@
...
+        DataTypeEncoding::DataMap(data_map) => {
+            let storage = SelfEncryptionStorage::new(client.clone());
+            let self_encryptor = fry!(SelfEncryptor::new(storage, data_map));
+            let length = self_encryptor.len();
+            self_encryptor.read(0, length)
+                .map_err(From::from)
+                .and_then(|serialised_data| {
+                    Ok(try!(deserialise(&serialised_data)))
+                })
+                .and_then(move |data| unpack(client, data))
```
ustulation: same as previous comments on extra combinators - pls check rest of the code too if you agree and think there might be more

src/core/immutable_data.rs
madadam: Done.

src/core/immutable_data.rs
madadam: :+1:

PR #356, opened 2016-10-12
refactor/nfs: update dir_helper in NFS to use new API

nbaksalyar:
Implement future-based NFS directory helper with symmetric encryption


maidsafe-highfive:
r? @ustulation

src/nfs/helper/dir_helper.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    save(client.clone(),
+         &parent,
+         &parent_locator,
+         parent_key.as_ref())
+        .and_then(move |structured_data| {
+            let client = client2.clone();
+            client2.put_recover(Data::Structured(structured_data), None)
+                .map_err(NfsError::from)
+                .and_then(move |_| {
```
ustulation: i think not nesting stuff like callback was one of the main advs of using futures. So something like:
```
future
     .and_then(|_| stuff())
     .and_then(|_| stuff())
     .and_then(|_| stuff());
```
is better than
```
future
     .and_then(|_| {
           stuff().and_then(|_|) {
                 stuff().and_then(|_|) {
                         stuff()
                 }
           }
    }
```
So the way to remember this is could be to make sure the level of indentation is exactly one in all places. Those which require more should be exceptional cases (to be dealt with case by case basis).

Pls scan if there are more such places.

src/nfs/helper/dir_helper.rs
```diff
@@ -0,0 +1,544 @@
...
+            client2.put_recover(Data::Structured(structured_data), None)
+                .map_err(NfsError::from)
+                .map(move |_| data_id)
+        })
+        .into_box()
+}
+
+/// Adds a sub directory to a parent.
+/// Parent directory is updated and the sub directory metadata returned as a result.
+pub fn add(client: Client,
```
ustulation: Maybe call it `add_sub_dir` as it seems confusing with `create` above

PR #358, opened 2016-10-17
Client recover

madadam:


maidsafe-highfive:
r? @krishnaIndia

src/core/client/mod.rs
```diff
@@ -378,6 +432,28 @@ impl Client {
...
+    /// A version of `delete` that returns success if the data was already not present on
+    /// the network.
+    pub fn delete_recover(&self, data: Data, dst: Option<Authority>) -> Box<CoreFuture<()>> {
+        trace!("DELETE with recovery for {:?}", data);
+
+        self.delete(data, dst)
+            .then(|result| {
+                match result {
+                    Ok(()) |
+                    Err(CoreError::MutationFailure { reason: MutationError::NoSuchData, .. }) => {
```
ustulation: This will now be `InvalidOperation` as delete no longer actually deletes stuff. It
1. blanks out owners + previous owners + signature + data fields
2. increments the version (to the same version as in the Delete request)

Now people can do a POST with version bump to reclaim that SD/AD.

Can you pls change this behavior in mock routing too and update any failing test cases ?

PR #360, opened 2016-10-18
StructureData deletion

ustulation:
Change in paradigm incorporated in code and test. Pls see individual commits.


maidsafe-highfive:
r? @krishnaIndia

PR #362, opened 2016-10-19
FFI basics and backports for SD deletion and size validation

nbaksalyar:


maidsafe-highfive:
r? @ustulation

src/ffi/low_level_api/struct_data.rs
```diff
@@ -657,8 +665,9 @@ mod tests {
 
             // Delete
             assert_eq!(struct_data_delete(&app, sd_h), 0);
-            // -18 is CoreError::GetFailure { reason: GetError::NoSuchData }
-            assert_eq!(struct_data_fetch(&app, data_id_h, &mut sd_h), -18);
+            // -22 is CoreError::MutationFailure { reason: MutationError::NoSuchData }
```
ustulation: These have all changed - pls check with `master`

src/ffi/low_level_api/struct_data.rs
```diff
@@ -725,8 +736,9 @@ mod tests {
...
```
ustulation: These have all changed - pls check with `master`

src/ffi/app.rs
```diff
@@ -21,143 +21,134 @@
...
-
-    /// Get SAFEdrive directory key.
-    pub fn get_safe_drive_dir_key(&self) -> Option<DirectoryKey> {
-        *unwrap!(self.session.lock()).get_safe_drive_dir_key()
-    }
-
-    /// Has this app access to the SAFEdrive?
-    pub fn has_safe_drive_access(&self) -> bool {
-        self.safe_drive_access
+    pub fn sym_key(&self) -> Result<secretbox::Key, FfiError> {
```
ustulation: Never clone a secret key (whether asym or sym) - that is why these keys don't implement `Copy` unlike their public counterpart (if asym). Cloning secret keys around in memory is seen as a security issue. So if cloning is absolutely required (for borrow checkers and stuff) then do it at the user/caller site, but never at the interface/api level itself. That way we can later analyse the security aspect and maybe change it to `Rc<SecretKey>` etc.

src/ffi/app.rs
```diff
@@ -21,143 +21,134 @@
...
-    pub fn get_root_dir_key(&self, is_shared: bool) -> Result<DirectoryKey, FfiError> {
-        if is_shared {
-            if !self.has_safe_drive_access() {
-                return Err(FfiError::PermissionDenied);
-            }
-
-            self.get_safe_drive_dir_key()
-                .ok_or(FfiError::from("Safe Drive directory key is not present"))
+    /// Get asymmetric encryption key for the app
+    pub fn asym_enc_keys(&self) -> Result<(box_::PublicKey, box_::SecretKey), FfiError> {
```
ustulation: Same as prev. comment about not cloning SecretKeys

src/ffi/launcher_config.rs
```diff
@@ -23,40 +23,28 @@ use ffi::config::{LAUNCHER_GLOBAL_CONFIG_FILE_NAME, LAUNCHER_GLOBAL_DIRECTORY_NA
...
-
-#[derive(RustcEncodable, RustcDecodable, Debug, Clone)]
-pub struct AppInfo {
-    pub app_root_dir: Dir,
-    pub asym_keys: (box_::PublicKey, box_::SecretKey),
-    pub sym_key: secretbox::Key,
-}
-
-pub fn app_info(client: Client,
+pub fn app_info(client: &Client,
```
ustulation: Maybe just rename to `app` instead of `app_info` if we are doing away with the notion of `AppInfo` ?

src/ffi/session.rs
```diff
@@ -17,143 +17,208 @@
...
 }
 
 impl Drop for Session {
     fn drop(&mut self) {
         debug!("Session is now being dropped.");
-
-        if let Some((terminator, _joiner)) = self.network_thread.take() {
-            let _ = terminator.send(NetworkEvent::Terminated);
-        }
+        unwrap!(self.send(CoreMsg::build_terminator()));
```
ustulation: Completely avoid `unwrap!` `unreachable!` `panic!` etc production code pls - we should take a very detailed look if we need one. Currently the only place it is in is `unwrap!(some_mutex.lock())` because the only time that can panic is if there is already a panic somewhere else in code. Otherwise there should be none of these in production code.

Here if you don't care about result - then do either `if let Err(e) = .... { info!("Unexepected error in drop: {:?}", e); }` or if you don't even care about that (since it is a drop) then `let _ = ....;`

src/ffi/low_level_api/data_id.rs
```diff
@@ -75,23 +82,27 @@ pub unsafe extern "C" fn data_id_new_appendable_data(id: *const [u8; XOR_NAME_LE
 
 /// Free DataIdentifier handle
 #[no_mangle]
-pub extern "C" fn data_id_free(handle: DataIdHandle) -> i32 {
+pub unsafe extern "C" fn data_id_free(session: *const Session, handle: DataIdHandle) -> i32 {
```
ustulation: I am not sure how good a practice it is to mix paradigms like this - so while the rest of the code is async all these are sync, which requires careful evaluation of whether it is expensive/time-consuming operation or not everytime we add functions/functionality. Besides it also forces us to use object-cache in both threads - FFI/JS and Core EL.

Would it be better to stick to one paradigm or do you think with all of these mods having:
```
#[no_mangle]
pub unsafe extern "C" fn data_id_free(session: *mut Session, handle: DataIdHandle, o_cb: extern "C" fn(i32)) -> i32 {
   helper::catch_unwind_i32(|| {
       let obj_cache = (*session).obj_cache.clone(); // Assuming it's Arc<Mutex<>>
       ffi_try!((*session).core_tx.send(CoreMsg::new(|_| o_cb(safe::data_id_free(&mut *unwrap!(obj_cache.lock()), handle)))); // Or if it is something very simple then just do it in here itself instead of calling `safe::data_id_free()` etc
       0
    })
}

mod safe {
    pub fn data_id_free(obj_cache: &mut ObjectCache, handle: DataIdHandle) -> Result<> { .. }
}

#[cfg(test)]
mod tests {
    .....
}
```

src/ffi/session.rs
```diff
@@ -190,104 +255,125 @@ pub unsafe extern "C" fn log_in(account_locator: *const u8,
...
     })
 }
 
 /// Return the amount of calls that were done to `post`
 #[no_mangle]
-pub unsafe extern "C" fn client_issued_posts(session_handle: *const SessionHandle) -> int64_t {
-    helper::catch_unwind_i64(|| {
+pub unsafe extern "C" fn client_issued_posts(session: *const Session,
+                                             user_data: usize,
+                                             o_cb: extern "C" fn(i32, usize, int64_t))
```
ustulation: Not mix idiom - so either make all C interface as `int32_t, size_t, int64_t` etc or make all `i32, usize, i64` etc. Pls scan the rest of the places this might be present too pls. Note for void it is recommended to use `libc::c_void` if using anywhere at all.

src/ffi/session.rs
```diff
@@ -288,8 +288,8 @@ pub unsafe extern "C" fn log_in(account_locator: *const u8,
 /// Return the amount of calls that were done to `get`
 #[no_mangle]
 pub unsafe extern "C" fn client_issued_gets(session: *const Session,
-                                            user_data: int64_t,
-                                            o_cb: extern "C" fn(int32_t, int64_t, int64_t))
+                                            user_data: OpaqueCtx,
```
ustulation: This is wrong. Why is an `extern "C"` function taking an `OpaqueCtx` ? C has no idea what this type is. It's supposed to be internal:
```
pub extern "C" fn foo_ffi(user_data: *mut c_void, o_cb: extern "C" fn(*mut c_void) {
    let ctx = OpaqueCtx(user_data);
    let msg = CoreMsg::new(move |client| {o_cb(ctx.0) .... });
    core_tx.send(msg); ...
}
```

src/ffi/session.rs
```diff
@@ -336,8 +336,8 @@ pub unsafe extern "C" fn client_issued_posts(session: *const Session,
 /// Return the amount of calls that were done to `delete`
 #[no_mangle]
 pub unsafe extern "C" fn client_issued_deletes(session: *const Session,
-                                               user_data: int64_t,
-                                               o_cb: extern "C" fn(int32_t, int64_t, int64_t))
+                                               user_data: OpaqueCtx,
+                                               o_cb: extern "C" fn(int32_t, OpaqueCtx, int64_t))
```
ustulation: these are all wrong - pls see the prev comment on it. This needs corrected in all similar places.

src/ffi/session.rs
```diff
@@ -190,104 +252,128 @@ pub unsafe extern "C" fn log_in(account_locator: *const u8,
...
     })
 }
 
 /// Return the amount of calls that were done to `delete`
 #[no_mangle]
-pub unsafe extern "C" fn client_issued_deletes(session_handle: *const SessionHandle) -> int64_t {
-    helper::catch_unwind_i64(|| {
+pub unsafe extern "C" fn client_issued_deletes(session: *const Session,
+                                               user_data: OpaqueCtx,
+                                               o_cb: extern "C" fn(int32_t, OpaqueCtx, int64_t))
```
ustulation: This is wrong. Why is an `extern "C"` function taking an `OpaqueCtx` ? C has no idea what this type is. It's supposed to be internal:
```
pub extern "C" fn foo_ffi(user_data: *mut c_void, o_cb: extern "C" fn(*mut c_void) {
    let ctx = OpaqueCtx(user_data);
    let msg = CoreMsg::new(move |client| {o_cb(ctx.0) .... });
    core_tx.send(msg); ...
}
```

src/ffi/session.rs
```diff
@@ -190,104 +252,128 @@ pub unsafe extern "C" fn log_in(account_locator: *const u8,
...
 }
 
 /// Return the amount of calls that were done to `append`
 #[no_mangle]
-pub unsafe extern "C" fn client_issued_appends(session_handle: *const SessionHandle) -> int64_t {
-    helper::catch_unwind_i64(|| {
+pub unsafe extern "C" fn client_issued_appends(session: *const Session,
+                                               user_data: OpaqueCtx,
+                                               o_cb: extern "C" fn(int32_t, OpaqueCtx, int64_t))
+                                               -> i32 {
```
ustulation: these are all wrong - pls see the prev comment on it. This needs corrected in all similar places.

PR #363, opened 2016-10-19
DNS

madadam:


maidsafe-highfive:
r? @krishnaIndia

src/dns/operations.rs
```diff
@@ -0,0 +1,873 @@
...
+                                encoded_dns_record,
+                                owners,
+                                vec![],
+                                private_signing_key,
+                                encryption_key)
+                .map(move |struct_data| (struct_data, saved_configs, long_name))
+                .map_err(DnsError::from)
+        })
+        .and_then(move |(struct_data, saved_configs, long_name)| {
+            client3.put_recover(Data::Structured(struct_data), None)
```
ustulation: We might need more than put recover here - if it fails it could also mean that it is present but free to be taken only if we supply the correct version.

src/dns/operations.rs
```diff
@@ -0,0 +1,873 @@
...
+use maidsafe_utilities::serialisation::{deserialise, serialise};
+use nfs::DirId;
+use routing::{Data, StructuredData, XorName, TYPE_TAG_DNS_PACKET};
+use routing::client_errors::{GetError, MutationError};
+use rust_sodium::crypto::{box_, sign, secretbox};
+use rust_sodium::crypto::hash::sha256;
+use std::collections::HashMap;
+use super::config::{self, DnsConfig};
+use super::DnsFuture;
+use super::errors::DnsError;
```
ustulation: maybe in `super` expose all these (whichever is required to the outside world like `DnsError` for sure) via `pub use` and simply use `super::DnsError` here. Also in that case make mods private, i.e. `pub mod errors -> mod errors`

Same for `core` mod above (like `routing`) - i think @nbaksalyar has already done something similar for `nfs` and `ffi` modules ?

So basically ppl will use `safe_core::core` or `safe_core::nfs` or `safe_core::dns` or `safe_core::ffi` base qualifications only and not more. These base ones need to exist however as they are fundamentally separate (and might even separate out into different crates in future).

src/core/client/mock_routing/mod.rs
```diff
@@ -343,7 +343,7 @@ impl MockRouting {
             match (data, storage.get_data(&data_name)) {
                 (Data::Structured(sd_new), Ok(Data::Structured(mut sd_stored))) => {
                     if sd_stored.is_deleted() {
-                        Some(MutationError::NoSuchData)
+                        Some(MutationError::InvalidOperation)
```
ustulation: Other places too have changes in master - for e.g. https://github.com/maidsafe/safe_core/blob/f4be866f8d474a2565f02bed99144f996ed828d3/src/core/client/non_networking_test_framework/mod.rs#L263 - maybe more - pls re-check.

src/core/client/mod.rs
```diff
@@ -331,114 +331,99 @@ impl Client {
...
+                            (Ok(Data::Structured(ref old)), Data::Structured(ref new)) if old.is_deleted() => {
+                                // The existing data is deleted. Attempt reclaim.
+                                let data = fry!(StructuredData::new(
+                                    new.get_type_tag(),
+                                    *new.name(),
+                                    old.get_version() + 1,
+                                    new.get_data().clone(),
+                                    new.get_owner_keys().clone(),
+                                    new.get_previous_owner_keys().clone(),
+                                    Some(&private_signing_key))
```
ustulation: rename as `sign_sk` - so the convention is `sk, pk` for secret key and public key. If qualification is required (i.e. in places where both sign and encryption keys are present) then use `sign_sk, enc_sk` etc. Pls scan in other places.

src/core/client/mod.rs
```diff
@@ -331,114 +331,99 @@ impl Client {
...
 
-                        if *owner_keys_after == owner_keys_before {
+                        if owner_match {
                             debug!("PUT recovery successful !");
-                            Ok(())
+                            ok!(version)
                         } else {
                             debug!("Data exists but we are not the owner.");
-                            Err(put_err)
+                            err!(put_err)
```
ustulation: `err!/ok!` already box stuffs - are things here not getting multi-boxed ? - If so pls scan other places too and correct.

src/core/client/mod.rs
madadam: OK, but for encryption we use symmetric keys, so the `_sk` / `_pk` prefix doesn't make sense, does it?

src/core/client/mod.rs
madadam: They are but it's necessary, unfortunately. The deleted SD branch returns a future, so need to box all the other branches too.

PR #364, opened 2016-10-21
(WIP) feat/core: implement network event observer

madadam:
Note: **DO NOT MERGE** yet. Missing tests.

### DIFF
--- a/src/core/client/mod.rs
+++ b/src/core/client/mod.rs
@@ -21,8 +21,9 @@ mod mock_routing;
 mod routing_el;
 
 use core::{CoreError, CoreFuture, CoreMsgTx, FutureExt, utility};
-use core::event::CoreEvent;
+use core::event::{CoreEvent, NetworkEvent};
 use futures::{self, Complete, Future, Oneshot};
+use futures::stream;
 use lru_cache::LruCache;
 use maidsafe_utilities::thread::{self, Joiner};
 use routing::{AppendWrapper, Authority, Data, DataIdentifier, Event, FullId, MessageId, Response,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @krishnaIndia

src/core/client/mod.rs
```diff
@@ -39,14 +40,18 @@ use self::mock_routing::MockRouting as Routing;
...
 use std::rc::Rc;
-use std::sync::mpsc::{self, Receiver};
+use std::sync::mpsc;
 use std::time::Duration;
 
 const CONNECTION_TIMEOUT_SECS: u64 = 60;
 const ACC_PKT_TIMEOUT_SECS: u64 = 60;
 const IMMUT_DATA_CACHE_SIZE: usize = 300;
 
+pub type NetworkEventSender = stream::Sender<NetworkEvent, ()>;
```
ustulation: We have the `Tx/Rx` convention, why are we using the longer alternative everywhere - `sender/receiver` ?

src/core/client/mod.rs
```diff
@@ -59,7 +64,8 @@ pub struct Client {
 
 struct Inner {
     routing: Routing,
-    heads: HashMap<MessageId, Complete<CoreEvent>>,
+    core_event_completes: HashMap<MessageId, Complete<CoreEvent>>,
+    network_event_senders: Vec<NetworkEventSender>,
```
ustulation: why is this a vector ?

src/core/client/routing_el.rs
```diff
@@ -104,15 +111,35 @@ pub fn parse_mutation_err(reason_raw: &[u8]) -> MutationError {
...
+        if let Some(complete) = complete {
+            complete.complete(event);
         }
         None
     });
 
     core_tx.send(msg).is_ok()
 }
+
+fn fire_network_event(core_tx: &CoreMsgTx, event: NetworkEvent) {
```
ustulation: Seems complicated for a simple task

src/core/event.rs
```diff
@@ -45,12 +45,12 @@ pub enum _NetworkEvent {
     Terminated,
```
ustulation: This variant may no longer be required as there is no `rx` as such.

ustulation:
Discarding in favour of https://github.com/maidsafe/safe_core/pull/365

PR #365, opened 2016-10-22
Misc fixes, Network event handling and Internal routing restarts

ustulation:
Many test cases would give false positives with previous test-utils which was error-prone and maybe even complicate to use.
MAID-1620
MAID-1623


maidsafe-highfive:
r? @krishnaIndia

PR #368, opened 2016-10-24
fix/ffi: fix sessions in FFI

nbaksalyar:
* Remove Clone in Session
* Refactor low_level_api/misc and low_level_api/data_id to always use
  asynchronous APIs
* Remove redundant ffi::app::drop_app function (app doesn't contain a
  session now, so it doesn't need to be dropped)

### DIFF
--- a/src/ffi/low_level_api/data_id.rs
+++ b/src/ffi/low_level_api/data_id.rs
@@ -157,25 +157,20 @@ mod tests {
 
         unsafe {
             let (tx, rx) = mpsc::channel::<DataIdHandle>();
-            assert_eq!(data_id_new_struct_data(sess_ptr,
-                                               type_tag,
-                                               &struct_id_arr,
-                                               Box::into_raw(Box::new(tx.clone())) as *mut _,
-                                               data_id_cb),
+            let tx = Box::into_raw(Box::new(tx)) as *mut c_void;
+
+            assert_eq!(data_id_new_struct_data(sess_ptr, type_tag, &struct_id_arr, tx, data_id_cb),
                        0);
             data_id_handle_struct = unwrap!(rx.recv());
 
-            assert_eq!(data_id_new_immut_data(sess_ptr,
-                                              &immut_id_arr,
-                                              Box::into_raw(Box::new(tx.clone())) as *mut _,
-                                              data_id_cb),
+            assert_eq!(data_id_new_immut_data(sess_ptr, &immut_id_arr, tx, data_id_cb),
                        0);
             data_id_handle_immut = unwrap!(rx.recv());
 
             assert_eq!(data_id_new_appendable_data(sess_ptr,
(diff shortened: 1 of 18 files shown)


src/ffi/low_level_api/data_id.rs
```diff
@@ -109,68 +151,101 @@ mod tests {
...
+
+            let tx: *mut mpsc::Sender<DataIdHandle> = tx as *mut _;
+            unwrap!((*tx).send(handle));
+        }
+
+        unsafe fn assert_free(sess_ptr: *const Session, handle: DataIdHandle, expected: i32) {
+            let (tx, rx) = mpsc::channel::<i32>();
+            assert_eq!(data_id_free(sess_ptr,
+                                    handle,
+                                    Box::into_raw(Box::new(tx.clone())) as *mut _,
```
ustulation: why do we need to clone it `tx.clone()` ?

src/ffi/low_level_api/data_id.rs
```diff
@@ -109,68 +151,101 @@ mod tests {
...
             assert_eq!(data_id_new_appendable_data(sess_ptr,
                                                    &priv_app_id_arr,
                                                    true,
-                                                   &mut data_id_handle_priv_appendable),
+                                                   Box::into_raw(Box::new(tx.clone())) as *mut _,
+                                                   data_id_cb),
                        0);
+            data_id_handle_priv_appendable = unwrap!(rx.recv_timeout(Duration::from_secs(15)));
+
+            let (tx, rx) = mpsc::channel::<DataIdHandle>();
```
ustulation: Why so many instances created ? Simply create one and use that:
```
let (tx, rx) = ...;
assert_eq!(ffi_0(..., &mut tx));
data_id_handle_... = unwrap!(rx.recv());

assert_eq!(ffi_1(..., &mut tx));
data_id_handle_... = unwrap!(rx.recv());
```
etc. - Pls aim to use use minimum `Box`s. Scan the code for more such issue.

src/ffi/low_level_api/data_id.rs
```diff
@@ -109,68 +151,101 @@ mod tests {
...
             assert_eq!(data_id_new_appendable_data(sess_ptr,
                                                    &pub_app_id_arr,
                                                    false,
-                                                   &mut data_id_handle_pub_appendable),
+                                                   Box::into_raw(Box::new(tx.clone())) as *mut _,
+                                                   data_id_cb),
                        0);
+            data_id_handle_pub_appendable = unwrap!(rx.recv_timeout(Duration::from_secs(15)));
```
ustulation: Do not use timeouts - when these are run against the network the timeout used may be insufficient - simply use `recv()` - we will code the overall timeout in `core`

ustulation:
will be merged with https://github.com/maidsafe/safe_core/pull/370

PR #369, opened 2016-10-24
implement ffi/dns

madadam:


maidsafe-highfive:
r? @krishnaIndia

src/nfs/helper/dir_helper.rs
```diff
@@ -290,6 +299,67 @@ pub fn configuration_dir(client: Client, dir_name: String) -> Box<NfsFuture<(Dir
...
+            dir.find_file(&file_name)
+                .map(|file| file.clone())
+                .ok_or(NfsError::FileNotFound)
+        })
+        .into_box()
+}
+
+/// Get the directory at the given tokenised path. The path is taken relative to
+/// `starting_directory`, unless it is `None`, in which case it is taken
+/// relative
```
ustulation: allignment broken - simple way to fix would be to write it as one long line - so in vim use capital `J` on the first line and keep using it until it becomes a long line. Then save it - rustfmt should do the required breaking up.

PR #372, opened 2016-10-27
Implement low-level api for struct data

madadam:


maidsafe-highfive:
r? @krishnaIndia

src/ffi/low_level_api/struct_data.rs
```diff
@@ -19,519 +19,583 @@
...
-
-
-        let handle = unwrap!(object_cache()).insert_sd(sd);
-        ptr::write(o_handle, handle);
-
-        0
-    })
+                                         data_len: usize,
+                                         user_data: *mut c_void,
+                                         o_cb: extern "C" fn(*mut c_void, int32_t, StructDataHandle)) {
```
nbaksalyar: This should be `unsafe extern "C"` because it dereferences a raw pointer in the first parameter.
Same applies for all `o_cb` callbacks

src/ffi/low_level_api/struct_data.rs
```diff
@@ -19,519 +19,583 @@
...
+                                         o_cb: extern "C" fn(*mut c_void, int32_t)) {
+    helper::catch_unwind_cb(|| {
+        (*session).send_cb(user_data, move |client, object_cache, user_data| {
+            let sd = {
+                let mut object_cache = unwrap!(object_cache.lock());
+                try_cb!(object_cache.get_sd(sd_h),
+                        |error| o_cb(user_data, error)).clone()
+            };
+
+            client.put(Data::Structured(sd), None)
```
ustulation: This should be put recover and after that update the SD in object cache (if successful put) if the version number changed.

src/ffi/low_level_api/struct_data.rs
```diff
@@ -19,519 +19,583 @@
...
-                                          sd.get_previous_owner_keys().clone(),
-                                          Some(&sign_key))
-        .map_err(CoreError::from));
-
-    let data = Data::Structured(new_sd.clone());
-    let resp_getter = try!(unwrap!(client.lock()).post(data, None));
-    try!(resp_getter.get());
-
-    Ok(new_sd)
+/// POST StructuredData
```
ustulation: There is also a function to make SD unclaimable - pls refer to master for changes.
Commits like https://github.com/maidsafe/safe_core/commit/f661b8f3b797ebe1e9194351013e37609a79c28f etc in case you missed and functions like https://github.com/maidsafe/safe_core/blob/f4be866f8d474a2565f02bed99144f996ed828d3/src/ffi/low_level_api/struct_data.rs#L372

Pls thoroughly check with master regarding changes. The only thing you can avoid is specifying `version` in create function https://github.com/maidsafe/safe_core/blob/f4be866f8d474a2565f02bed99144f996ed828d3/src/ffi/low_level_api/struct_data.rs#L39 because that was coded when put-recover wasn't there. With put-recover we can start with version-0 (default) and then put-recover will return the correct version.

Also check the changes in test cases thoroughly to see if everything is covered - there are changes in error variants etc.

ustulation: One more commit to keep in mind from master: https://github.com/maidsafe/safe_core/commit/b37c78ca0783c9e049017f6da58ad8b32628cd8c

src/ffi/low_level_api/struct_data.rs
madadam: Hm, but this is true for all callbacks. Should they all be `unsafe`?

src/ffi/low_level_api/struct_data.rs
```diff
@@ -529,40 +520,82 @@ pub unsafe extern "C" fn struct_data_free(session: *const Session,
...
+            let object_cache = session.object_cache();
+            let mut object_cache = unwrap!(object_cache.lock());
+            object_cache.insert_app(app)
+        };
+
+        // Create cipher opt.
+        let cipher_opt_h: CipherOptHandle = {
+            let (tx, rx) = mpsc::channel();
+
+            unsafe extern "C" fn callback(user_data: *mut c_void, error: int32_t, handle: u64) {
```
madadam: Good idea. There is definitely space to reduce boilerplate, but I only started porting the tests so didn't get to do it yet.

PR #373, opened 2016-10-27
fix/nfs: make dir_helper::create private (MAID-1624)

nbaksalyar:
* Make create, shallow_copy, and deep_copy functions private in
  dir_helper.
* Move move_dir function from FFI to NFS.
* Fix keys generation for private dirs in NFS.
* Refactor tests code in FFI


maidsafe-highfive:
r? @krishnaIndia

PR #377, opened 2016-10-31
feat/core: make deleted SD return empty data

madadam:


maidsafe-highfive:
r? @krishnaIndia

PR #378, opened 2016-10-31
Single threaded object cache

madadam:


maidsafe-highfive:
r? @krishnaIndia

src/core/utility/test_utils.rs
```diff
@@ -113,7 +113,7 @@ pub fn setup_client_with_net_obs<Create, NetObs, Run, I, E>(c: Create, mut n: Ne
     el_h.spawn(net_fut);
 
     let core_tx_clone = core_tx.clone();
-    unwrap!(core_tx.send(CoreMsg::new(move |client| {
+    unwrap!(core_tx.send(CoreMsg::new(move |client, _| {
```
ustulation: Instead of `_` maybe use `()` in `move |client, ()|` to be more explicit.

src/ffi/object_cache.rs
```diff
@@ -64,250 +66,204 @@ pub type SignKeyHandle = ObjectHandle;
...
         }
-
-        handle
     }
 
-    pub fn get_se_reader(&mut self,
-                         handle: SelfEncryptorReaderHandle)
-                         -> Result<&mut SelfEncryptorReaderWrapper, FfiError> {
-        self.se_reader.get_mut(&handle).ok_or(FfiError::InvalidSelfEncryptorHandle)
+    pub fn reset(&self) {
```
ustulation: You can just do `let _ = mem::replace(self, Self::new())` with `reset(&mut self)` or some one liner.

madadam: `reset` can't be `mut`, it would be impossible to call it from https://github.com/madadam/safe_core/blob/single-threaded-object-cache/src/ffi/low_level_api/misc.rs#L402

PR #380, opened 2016-11-01
feat/core: add timeout for client ops

madadam:


maidsafe-highfive:
r? @ustulation

src/core/client/mod.rs
```diff
@@ -46,8 +46,10 @@ use std::fmt;
 use std::rc::Rc;
 use std::sync::mpsc::{self, Receiver};
 use std::time::Duration;
+use tokio_core::reactor::{Handle, Timeout};
 
 const CONNECTION_TIMEOUT_SECS: u64 = 10;
+const REQUEST_TIMEOUT_SECS: u64 = 120;
```
ustulation: Pls arrange alphabetically.

And this should be feature dependent i think - if mock-routing then maybe 5 secs else 120 sec so that it is easier to test.

src/core/client/mod.rs
```diff
@@ -48,11 +48,15 @@ use std::sync::mpsc::{self, Receiver};
 use std::time::Duration;
 use tokio_core::reactor::{Handle, Timeout};
 
-const CONNECTION_TIMEOUT_SECS: u64 = 10;
-const REQUEST_TIMEOUT_SECS: u64 = 120;
 const ACC_PKT_TIMEOUT_SECS: u64 = 60;
+const CONNECTION_TIMEOUT_SECS: u64 = 10;
 const IMMUT_DATA_CACHE_SIZE: usize = 300;
 
+#[cfg(test)]
```
ustulation: Rather than for tests it should be feature dependent i think `#[cfg(feature = use-mock-routing)] ... 5; // else 120` and similarly the corresponding test should compile only if using mock-routing. This is because if we ever run tests against actual network we don't want a barrage of failures because the timeout was only 5 secs.

PR #382, opened 2016-11-02
Ffi refactor

madadam:
- use  `catch_unwind_cb` everywhere where appropriate
- make `catch_unwind_cb` more ergonomic
- introduce `try_cb` macro (like `try`, but calls the callback on error)
- fix order of callback arguments (user data first, error code second) in few odd places
- remove `ffi_try` and `ffi_try_ptr` macros (not needed anymore)


maidsafe-highfive:
r? @ustulation

PR #385, opened 2016-11-07
fix/core: remove client hook on timeout

madadam:


maidsafe-highfive:
r? @krishnaIndia

PR #386, opened 2016-11-07
chore/deps: remove redundant libc dependency

nbaksalyar:
libc dependency can be replaced with `std::os::raw::*`


maidsafe-highfive:
r? @ustulation

PR #388, opened 2016-11-11
refactor/ffi: impl Drop for FileDetails/DirDetails

madadam:
Made possible by the removal of drop flags in 1.13.


maidsafe-highfive:
r? @krishnaIndia

PR #392, opened 2016-11-22
MAID-1722 - add support for mutable data in mock routing

madadam:


maidsafe-highfive:
r? @krishnaIndia

PR #398, opened 2016-11-25
feat/client: update client creation functions

nbaksalyar:
Updates client creation functions to use MutableData and a new client authentication scheme for apps (using keys)


maidsafe-highfive:
r? @ustulation

PR #400, opened 2016-11-28
Implement client APIs

nbaksalyar:


maidsafe-highfive:
r? @ustulation

src/core/client/mod.rs
```diff
@@ -459,253 +418,152 @@ impl Client {
...
+
+        self.get(CoreEvent::ListMDataEntries, |routing, msg_id| {
+                routing.list_mdata_entries(dst.unwrap_or_else(|| Authority::NaeManager(name)),
+                                           name,
+                                           tag,
+                                           msg_id)
+            })
+            .and_then(|event| match event {
+                CoreEvent::ListMDataEntries(res) => res,
+                _ => Err(CoreError::ReceivedUnexpectedEvent),
```
ustulation: It would be good to at-least print these - so minimise underscore except in cases where we don't care at all (`let _ = stuff.insert(key, value)` etc. maybe). Here we can follow the convention of calling it `x` and printing it in a `debug!` call - `debug!("Expected ListMDataEntries, rxd unexpected event: {:?}", x);`

madadam: It might be even worth to write an internal macro for this, like: `expect_event!(event, CoreEvent::ListMDataEntries)`, as this snippet is being used repeatedly.

PR #401, opened 2016-11-28
feat/auth: implement basic auth types

vinipsmaker:


maidsafe-highfive:
r? @ustulation

src/auth/mod.rs
```diff
@@ -0,0 +1,91 @@
+
+use routing::XorName;
```
ustulation: Copyright should be the first thing.

src/auth/ffi.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub id: *mut u8,
+    /// TODO: doc
+    pub id_len: usize,
+    /// TODO: doc
+    pub id_cap: usize,
+
+    /// TODO: doc
+    pub scope: *mut u8, // null if not present
+    /// TODO: doc
+    pub scope_len: usize, // 0 if above is null
```
ustulation: Move the inline comment to the Doc comment (Keep the TODO)

src/auth/ffi.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// TODO: doc
+    pub scope_cap: usize, // 0 if above is null
```
ustulation: Move the inline comment to the Doc comment (Keep the TODO)

src/auth/mod.rs
vinipsmaker: `rustfmt` got confused because accidentally I've put an additional `/` on the first copyright line. Thanks. Fixing...

PR #404, opened 2016-11-30
Maid 1807

vinipsmaker:


maidsafe-highfive:
r? @ustulation

src/auth/mod.rs
```diff
@@ -39,6 +40,48 @@ pub struct ContainerPermission {
...
+    }
+
+    /// TODO: doc
+    #[allow(unsafe_code)]
+    pub unsafe fn from_raw(raw: *mut ffi::ContainerPermission) -> Self {
+        let raw = Box::from_raw(raw);
+        let ck = String::from_raw_parts(raw.container_key as *mut u8,
+                                        raw.container_key_len,
+                                        raw.container_key_cap);
+        let a = Vec::from_raw_parts(raw.access, raw.access_len, raw.access_cap);
```
ustulation: We don't need to collect it in a var (it's more confusing to read a variable name `a` for this purpose). Simply assign below.

src/auth/mod.rs
```diff
@@ -75,6 +227,38 @@ pub struct AppAccessToken {
...
+        }))
+    }
+
+    /// TODO: doc
+    #[allow(unsafe_code)]
+    pub unsafe fn from_raw(raw: *mut ffi::AppAccessToken) -> Self {
+        let raw = Box::from_raw(raw);
+        AppAccessToken {
+            enc_key: secretbox::Key::from_slice(&raw.enc_key)
+                .expect("Only fails if struct layout in code is wrong"),
```
ustulation: We don't use either `unwrap` or `expect` etc. Use `unwrap!()` that takes comments too.
But why not simply `secretbox::Key(raw.enc_key)`; (i am hoping enc_key is copyable). The inner stuff is `public` i think and then you don't require any of this.

src/auth/mod.rs
```diff
@@ -84,8 +268,124 @@ pub enum AuthResponse {
...
 }
+
+#[cfg(test)]
+#[allow(unsafe_code)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn test_container_permission() {
```
ustulation: it's already inside tests module and is marked as a test - we don't have to prefix tests with `test_`.

src/auth/mod.rs
vinipsmaker: You're right. Thanks for noting that. `secretbox::Key(raw.enc_key)` is much better

PR #406, opened 2016-11-30
feat/auth: authenticator client creation

nbaksalyar:


maidsafe-highfive:
r? @ustulation

src/util/ffi.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[test]
+    fn string_conversion() {
+        let (ptr, size, cap) = string_to_c_utf8(String::new());
+        assert_eq!(size, 0);
+        unsafe {
+            let _ = Vec::from_raw_parts(ptr, size, cap);
+        }
+
+        let (ptr, size, cap) = string_to_c_utf8("hello world".to_owned());
+        assert!(ptr != 0 as *mut u8);
```
ustulation: Don't we have `ptr.is_null()` or similar in the std ? If applicable pls see else where too.

PR #407, opened 2016-11-30
feat/auth: authenticator client creation

nbaksalyar:


maidsafe-highfive:
r? @ustulation

src/auth/authenticator.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        Ok(())
+    })
+}
+
+/// Create an authenticator as an unregistered client. This or any one of the other
+/// companion functions to get an auth instance must be called before initiating any
+/// operation allowed by this module.
+#[no_mangle]
+pub unsafe extern "C" fn create_unregistered(_auth_handle: *mut *mut Authenticator,
```
ustulation: Why is this here - it's doing nothing anyway

src/auth/authenticator.rs
```diff
@@ -0,0 +1,301 @@
...
+                                             -> i32 {
+    0
+}
+
+/// Log into a registered account. This or any one of the other companion
+/// functions to get an authenticator instance must be called before initiating
+/// any operation allowed for authenticator. `auth_handle` is a pointer to a pointer
+/// and must point to a valid pointer not junk, else the consequences are undefined.
+#[no_mangle]
+pub unsafe extern "C" fn log_in(account_locator: *const u8,
```
ustulation: Since there is already a review comment do you also mind changing this to `login` (without underscore) pls

PR #408, opened 2016-12-01
MAID-1825: implement key auth in mock routing

madadam:


maidsafe-highfive:
r? @ustulation

PR #410, opened 2016-12-06
Maid 1849

vinipsmaker:


maidsafe-highfive:
r? @krishnaIndia

src/ipc/mod.rs
```diff
@@ -411,4 +554,18 @@ mod tests {
...
         unsafe { ffi::auth_request_drop(a.into_ffi()) };
     }
+
+    #[test]
+    fn ipcresp_serialization() {
+        use rustc_serialize::json;
+
+        let value = IpcResp::Auth(Err(IpcError::ContainersDenied));
+        let json_value = json::encode(&value).unwrap();
+        assert_eq!(json_value, "{\"variant\":\"Auth\",\"fields\":[{\"variant\":\"Err\",\"fields\":[\"ContainersDenied\"]}]}");
```
ustulation: I am sure this will not pass rustfmt check - it's a long line

src/ipc/mod.rs
```diff
@@ -411,4 +554,18 @@ mod tests {
 
         unsafe { ffi::auth_request_drop(a.into_ffi()) };
     }
+
+    #[test]
+    fn ipcresp_serialization() {
```
ustulation: `ipc_resp_serialisation()`  note: `z` -> `s` (British english).

src/ipc/mod.rs
vinipsmaker: Fixed. I'll configure Emacs to run `rustfmt` on save tonight.

src/ipc/mod.rs
```diff
@@ -411,4 +554,20 @@ mod tests {
...
+        use rustc_serialize::json;
```
ustulation: Why not use our `maidsafe_utilities::serialise/deserialise` instead ? This is also what we will normally be using, we never use json anywhere in rust code.

src/ipc/mod.rs
```diff
@@ -411,4 +554,20 @@ mod tests {
...
+
+        let value = IpcResp::Auth(Err(IpcError::ContainersDenied));
+        let json_value = json::encode(&value).unwrap();
```
ustulation: so simply `serialise(&value)`.
And use unwrap!() instead of unwrap(). We never user unwrap(). It was mentioned previously too.

src/ipc/mod.rs
```diff
@@ -411,4 +554,20 @@ mod tests {
...
+        assert_eq!(json_value,
+                   "{\"variant\":\"Auth\",\"fields\":[{\"variant\":\"Err\",\"fields\":\
+                    [\"ContainersDenied\"]}]}");
+        let decoded_value: IpcResp = json::decode(&json_value).unwrap();
```
ustulation: unwrap() -> unwrap!()

Check elsewhere too.

PR #411, opened 2016-12-06
feat/auth: default dirs on account creation

vinipsmaker:


maidsafe-highfive:
r? @ustulation

src/auth/authenticator.rs
```diff
@@ -78,7 +83,43 @@ impl Authenticator {
...
+                        let encrypted_key = config_dir.enc_entry_key(Vec::from("access-container"))?;
+                        let encrypted_value = config_dir.enc_entry_value(serialised_dir)?;
+
+                        let _ = actions.insert(encrypted_key,
+                                               EntryAction::Ins(Value {
+                                                   content: encrypted_value,
+                                                   entry_version: 0,
+                                               }));
+
+                        Ok(cl3.mutate_mdata_entries(dir.name, dir.type_tag, actions))
```
ustulation: Do we need Ok(...) wrapper here ? `mutate_mdata_entries` returns a `Box<CoreFuture<()>>` so isn't that alone sufficient to return instead of wrapping in an `Ok()` ?

src/auth/authenticator.rs
```diff
@@ -78,7 +83,43 @@ impl Authenticator {
...
+                    }).and_then(move |x| {
```
ustulation: `x` is too vague - what does it represent here ? isn't it of type `()` ?

Rename `x` to something more explicit.

src/auth/authenticator.rs
```diff
@@ -78,7 +83,43 @@ impl Authenticator {
...
+                        x.map_err(AuthError::from)
+                    }).map(move |_| {
```
ustulation: Why is this an `_` ? If it has a `()` type then write it as `map(move |()| {` instead of `_`.

If it is some other type and needs to be `_` do reply to this review comment saying what it is and why we don't care.

src/auth/errors.rs
```diff
@@ -83,12 +91,26 @@ impl From<Utf8Error> for AuthError {
...
+}
+
 impl Debug for AuthError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match *self {
             AuthError::CoreError(ref error) => write!(f, "AuthError::CoreError -> {:?}", error),
             AuthError::IoError(ref error) => write!(f, "AuthError::IoError -> {:?}", error),
             AuthError::Unexpected(ref s) => write!(f, "AuthError::Unexpected{{{:?}}}", s),
+            AuthError::NfsError(ref error) => write!(f, "AuthError::NfsError -> {:?}", error),
+            AuthError::SerialisationError(ref error) => write!(f, "AuthError::SerialisationError -> {:?}", error),
```
ustulation: This line seems too too long for rustfmt. If it is then pls take time and effort to integrate rustfmt on save into your Emac or switch to vim and do it. We cannot have this repeated.

vinipsmaker: Done.

src/auth/authenticator.rs
vinipsmaker: > If it is some other type and needs to be `_` do reply to this review comment

No, it is `()`. Changed.

src/auth/authenticator.rs
vinipsmaker: Done. Changed to `fut`.

vinipsmaker:
Rebased as well.

PR #412, opened 2016-12-07
refactor/ipc: remove manual impl of RustCEncodable for IpcResp

vinipsmaker:
New version of rustc-serialise was released with builtin impl for
Result.


maidsafe-highfive:
r? @krishnaIndia

PR #420, opened 2016-12-14
feat/authenticator: implement IPC workflow

nbaksalyar:
Implements MAID-1852 (Authenticator IPC workflow).


maidsafe-highfive:
r? @krishnaIndia

src/authenticator/mod.rs
```diff
@@ -226,6 +240,19 @@ pub unsafe extern "C" fn create_acc(account_locator: FfiString,
     })
 }
 
+/// Create an authenticator as an unregistered client. This or any one of the other
```
ustulation: why is this back here

src/core/dir.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright 2016 MaidSafe.net limited.
```
ustulation: This file is already covered by mdata_info in upstream - there is no need for this file. The type Dir is now MDataInfo

src/ipc/resp/mod.rs
```diff
@@ -96,6 +97,21 @@ pub struct AppKeys {
 }
 
 impl AppKeys {
+    /// Generate random keys
+    pub fn generate(owner_key: sign::PublicKey) -> AppKeys {
```
ustulation: naming it `random` might be better in line with `MDataInfo::random_priv/pub()`

src/ipc/resp/mod.rs
```diff
@@ -164,6 +180,19 @@ impl AccessContainer {
             nonce: secretbox::Nonce(repr_c.nonce),
         }
     }
+
+    /// Creates an `AccessContainer` from a given `MDataInfo`
+    pub fn from_mdata_info(md: MDataInfo) -> Result<AccessContainer, IpcError> {
+        if let Some((_, Some(nonce))) = md.enc_info {
+            Ok(AccessContainer {
```
ustulation: Let's rename this to `AccessContInfo` to be in sync with MDataInfo

src/authenticator/ipc.rs
```diff
@@ -0,0 +1,559 @@
...
+    let user_data = OpaqueCtx(user_data);
+
+    catch_unwind_cb(user_data.0, o_err, || -> Result<_, IpcError> {
+        let msg = decode_msg(msg.as_str()?)?;
+        match msg {
+            IpcMsg::Req { req: IpcReq::Auth(auth_req), req_id } => {
+                let app_id = auth_req.app.id.clone();
+                let app_id2 = app_id.clone();
+
+                let _ = (*auth).send(move |client| {
```
ustulation: Why are we ignoring this result ? Do we not want to handle it in the normal way as others (i.e. return the error-code etc.) ? so maybe just `(*auth).send(...)?;`

src/authenticator/ipc.rs
```diff
@@ -0,0 +1,559 @@
...
+                return Err(IpcError::InvalidMsg.into());
+            }
+        }
+        Ok(())
+    })
+}
+
+/// Retrieves the authenticator configuration file
+fn get_config_file(client: Client) -> Box<AuthFuture<(u64, File)>> {
+    let dir = fry!(client.config_root_dir());
```
ustulation: Calling it `parent` instead of `dir` is better i think - so we immediately know that it's a parent dir.

src/authenticator/ipc.rs
```diff
@@ -0,0 +1,650 @@
...
+    file_helper::fetch(client, dir, CONFIG_FILE)
+        .map_err(From::from)
+        .into_box()
+}
+
+/// Retrieves the authenticator configuration
+pub fn get_config(client: Client, file: File) -> Box<AuthFuture<HashMap<sha256::Digest, AppInfo>>> {
```
ustulation: I think instead of going nfs::File route to store the config file, we can directly store encrypted `HashMap<Digest, AppInfo>` inline in entry-value. That will certainly make it more efficient and easier to manipulate. Even if each App corresponds to 1KiB data (very unlikely, in practice we'll probably be < 300 bytes per App) we will still need ~1000 apps to reach 1MiB. The limit is 100 for number of entries anyway so we will hit that cap 1st.
Can we write a TODO here mentioning something like this for later ?

src/authenticator/ipc.rs
```diff
@@ -0,0 +1,650 @@
...
+                let plaintext = fry!(dir2.decrypt(&val.content));
+                let dir = fry!(deserialise::<MDataInfo>(&plaintext));
+
+                c2.get_mdata_version(dir.name, dir.type_tag)
+                    .map(move |version| (dir, version))
+                    .into_box()
+            })
+            .and_then(move |(dir, version)| {
+                c3.set_mdata_user_permissions(dir.name,
+                                                dir.type_tag,
```
ustulation: is this alligned correctly ?

src/authenticator/ipc.rs
```diff
@@ -0,0 +1,650 @@
...
+        .into_box()
+}
+
+/// Creates a new app dedicated container
+fn create_app_container(client: Client,
+                        app_id: String,
+                        app_sign_pk: sign::PublicKey)
+                        -> Box<AuthFuture<MDataInfo>> {
+    let root_dir = fry!(client.user_root_dir());
+    let app_key = format!("apps/{}", app_id);
```
ustulation: better to call this `app_cont_name` - app_key doesn't make sense

src/authenticator/ipc.rs
nbaksalyar: > is this alligned correctly ?

Rustfmt thinks so :) I think this has to be tweaked in the format config

PR #421, opened 2016-12-15
test/app: MAID-1835

vinipsmaker:


maidsafe-highfive:
r? @krishnaIndia

src/app/ffi/mutable_data/tests.rs
```diff
@@ -0,0 +1,357 @@
...
+        client.put_mdata(mdata)
+            .map(|()| panic!("Put should be rejected by MaidManagers"))
+            .or_else(move |e| {
+                match e {
+                    CoreError::RoutingClientError(ClientError::InvalidOwners) => (),
+                    _ => panic!("Expected RoutingClientError"),
+                }
+                let mut owners = BTreeSet::new();
+                owners.insert(unwrap!(cl2.owner_key()));
+                let mdata = unwrap!(MutableData::new(rng.gen(),
```
ustulation: The previous data had failed put - so you should be able to use the same name instead of generating a new random one - that will showcase better testing.

src/app/ffi/mutable_data/tests.rs
```diff
@@ -0,0 +1,357 @@
...
+                                                     BTreeMap::new(),
+                                                     BTreeMap::new(),
+                                                     owners));
+                cl2.put_mdata(mdata)
+            })
+            .map(move |()| unwrap!(tx.send(())))
+            .map_err(|e| panic!("{:?}", e))
+            .into_box().into()
+    }));
+    unwrap!(rx.recv());
```
ustulation: Is this required ? Doesn't `App`'s dtor ensure you don't exit scope until graceful termination ? Try removing this channel mechanism and check that function is not existed until everything is done. Just test it using sleeps https://github.com/maidsafe/safe_core/pull/421/files#diff-b8b1fc7c47dc2bcb427282d188365ac2R63 etc until you are sure and if works then remove the sleeps of-course before final submission.

If it works pls remove channel mechanism from else where too.

src/app/ffi/mutable_data/tests.rs
```diff
@@ -0,0 +1,357 @@
...
+        let mut owners = BTreeSet::new();
+        owners.insert(unwrap!(client.owner_key()));
+
+        let name: XorName = rng.gen();
+        let mdata =
+            unwrap!(MutableData::new(name.clone(), DIR_TAG, permissions, BTreeMap::new(), owners));
+        let name2 = name.clone();
+        let cl2 = client.clone();
+        client.put_mdata(mdata)
+            .map_err(|e| panic!("{:?}", e))
```
ustulation: Instead of using map_err and map everywhere you can simply use `then(|res| let stuff = unwarp!(res); OR match res {` etc to assert it's the right error if it's an error etc. Scan and replace in other places too if it means less combinators and code.

src/app/ffi/mutable_data/tests.rs
```diff
@@ -0,0 +1,357 @@
...
+                match e {
+                    CoreError::RoutingClientError(ClientError::AccessDenied) => (),
+                    _ => panic!("Expected RoutingClientError"),
+                }
+                unwrap!(alt_client_tx.send((name2, sign_pk)))
+            })
+            .into_box()
+            .into()
+    }));
+    let _joiner = thread::named("", || {
```
ustulation: Give the thread some name - helps debugging when things fail. Scan elsewhere and correct too if used this way.

src/app/ffi/mutable_data/tests.rs
vinipsmaker: > Doesn't `App`'s dtor ensure you don't exit scope until graceful termination ?

I coded a sample where the test should fail, but test passed. I reintroduced the synchronization, and test failed as expected.

Also, the synchronization was inspired by another code already in place: https://github.com/maidsafe/safe_core/blob/c49b2392990693076847eda7c23c80907807f374/src/app/test_util.rs#L50

The dtor will do other thing.

src/app/ffi/mutable_data/tests.rs
vinipsmaker: Okay. Replacing where it means less code and less combinators. Leaving as is in other places.

PR #426, opened 2016-12-16
feat/authenticator: add API to get all registered apps

nbaksalyar:
Also contains changes to data types stored in the access container (replaced ContainerPermissions with BTreeSet<Permission>)

### DIFF
--- a/src/authenticator/ffi/mod.rs
+++ b/src/authenticator/ffi/mod.rs
@@ -22,7 +22,8 @@
 use core::FutureExt;
 use core::utility::symmetric_decrypt;
 use futures::Future;
-use ipc::Permission;
+use ipc::req::ffi::{ContainerPermissions, ContainerPermissionsArray, PermissionArray,
+                    container_permissions_array_free};
 use maidsafe_utilities::serialisation::deserialise;
 use std::{mem, ptr};
 use std::os::raw::c_void;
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @krishnaIndia

src/authenticator/ffi/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub containers: *mut Container,
+    /// Length of containers vector
+    pub containers_len: usize,
+    /// Capacity of containers vector
+    pub containers_cap: usize,
+}
+
+/// Container and a list of permissions given to an app
+#[repr(C)]
+pub struct Container {
```
ustulation: Would we not be able to reuse https://github.com/maidsafe/safe_core/blob/f02ef97836b41daac564cc64307eaa12d6fb56f4/src/ipc/req/ffi.rs#L116 instead of new type here ?

src/authenticator/ffi/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+use util::ffi::{FfiString, OpaqueCtx, catch_unwind_cb};
+use util::ffi::string::ffi_string_free;
+
+/// Application registered in the authenticator
+#[repr(C)]
+pub struct RegisteredApp {
+    /// Unique application identifier
+    pub app_id: FfiString,
+    /// List of containers that this application has access to
+    pub containers: *mut Container,
```
ustulation: Will we not be able to reuse https://github.com/maidsafe/safe_core/blob/f02ef97836b41daac564cc64307eaa12d6fb56f4/src/ipc/req/ffi.rs#L135 here ?

PR #436, opened 2016-12-20
MAID-1918: implement create_public_id

madadam:


safe_authenticator/src/public_id.rs
```diff
@@ -0,0 +1,85 @@
...
+        })
+        .and_then(move |(root_dir, info, data)| {
+            client3.put_mdata(data).map(move |_| (root_dir, info)).map_err(From::from)
+        })
+        .and_then(|(root_dir, info)| {
+            serialise(&info).map(move |encoded_info| (root_dir, encoded_info)).map_err(From::from)
+        })
+        .and_then(move |(root_dir, encoded_info)| {
+            let actions = EntryActions::new()
+                .ins(PUBLIC_ID_ENTRY_KEY.to_vec(), encoded_info, 0)
```
ustulation: Do we not have to 1st encrypt the key-value manually using root-dir mdata_info ?

safe_authenticator/src/public_id.rs
```diff
@@ -0,0 +1,88 @@
...
+        })
+        .and_then(|(root_dir, info)| {
+            let encoded = serialise(&info)?;
+            let encrypted = root_dir.enc_entry_value(&encoded)?;
+
+            Ok((root_dir, encrypted))
+        })
+        .and_then(move |(root_dir, encrypted_info)| {
+            let actions = EntryActions::new()
+                .ins(PUBLIC_ID_ENTRY_KEY.to_vec(), encrypted_info, 0)
```
ustulation: entry key is still plain text and needs to be encrypted using `root_dir.enc_entry_key(...)`

PR #439, opened 2017-01-02
Revoked apps deletion and more authenticator tests

nbaksalyar:
This PR covers both MAID-1899 and MAID-1904

### DIFF
--- a/safe_core/src/ipc/req/mod.rs
+++ b/safe_core/src/ipc/req/mod.rs
@@ -22,11 +22,13 @@
 /// Ffi module
 pub mod ffi;
 
+use errors::CoreError;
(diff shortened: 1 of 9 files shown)


safe_core/src/ipc/req/mod.rs
```diff
@@ -178,6 +180,16 @@ pub struct AppExchangeInfo {
 }
 
 impl AppExchangeInfo {
+    /// Creates a random AppExchangeInfo
+    pub fn rand() -> Result<AppExchangeInfo, CoreError> {
```
madadam: Is this function being used outside of tests? If not, then I'd suggest moving the logic over to a test-only utility,

PR #445, opened 2017-01-20
refactor/ffi: change FFI API to use copying only

nbaksalyar:


safe_core/src/ipc/req/mod.rs
```diff
@@ -55,40 +56,45 @@ pub struct AuthReq {
...
 /// After calling this function, the raw pointer is owned by the resulting
 /// object.
 #[allow(unsafe_code)]
-pub unsafe fn containers_from_repr_c(raw: ffi::ContainerPermissionsArray)
+pub unsafe fn containers_from_repr_c(raw: *const ffi::ContainerPermissions,
+                                     len: usize)
                                      -> Result<HashMap<String, BTreeSet<Permission>>, IpcError> {
     let mut result = HashMap::new();
-    let vec = raw.into_vec();
+    let vec = slice::from_raw_parts(raw, len).to_owned();
```
ustulation: Why not use the facility in ffi_utils::vec.rs - That uses [to_vec](https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec) which is a guaranteed copy as against [to_owned](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#tymethod.to_owned) used here which seems to contain a suspicious description:
> Creates owned data from borrowed data, _**usually**_ by cloning.

Pls scan elsewhere too.

safe_core/src/ipc/req/mod.rs
```diff
@@ -141,25 +150,30 @@ impl ContainersReq {
...
     }
 
     /// Constructs the object from the FFI counterpart.
     ///
     /// After calling this functions, the subobjects memory is owned by the
     /// resulting object.
     #[allow(unsafe_code)]
-    pub unsafe fn from_repr_c(repr_c: ffi::ContainersReq) -> Result<Self, IpcError> {
-        let ffi::ContainersReq { app, containers } = repr_c;
+    pub unsafe fn from_repr_c(repr_c: *const ffi::ContainersReq) -> Result<Self, IpcError> {
```
ustulation: In this PR it would be nice to rename all `from_xxx` family of functions to `from_xxx_cloned` so that merely looking at it the read knows what to expect and then if the code does not clone it one can file a bug report. Ofcourse do it for ones that clone it (most of our stuff including strings i guess). For ones that take ownership continue using `from_xxx` of-course. So pls do this in ffi_utils/safe_app/safe_authenticator/safe_core etc wherever applicable.

One e.g. where you would **_NOT_** change to `from_xxx_cloned` [is here](https://github.com/nbaksalyar/safe_core/blob/68017f31b3b9236665793a44a83ed087fc1e92ff/safe_core/src/ipc/resp/mod.rs#L134). That does not have internal pointers and everything is a copy and no drop needs to be implemented for it - so that's fine - FfiAppKeys should be taken and consumed by value and given by value.

PR #450, opened 2017-01-30
Expose testing framework in FFI API

nbaksalyar:


safe_core/src/utils/mod.rs
```diff
@@ -21,6 +21,7 @@
 
 #[macro_use]
 mod futures;
+#[cfg(feature = "testing")]
```
ustulation: Should be below the doc comment - pls scan the rest of places for this review point

safe_app/src/test_utils.rs
```diff
@@ -76,152 +81,70 @@ pub fn run<F, I, T>(app: &App, f: F) -> T
...
 
-/*
-// Create unregistered app.
-pub fn create_unregistered_app() -> App {
-    unwrap!(App::unregistered(|_| ()))
+/// Create app and grant it access to the specified containers.
+/// If `create_containers` is true, also create all the containers specified in
+/// the `access_info` and set their permissions accordingly.
+pub fn create_app_with_access(access_info: HashMap<String, BTreeSet<Permission>>) -> App {
+    let authenticator = authenticator::create_account_and_login();
```
ustulation: clashing names to variable and module authenticator - pls use diff name for var - e.g. `auth`

safe_app/src/ffi/ipc.rs
```diff
@@ -48,23 +52,28 @@ pub unsafe extern "C" fn encode_auth_req(req: *const FfiAuthReq,
...
         let encoded = CString::new(encoded)?;
 
-        o_cb(user_data, req_id, encoded.as_ptr());
+        o_cb(user_data, 0, req_id, encoded.as_ptr());
 
         Ok(())
-    })
+    });
+
+    if error_code != 0 {
```
ustulation: Wasn't there a helper in ffi-utils to already do this instead of call the cb explicitly for error ? If yes pls use that here and below too - also scan elsewhere for this

PR #459, opened 2017-02-16
fix/ffi: fix handles indirections in ObjectCache

nbaksalyar:


safe_app/src/object_cache.rs
```diff
@@ -83,7 +83,7 @@ pub struct ObjectCache {
     mdata_keys: Store<BTreeSet<Vec<u8>>>,
     mdata_values: Store<Vec<Value>>,
     mdata_entry_actions: Store<BTreeMap<Vec<u8>, EntryAction>>,
-    mdata_permissions: Store<BTreeMap<SignKeyHandle, MDataPermissionSetHandle>>,
+    mdata_permissions: Store<BTreeMap<User, MDataPermissionSetHandle>>,
```
ustulation: The handles so far are only present as keys to LRU cache. This one was an exception where it was stored in the value part as well (Key being `PermissionsHandle == Handle0` and Value being `BTreeMap<Handle1, Handle2>` in this case). This led to obfuscated API in which the API-user was not only supposed to remember `Handle0` but also `Handle1` and `Handle2`. So even if he obtained the value using `Handle0` he could not index it using newly obtained handles. Instead the key of the BTreeMap is now an object instead of a handle.

So previously:
```
/// Permissions-handle = h0. Say it contains one entry in BTreeMap - <SignKeyHandle0, PermissionSetHandle0>
/// To get permission-set the operation would be:
get_permission_set(permissions_h == h0, user == SignKeyHandle0) -> PermissionSetHandle0;
```
With this change it will be:
```
/// Permissions-handle = h0. Say it contains one entry in BTreeMap - <SignKey, PermissionSetHandle0>
/// To get permission-set the operation would be:
let sign_key_h = get_some_pub_sign_key();
get_permission_set(permissions_h == h0, user == sign_key_h) -> PermissionSetHandle0;
```
So it is more intuitivie to use as the user of api does not have to remember that one handle (`SignKeyHandle0`) was related and has to be used in a certain conjunction with another handle (`h0`)

This is better. But i think we are still not completely there. If the PermissionSetHandle0 was deleted explicitly from PermissionSet LRU or got auto-purged due to LRU behavior, then this particular `BTreeSet` corresponding to that will be useless (thus `h0` would be useless) and it will error out and then user will have to again figure out what happend (which is complicated to remember) and generate the new handles again from scratch). Rather, if the `BTreeMap<User, PermissionSet>` (instead of `PermissionSetHandle`) then we would completely have avoided the problem as even `PermissionSetHandles` and objects would be generated on the fly and things would be independent of previous handles (at the expense of a little bigger memory footprint - but PermissionSet is all stack obect and should be very fast to be copied ?)

madadam: Those are good points, but there is another angle to consider: There are conceptually two types - a single permission set, and a map of all permissions sets by users. We need APIs to query/manipulate them. If the map stores only handles to `PermissionSet`, then the API for permission sets can be used both for stand-alone `PermissionSet` and for ones inside the map.

On the other hand, if the map stored `PermissionSet`s directly, we would need another set API for permission sets in the map, different from the API for stand-alone permission sets.

ustulation: Ah right makes sense :+1:

PR #475, opened 2017-03-31
refactor/all: replace rustc-serialize with serde

nbaksalyar:
Also replace a deprecated dependency `time` with `chrono` and use `base64` instead of rustc-serialize functions.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,9 +31,9 @@ configuration:
 build_script:
   - |-
     cd ffi_utils && cargo build --verbose --release
-    cd ..\safe_core && cargo rustc --verbose --release --features testing --profile test -- -Zno-trans
-    cd ..\safe_app && cargo rustc --verbose --release --features testing --profile test -- -Zno-trans
-    cd ..\safe_authenticator && cargo rustc --verbose --features testing --release --profile test -- -Zno-trans
+    cd ..\safe_core && cargo rustc --verbose --release --features testing --profile test --lib -- -Zno-trans
+    cd ..\safe_app && cargo rustc --verbose --release --features testing --profile test --lib -- -Zno-trans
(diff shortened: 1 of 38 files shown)


appveyor.yml
```diff
@@ -31,9 +31,9 @@ configuration:
 build_script:
   - |-
     cd ffi_utils && cargo build --verbose --release
-    cd ..\safe_core && cargo rustc --verbose --release --features testing -- --test -Zno-trans
-    cd ..\safe_app && cargo rustc --verbose --release --features testing -- --test -Zno-trans
-    cd ..\safe_authenticator && cargo rustc --verbose --features testing --release -- --test -Zno-trans
+    cd ..\safe_core && cargo rustc --verbose --release --features testing --profile test -- -Zno-trans
```
ustulation: This is again not the same as in S.E that was linked by the previous review comment - it's missing `--lib` - pls follow exactly as done there unless there is a problem (then do report that).

safe_app/Cargo.toml
```diff
@@ -10,21 +10,22 @@ repository = "https://github.com/maidsafe/safe_client_libs"
 version = "0.1.0"
 
 [dependencies]
+chrono = { version = "~0.3.0", features = ["serde"] }
 ffi_utils = { path = "../ffi_utils", version = "~0.1.0" }
 futures = "~0.1.11"
 log = "~0.3.7"
 lru-cache = "~0.1.0"
-maidsafe_utilities = "~0.10.0"
+maidsafe_utilities = { git = "https://github.com/maidsafe/maidsafe_utilities.git", branch = "master" }
```
ustulation: pls point to crates 0.11.0 - all (except routing) have been published - so use latest for all dependencies in all the tomls of this workspace pls

PR #480, opened 2017-04-11
fix/ffi: fix users handling in permission sets

nbaksalyar:
Special value 0 for 'User::Anyone' wasn't handled correctly in several permission sets functions

### DIFF
--- a/safe_app/src/ffi/mutable_data/permissions.rs
+++ b/safe_app/src/ffi/mutable_data/permissions.rs
@@ -20,10 +20,15 @@
 use App;
 use errors::AppError;
 use ffi::helper::send_sync;
+use ffi::mutable_data::helper;
 use ffi_utils::{OpaqueCtx, catch_unwind_cb};
 use object_cache::{MDataPermissionSetHandle, MDataPermissionsHandle, SignKeyHandle};
 use routing::{Action, PermissionSet, User};
 use std::os::raw::c_void;
+
+/// Special value that represents `User::Anyone` in permission sets.
+#[no_mangle]
+pub static USER_ANYONE: u64 = 0;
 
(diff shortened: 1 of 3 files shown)


safe_app/src/ffi/mutable_data/permissions.rs
```diff
@@ -175,9 +175,12 @@ pub unsafe extern "C" fn mdata_permissions_get(app: *const App,
             let permissions = context
                 .object_cache()
                 .get_mdata_permissions(permissions_h)?;
-            let user_key = *context.object_cache().get_sign_key(user_h)?;
+            let user_key = match user_h {
+                0 => User::Anyone,
```
ustulation: expose this as cross-ffi const instead of magic number. Also the const might need to be a static if exposed this way. Might also require no-mangle attribute as otherwise the frontend will not be able to find the symbol easily. Pls check all these and ammend.

safe_app/src/ffi/mutable_data/permissions.rs
```diff
@@ -20,11 +20,19 @@
...
 use errors::AppError;
 use ffi::helper::send_sync;
+use ffi::mutable_data::helper;
 use ffi_utils::{OpaqueCtx, catch_unwind_cb};
 use object_cache::{MDataPermissionSetHandle, MDataPermissionsHandle, SignKeyHandle};
 use routing::{Action, PermissionSet, User};
 use std::os::raw::c_void;
 
+/// Special value that represents `User::Anyone` in permission sets.
+pub const ANYONE: u64 = 0;
```
ustulation: Unnecessary (already discussed but noting it here so that it's not forgotton or merged)

PR #483, opened 2017-04-20
feat/invitation: implement invitation-based account creation

nbaksalyar:
MAID-2056

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ script:
         (
             set -x;
             cd ffi_utils && unset Features && cargo clippy && cargo clippy --profile=test && export Features="use-mock-routing testing" &&
-            cd ../safe_core && cargo clippy && cargo clippy --profile=test --features="$Features" &&
+            cd ../safe_core && cargo clippy --features="$Features" && cargo clippy --profile=test --features="$Features" &&
             cd ../safe_app && cargo clippy --profile=test --features="$Features" &&
             cd ../safe_authenticator && cargo clippy --profile=test --features="$Features";
         )
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ script:
         (
             set -x;
             cd ffi_utils && unset Features && cargo clippy && cargo clippy --profile=test && export Features="use-mock-routing testing" &&
-            cd ../safe_core && cargo clippy --features="$Features" && cargo clippy --profile=test --features="$Features" &&
+            cd ../safe_core && cargo clippy && cargo clippy --profile=test --features="$Features" &&
             cd ../safe_app && cargo clippy --profile=test --features="$Features" &&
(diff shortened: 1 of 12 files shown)


safe_core/src/client/mod.rs
```diff
@@ -187,6 +237,10 @@ impl Client {
             b"Login".to_vec() => Value {
                 content: acc.encrypt(&user_cred.password, &user_cred.pin)?,
                 entry_version: 0,
+            },
+            b"Invitation".to_vec() => Value {
```
ustulation: (1) Since it's key value, maybe put it only if `!invitation.is_empty()`
(2) In case it is not empty and put, this must be undone soon. Otherwise looking at the account packet it can be associated with the account-holder easily because say the forum-mods know who they gave what invitation to. Then if the account packet contains that invitation, it can be easily pointed out who it belongs to. So `master` blanks it out on the subsequent update of the account packet. This branch should also do the same pehaps somewhere here: https://github.com/maidsafe/safe_client_libs/pull/483/files#diff-34a9cb68f9a2eeb813c7d40618df6d93L742 as it's just for one time use by the vaults.
(3) Since vaults are going to search for this key, make `Invitation` a static str in routing (mutable data branch):
```
const ACC_INVITATION_ENTRY_KEY: &'static str = "Invitation";
```
so that vaults and client use the same key. Put it in client-errors module in routing for now, also putting a TODO to move it to a better place later.

safe_core/Cargo.toml
```diff
@@ -28,6 +28,16 @@ self_encryption = { git = "https://github.com/maidsafe/self_encryption.git", bra
...
+rustc-serialize = "~0.3.23"
+
 [features]
 use-mock-routing = []
 testing = []
+
+[[example]]
+bench = false
+name = "gen_invites"
+required-features = ["testing", "use-mock-routing"]
```
ustulation: Why are these required and how will we use it against actual routing network to populate the invites if it only works with the mock ?

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
         (
             set -x;
             cd ffi_utils && unset Features && cargo clippy && cargo clippy --profile=test && export Features="use-mock-routing testing" &&
-            cd ../safe_core && cargo clippy && cargo clippy --profile=test --features="$Features" &&
+            cd ../safe_core && cargo clippy --features="$Features" && cargo clippy --profile=test --features="$Features" &&
```
ustulation: Check comment on line: https://github.com/maidsafe/safe_client_libs/pull/483/files#diff-4c8012f3397ebf5680fca0be3704dc25R43 - hence this change might not be necessary.

safe_core/Cargo.toml
nbaksalyar: Yeah, I think only `testing` is required for `test_utilits`, `use-mock-routing` should be dropped from the requirements.

PR #485, opened 2017-04-26
feat/safe_app: add crypto functions to FFI

nbaksalyar:


safe_app/src/ffi/crypto.rs
```diff
@@ -0,0 +1,524 @@
...
+}
+
+/// Encrypts arbitrary data using a given key pair.
+/// You should provide a recipient's public key and a sender's secret key.
+#[no_mangle]
+pub unsafe extern "C" fn encrypt(app: *const App,
+                                 data: *const u8,
+                                 len: usize,
+                                 pk_h: EncryptKeyHandle,
+                                 sk_h: SecretKeyHandle,
```
ustulation: It's far better to call one `EncryptPubKeyHandle` and the other `EncryptSecKeyHandle` otherwise the naming is very confusing and asymmetrical.

PR #488, opened 2017-05-01
refactor/ffi: remove automatic encryption of MData actions

nbaksalyar:
To make the API more consistent, we don't encrypt values and keys in MutableData actions automatically now - a user has to do this manually. The same applies to mdata_get_value function, which doens't decrypt values from private MData anymore. This commit also adds tests for MutableData FFI APIs which were missing.

### DIFF
--- a/safe_app/src/ffi/mutable_data/mod.rs
+++ b/safe_app/src/ffi/mutable_data/mod.rs
@@ -35,11 +35,21 @@ use safe_core::{CoreError, FutureExt, mdata_info};
 use std::os::raw::c_void;
 use std::ptr;
 
+/// Special value that represents an empty permission set.
+#[no_mangle]
+pub static PERMISSIONS_EMPTY: u64 = 0;
+
+/// Special value that represents an empty entry set.
+#[no_mangle]
+pub static ENTRIES_EMPTY: u64 = 0;
+
 /// Create new mutable data and put it on the network.
 ///
 /// `permissions_h` is a handle to permissions to be set on the mutable data.
-/// If 0, the permissions will be empty.
-/// `entries_h` is a handle to entries for the mutable data. If 0, the entries will be empty.
+/// If `PERMISSIONS_EMPTY`, the permissions will be empty.
+///
+/// `entries_h` is a handle to entries for the mutable data.
+/// If `ENTRIES_EMPTY`, the entries will be empty.
 #[no_mangle]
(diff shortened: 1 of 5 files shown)


ustulation:
can you rebase with latest dev pls @nbaksalyar

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -121,6 +131,9 @@ pub unsafe extern "C" fn mdata_get_version(app: *const App,
 ///     3. pointer to content
 ///     4. content length
 ///     5. entry version
+///
+/// Please notice that if a value is fetched from a private `MutableData`,
+/// it's not automatically decrypted.
```
ustulation: If the get is not automatically decrypting, hope the put does not do it as well. Can you pls confirm that too ?

nbaksalyar:
Yep, it's been an inconsistent behaviour. Put operations don't encrypt keys
& values automatically now, too.

2 мая 2017 г. 4:25 ПП пользователь "Spandan Sharma" <
notifications@github.com> написал:

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -144,10 +157,7 @@ pub unsafe extern "C" fn mdata_get_value(app: *const App,
 
             client
                 .get_mdata_value(info.name, info.type_tag, key)
-                .and_then(move |value| {
-                              let content = info.decrypt(&value.content)?;
-                              Ok((content, value.entry_version))
-                          })
+                .and_then(move |value| Ok((value.content, value.entry_version)))
```
bochaco: I don't think there is a way to decrypt the value after it's retrieved with this, perhaps we are missing a `mdata_info_decrypt_entry_value` function?

nbaksalyar: @bochaco, thanks, good catch - I should add `mdata_info_decrypt`.

bochaco: I'm not sure but I assume we'd need the `mdata_info_decrypt_entry_key` counterpart if the entries keys and values are returned encrypted when I iterate thru them.

nbaksalyar: That should be fine, as we can use the same decryption function for both keys and values. E.g. look there: https://github.com/maidsafe/safe_client_libs/blob/dev/safe_core/src/client/mdata_info.rs#L97-L104

PR #490, opened 2017-05-08
refactor/ffi: use new convention for errors in callbacks

nbaksalyar:
Now, instead of passing just an error code (which isn't useful in case of an error types that carry extra info,
like e.g. `AppError::Unexpected` and others), we pass an error description along with it. All callbacks should take the `FfiResult` argument in place of `i32` error code.


ffi_utils/src/lib.rs
```diff
@@ -83,3 +83,18 @@ pub trait ErrorCode {
...
+#[repr(C)]
+pub struct FfiResult {
+    /// Unique error code
+    pub error_code: i32,
+    /// Error description
+    pub description: *const c_char,
+}
+
+/// Constant value to be used for OK result
+pub const FFI_RESULT_OK: FfiResult = FfiResult {
```
ustulation: Do we want this value to be exposed to the FFI too ? In that case shouldn't it be a static because const is not guaranteed to be exported as a symbol of the library.

PR #504, opened 2017-06-21
fix/mock: handle low balance

madadam:


safe_core/src/client/mock/routing.rs
```diff
@@ -58,6 +58,12 @@ lazy_static! {
     static ref VAULT: Mutex<Vault> = Mutex::new(Vault::new());
 }
 
+fn lock_vault() -> MutexGuard<'static, Vault> {
```
ustulation: ah nice - should be less prone to errors in case we forgot to call load everytime :+1:

PR #508, opened 2017-06-27
feat/client: repeat operation on rate limit error

madadam:


safe_core/src/client/mod.rs
```diff
@@ -913,73 +897,48 @@ impl Client {
 
     /// Generic GET request
     fn get<T, F, G>(&self, err_event: F, req: G) -> Box<CoreFuture<CoreEvent>>
```
ustulation: Should this be named better - it sounds misleading

safe_core/src/utils/futures.rs
```diff
@@ -69,3 +69,48 @@ impl<F: Future + 'static> FutureExt for F {
...
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let result = match self.future.poll() {
+            Ok(Async::NotReady) => return Ok(Async::NotReady),
+            Ok(Async::Ready(item)) => Ok(item),
+            Err(error) => Err(error),
+        };
+
+        if (self.cond)(&result) {
+            self.future = (self.init)();
```
ustulation: Will this not spam re-try on rate-exceeded error ? Ideally we want to wait for RETRY_TIMEOUT = 800 ms.

madadam: Implemented directly in `Client`.

PR #510, opened 2017-06-29
refactor/all: change network re-connect fns

nbaksalyar:
Now when the network connection is failed, we do not reconnect to the network automatically - it should be done by a user. For this purpose several new FFI functions have been added - `app_reconnect` and `auth_reconnect`.

Because of this change we have to store `core_tx` channel with the client. That means that `Client` now has to accept a type parameter for the message context - so all `Client` signatures have been changed to `Client<T>`.

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -449,8 +449,6 @@ impl<T: 'static> Client<T> {
 
         let (routing, routing_rx) = setup_routing(opt_id, self.inner().client_type.config())?;
 
-        self.inner().net_tx.send(NetworkEvent::Connected)?;
-
         let joiner = spawn_routing_thread(routing_rx,
                                           self.inner().core_tx.clone(),
                                           self.inner().net_tx.clone());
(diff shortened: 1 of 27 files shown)


ustulation:
Nicely done :+1:

safe_core/src/client/mod.rs
```diff
@@ -434,44 +439,27 @@ impl Client {
...
-                let _ = core_tx.send(msg);
-                return;
-            }
-        };
+        let (routing, routing_rx) = setup_routing(opt_id, self.inner().client_type.config())?;
 
-        if let Err(e) = net_tx.send(NetworkEvent::Connected) {
-            trace!("Couldn't send: {:?}", e);
-        }
+        self.inner().net_tx.send(NetworkEvent::Connected)?;
```
ustulation: The only thing i am thinking here is that if it's already connected and someone calls this - we will send `connected` followed by `disconnected` for the previous instance of routing which will only be destroyed below when doing `self.routing = new-routing`. If so that'll apprear strange to the FE apps. See if you can fire the connection-established event at the end maybe (atleast after dropping previous routing by way of replacement below) or some other way ?

PR #514, opened 2017-07-06
refactor/safe_core: improve NFS implementation

nbaksalyar:
Previously, the NFS API in safe_app didn't use all functions provided by safe_core and had a confusing
interface. This commit applies multiple breaking changes to the API with the purpose of cleaning it up and fixing multiple bugs (e.g. file modification time wasn't updated properly).

### DIFF
--- a/safe_core/src/nfs/errors.rs
+++ b/safe_core/src/nfs/errors.rs
@@ -74,6 +74,45 @@ impl From<SelfEncryptionError<SelfEncryptionStorageError>> for NfsError {
     }
 }
 
+impl fmt::Display for NfsError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            NfsError::CoreError(ref error) => write!(f, "Client Errror: {}", error),
+            NfsError::DirectoryExists => {
+                write!(f,
+                       "Directory already exists with the same name in the same level")
+            }
+            NfsError::DestinationAndSourceAreSame => write!(f, "Destination is Same as the Source"),
+            NfsError::DirectoryNotFound => write!(f, "Directory not found"),
+            NfsError::FileExists => {
+                write!(f, "File Already exists with the same name in a directory")
+            }
+            NfsError::FileDoesNotMatch => {
+                write!(f,
+                       "File does not match with the existing file in the directory listing")
+            }
+            NfsError::FileNotFound => write!(f, "File not found"),
+
+            NfsError::InvalidRange => write!(f, "Invalid byte range specified"),
+            NfsError::InvalidParameter => {
+                write!(f,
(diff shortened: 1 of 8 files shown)


safe_core/src/nfs/errors.rs
```diff
@@ -74,6 +74,45 @@ impl From<SelfEncryptionError<SelfEncryptionStorageError>> for NfsError {
     }
 }
 
+impl fmt::Display for NfsError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            NfsError::CoreError(ref error) => write!(f, "Client Errror: {}", error),
+            NfsError::DirectoryExists => {
+                write!(f,
+                       "Directory already exists with the same name in the same level")
```
madadam: What does "level" mean here?

safe_core/src/nfs/errors.rs
```diff
@@ -74,6 +74,45 @@ impl From<SelfEncryptionError<SelfEncryptionStorageError>> for NfsError {
...
+            }
+            NfsError::DestinationAndSourceAreSame => write!(f, "Destination is Same as the Source"),
```
madadam: I wouldn't capitalize "Same" and "Source"

safe_core/src/nfs/errors.rs
```diff
@@ -74,6 +74,45 @@ impl From<SelfEncryptionError<SelfEncryptionStorageError>> for NfsError {
...
+            NfsError::DirectoryNotFound => write!(f, "Directory not found"),
+            NfsError::FileExists => {
+                write!(f, "File Already exists with the same name in a directory")
```
madadam: I wouldn't capitalize "Already".

safe_core/src/nfs/errors.rs
```diff
@@ -74,6 +74,45 @@ impl From<SelfEncryptionError<SelfEncryptionStorageError>> for NfsError {
...
+            NfsError::FileDoesNotMatch => {
+                write!(f,
+                       "File does not match with the existing file in the directory listing")
+            }
+            NfsError::FileNotFound => write!(f, "File not found"),
+
+            NfsError::InvalidRange => write!(f, "Invalid byte range specified"),
+            NfsError::InvalidParameter => {
+                write!(f,
+                       "Validation error - if the field passed as parameter is not valid")
```
madadam: Shouldn't this be "Validation error - the field passed as parameter is not valid" (that is, remove "if") ?

safe_core/src/nfs/errors.rs
nbaksalyar: Actually, these are remnants of the previous NFS implementation. This error case isn't used anywhere, so it can be removed. Nice catch! :+1:

PR #516, opened 2017-07-19
Expose API to get account info

nbaksalyar:
* Expose API to get the account info (used & available mutations)
* Rename functions prefixed with `authenticator_*` to `auth_*`
* Refactor & improve macros in `ffi_utils`


safe_app/src/ffi/mod.rs
```diff
@@ -131,14 +132,34 @@ pub unsafe extern "C" fn app_reconnect(app: *mut App,
...
-                 description: description.as_ptr(),
-             });
+    if let Err(..) = res {
+        call_result_cb!(res, user_data, o_cb);
+    }
+}
+
+/// Get the account usage statistics.
+#[no_mangle]
+pub unsafe extern "C" fn app_account_info(app: *mut App,
```
bochaco: I think the `app_account_info` should return an object with the info, now we are returning just a couple of values (mutations done & available), but I presume we will want to return more information in the future...?...

PR #517, opened 2017-07-21
(WIP) Recovery

madadam:

PR #518, opened 2017-07-26
Operations recovery

nbaksalyar:
Implements recovery for composite operations in safe_authenticator and a new testing framework in safe_core mock-routing. Supersedes #517.


safe_core/src/client/mock/routing.rs
```diff
@@ -141,12 +151,29 @@ impl Routing {
                      data: ImmutableData,
                      msg_id: MessageId)
                      -> Result<(), InterfaceError> {
+        let data_name = *data.name();
+        let nae_auth = Authority::NaeManager(data_name);
+
+        if self.request_hook.is_some() {
+            let override_response = if let Some(ref mut hook) = self.request_hook {
```
ustulation: This looks a bit off - you have already asserted `self.request_hook` is `Some` in the `if` condition above. So when will it ever come to the `else` of this inner `if` ?
Could this whole thing be not just written as:
```
if let Some(ref mut hook) = self.request_hook {
   let resp = hook(...);
   self.send_response(..., resp);
   return Ok(());
}```

File: safe_authenticator/src/revocation.rs [hunk 2]
```diff
@@ -0,0 +1,463 @@
...
+                let fut = if !queue.contains(&app_id) {
+                    config::push_to_revocation_queue(&c3, app_id)
+                } else {
+                    // The queue already contains this app, do nothing
+                    ok!(())
+                };
+                fut.map(move |_| current_item)
+            })
+            .and_then(move |app_id| revoke_single_app(&c4, &app_id))
+            .and_then(move |_| config::pop_from_revocation_queue(&c5))
```
madadam: We should make sure we are popping the right item. If we allow concurrent revocations, then someone else could have popped the front of the queue already and we would be removing wrong item. Perhaps better approach would be to fetch the whole queue, check whether the app_id we just processed is still there, and remove it only if it is, then put the queue back. 

So this, together with the above comments make me think whether we even need `push_to_revocation_queue` and `pop_from_revocation_queue` at all?

safe_core/src/client/mock/routing.rs
```diff
@@ -190,6 +211,18 @@ impl Routing {
...
+        let override_response = if let Some(ref mut hook) = self.request_hook {
+            hook(&Request::GetIData { name, msg_id })
+        } else {
+            None
+        };
+        if let Some(response) = override_response {
+            self.send_response(GET_IDATA_DELAY_MS, nae_auth, self.client_auth, response);
+            return Ok(());
+        }
+
```
madadam: Consider extracting this block of code (and others like it) into a method, to reduce repetition. Something like:

```rust
fn fire_request_hook(&mut self, src: Authority, dst: Authority, delay: u64, request: Reqest) -> bool;
```
Where the return value of `true` means proceed normally and `false` means stop and return `Ok())` (or the other way around).

Alternatively, `request` could be a `FnOnce() -> Request`, for lazy request initialization to avoid unnecessary clones.

safe_core/src/client/account.rs
```diff
@@ -32,6 +32,8 @@ pub struct Account {
     pub user_root: MDataInfo,
     /// The users configuration directory
     pub config_root: MDataInfo,
+    /// Flag set to true when user_root and config_dir are actually created
+    pub root_dirs_created: bool,
```
madadam: Why do we need this as opposed to checking the corresponding mutable data exists (with `get_mdata_version` for example)?

safe_core/src/client/mock/tests.rs
```diff
@@ -1216,6 +1024,98 @@ fn simulate_rate_limit_errors() {
...
+            Request::MutateMDataEntries { tag, msg_id, .. } if tag == 12345u64 => {
+                Some(Response::MutateMDataEntries {
+                    res: Err(ClientError::from("hello world")),
+                    msg_id,
+                })
+            }
+            // Pass-through
+            _ => None,
+        }
+    });
```
madadam: Minor, but I'd extract those two magic numbers (`10000` and `12345`) into named constants (local to this function), as they are used more than once.

safe_core/src/client/account.rs
nbaksalyar: @madadam it's to avoid doing extra requests on `login` - now we always fetch the account packet from the network, so checking whether root dirs have been created is an atomic and simple operation (as opposed to doing `get_mdata_version` every time a user logs into their account)

safe_authenticator/src/revocation.rs
```diff
@@ -0,0 +1,463 @@
...
+                version + 1,
+            )
+        })
+        .into_box()
+}
+
+/// Delete the app's auth key from the Maid Manager - this prevents the app from
+/// performing any more mutations.
+fn delete_app_auth_key(client: &Client<()>, key: sign::PublicKey) -> Box<AuthFuture<()>> {
+    let client = client.clone();
```
madadam: Maybe this can be extracted into the `recovery` module? Especially since `ins_auth_key` is already there.

PR #519, opened 2017-07-27
Support ShareMData IPC request to get access to arbitrary MD on the network (MAID-2149)

canndrew:


safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +236,123 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
                                    req_id,
                                }) => {
                                 o_unregistered(user_data.0, req_id);
+                                future::ok(()).into_box()
+                            }
+                            Ok(IpcMsg::Req {
+                                   req: IpcReq::ShareMData(share_mdata_req),
+                                   req_id,
+                               }) => {
+                                match client_cloned.public_signing_key() {
```
nbaksalyar: Better to use a `?` operator here. If it's not available due to result being a `Future` type, you can use our [custom `fry!` macro](https://github.com/maidsafe/safe_client_libs/blob/f642805c20450981b0600c6ad88e049a50b8b172/safe_core/src/utils/futures.rs#L20)

safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +236,123 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                    Ok(user) => {
+                                        let user_data_inner = user_data.0;
+                                        let num_mdatas = share_mdata_req.mdata.len();
+                                        let mut mdata_futures = Vec::with_capacity(num_mdatas);
+                                        for mdata in &share_mdata_req.mdata {
+                                            let name = mdata.name;
+                                            let type_tag = mdata.type_tag;
+                                            let metadata_key = mdata.metadata_key.clone();
+                                            let client_cloned_cloned = client_cloned.clone();
```
nbaksalyar: Better to name it like `c2` and `c3` instead of longer names like `client_cloned` or `client_cloned_cloned`, it's an informal convention that we follow

safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +236,123 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                            let mdata_future = client_cloned
+                                                    .get_mdata_shell(name, type_tag)
+                                                    .and_then(move |md| {
+                                                        if md.owners().contains(&user) {
+                                                            match metadata_key {
+                                                                None => future::ok(Ok(None)).into_box(),
```
nbaksalyar: We have a handy macro for `future::ok(val).into_box()` - `ok!(val)`. So in this case it'd be better to write it as `ok!(Ok(None))`

safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +236,123 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                                                },
+                                                            }
+                                                        } else {
+                                                            future::ok(Err((type_tag, name))).into_box()
+                                                        }
+                                                     })
+                                                     .map_err(AuthError::CoreError);
+                                            mdata_futures.push(mdata_future);
+                                        }
+                                        match share_mdata_req.into_repr_c() {
```
nbaksalyar: Same thing here for `into_repr_c()?` or `fry!(share_mdata_req.into_repr_c())`

safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +236,123 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                                future::join_all(mdata_futures)
+                                                    .and_then(move |results| {
+                                                        let mut metadatas = Vec::with_capacity(num_mdatas);
+                                                        let mut results_iter = results.into_iter();
+                                                        while let Some(result) = results_iter.next() {
+                                                            match result {
+                                                                Ok(metadata_opt) => {
+                                                                    let (data, len) = match metadata_opt {
+                                                                        Some(metadata) => {
+                                                                            let (data, len, _) = vec_into_raw_parts(metadata);
```
nbaksalyar: `vec_into_raw_parts` is a tricky thing: it calls `mem::forget` on the vector, so further it will be a responsibility of a user to deallocate this data. We don't do manual deallocations on the front-end side though (we copy/clone all data verbatim), so it will be a memory leak here. Better to just use pointers to raw vec data (it's guaranteed that the vec memory is contigious) - e.g. as `MDataMeta { data: data.as_ptr(), len: data.len() }`. This way vectors will be deallocated automatically after they go out of scope, but make sure that they go out of scope only after `o_cb` is called though.

safe_authenticator/src/ipc.rs
```diff
@@ -265,6 +384,104 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+    let user_data = OpaqueCtx(user_data);
+
+    catch_unwind_cb(user_data.0, o_cb, || -> Result<(), AuthError> {
+        let share_mdata_req = ShareMDataReq::clone_from_repr_c(req)?;
+        if is_granted {
+            (*auth).send(move |client| {
+                let client_cloned0 = client.clone();
+                let client_cloned1 = client.clone();
+                let user_data = user_data.0;
+                Some(get_config(client).and_then(move |(_cfg_version, config)| {
```
nbaksalyar: For convenience, you can use the `future.into_box().into()` method that will convert the future into `Option<Box<Future>>` as a last step (so no need to wrap it in `Some` explicitly)

safe_authenticator/src/ipc.rs
```diff
@@ -265,6 +384,104 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                    let app_id = share_mdata_req.app.id;
+                    let app_info = match config.get(&sha3_256(app_id.as_bytes())) {
+                        Some(app_info) => app_info,
+                        None => {
+                            return future::err(AuthError::Unexpected(String::from("malformed app_id"))).into_box();
```
nbaksalyar: We have a macro for `future::err(err).into_box()` too, so you can replace it with just `err!(err)`

safe_authenticator/src/ipc.rs
```diff
@@ -265,6 +384,104 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                resp.as_ptr()
+                            );
+                            Ok(())
+                        })
+                        .for_each(|()| Ok(()))
+                        .into_box()
+                })
+                .map_err(move |e| {
+                    let (error_code, description) = ffi_error!(e);
+                    o_cb(
```
nbaksalyar: You can use a macro for such simple errors that are returned to callbacks - e.g. `call_result_cb!(Err::<_, ()>(e), user_data, o_cb)`

safe_core/src/ipc/req/ffi.rs
```diff
@@ -167,3 +167,43 @@ impl Drop for ContainerPermissions {
...
+/// Represents a request to share mutable data
+pub struct ShareMDataReq {
+    /// Info about the app requesting shared access
+    pub app: AppExchangeInfo,
+    /// List of MD names & type tags and permissions that need to be shared
+    pub mdata: *const ShareMData,
+    /// Length of the mdata array
+    pub mdata_len: usize,
+    /// Allocated capacity of the mdata array
+    pub mdata_cap: usize,
```
nbaksalyar: We don't need to add a `cap` field, as a vector capacity is just a Rust stdlib implementation detail. We can ignore it in this case as we don't need to maintain ownership - the vector borrowed and copied by the front-end and the ownership is retained on our side

safe_core/src/ipc/req/ffi.rs
```diff
@@ -167,3 +167,43 @@ impl Drop for ContainerPermissions {
...
+    /// The mutable data name.
+    pub name: XorName,
+    /// The metadata key.
+    pub metadata_key: *const c_char,
+    /// The permissions being requested.
+    pub perms: *const PermissionModifier,
+    /// Lengths of the perms array
+    pub perms_len: usize,
+    /// Allocated capacity of the perm array
+    pub perms_cap: usize,
```
nbaksalyar: Same thing about `cap` as above

safe_authenticator/src/ipc.rs
```diff
@@ -215,26 +231,107 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                                            None => {
+                                                let (share_mdata_req_repr_c, keep_alive) = fry!(
+                                                    share_mdata_req.into_repr_c().map_err(AuthError::IpcError)
+                                                );
+                                                o_share_mdata(
+                                                        user_data_inner,
+                                                        req_id,
+                                                        &share_mdata_req_repr_c,
+                                                        metadatas.as_ptr());
+                                                drop(keep_alive);
```
nbaksalyar: I don't think you need to do an explicit `drop` here, as it will be dropped automatically after it goes out of the scope.

safe_authenticator/src/ipc.rs
```diff
@@ -265,6 +363,98 @@ pub unsafe extern "C" fn auth_decode_ipc_msg(auth: *const Authenticator,
...
+                    let app_id = share_mdata_req.app.id;
+                    let app_info = match config.get(&sha3_256(app_id.as_bytes())) {
+                        Some(app_info) => app_info,
+                        None => {
+                            return err!(AuthError::Unexpected(String::from("malformed app_id")));
+                        }
+                    };
+                    let user = User::Key(app_info.keys.sign_pk);
+                    let num_mdatas = share_mdata_req.mdata.len();
+                    stream::iter(share_mdata_req.mdata.into_iter().map(Ok))
```
nbaksalyar: Probably we can use `future::join_all()` instead of `stream::iter`? Not sure if that would work in this case.

safe_authenticator/src/tests.rs
```diff
@@ -778,6 +785,209 @@ fn lists_of_registered_and_revoked_apps() {
...
+    });
+
+    const NUM_MDATAS: usize = 3;
+
+    let mut mdatas = Vec::new();
+    for _ in 0..NUM_MDATAS {
+        let name = rand::random();
+        let mdata = {
+            let mut owners = BTreeSet::new();
+            owners.insert(user);
```
nbaksalyar: There's a handy macro for such cases, which you can use as `let owners = btree_set![user];`

safe_authenticator/src/tests.rs
```diff
@@ -778,6 +785,209 @@ fn lists_of_registered_and_revoked_apps() {
...
+    let encoded_msg = unwrap!(ipc::encode_msg(&msg, "safe-auth"));
+
+    let decoded = unwrap!(decode_ipc_msg(&authenticator, &encoded_msg));
+    match decoded {
+        (IpcMsg::Req { req: IpcReq::ShareMData(ShareMDataReq { mdata, .. }), .. }, Some(metadatas)) => {
+            assert_eq!(mdata, mdatas);
+            assert_eq!(metadatas, iter::repeat(Vec::<u8>::new()).take(NUM_MDATAS).collect::<Vec<_>>());
+        },
+        _ => panic!("Unexpected: {:?}", decoded),
+    };
```
nbaksalyar: It'd be good to check that the permissions for shared Mutable Data objects have been updated too (i.e. that the app's key has been added to the permissions list)

safe_authenticator/src/tests.rs
```diff
@@ -778,6 +785,209 @@ fn lists_of_registered_and_revoked_apps() {
...
+
+#[test]
+fn share_some_mdatas_with_ownership_error() {
+    let authenticator = create_account_and_login();
+
+    let user = run(&authenticator, move |client| {
+        client.public_signing_key().map_err(AuthError::CoreError)
+    });
+
+    let (someone_else, _) = rust_sodium::crypto::sign::gen_keypair();
```
nbaksalyar: Probably no need for a fully-qualified name, better to import `rust_sodium::crypto::sign` at the module level

safe_authenticator/src/ipc.rs
canndrew: True, but I want to explicitly make sure that it survives until that line of code. Without the `drop` it's possible someone could change the code in a way that breaks that condition.

safe_authenticator/src/ipc.rs
canndrew: That would work too. Do you want me to change it?

safe_authenticator/src/tests.rs
canndrew: Ah I didn't know about that :)
I was wishing that existed while I was writing this.

safe_authenticator/src/ipc.rs
nbaksalyar: @canndrew, yeah, I guess it'll speed up this operation a little bit (as these requests can be processed in parallel). Hopefully it's not a big change too.

safe_authenticator/src/tests.rs
nbaksalyar: @canndrew I wish this macro existed in Rust's std lib :) It's [defined in our own safe_core](https://github.com/maidsafe/safe_client_libs/blob/f642805c20450981b0600c6ad88e049a50b8b172/safe_core/src/utils/mod.rs#L31-L51), along with `btree_map!` too.

safe_authenticator/src/ipc.rs
canndrew: The `buffer_unordered` calls already make it run in parallel, but using `join_all` rather than that weird use of `for_each` would prolly be clearer.

safe_authenticator/src/tests.rs
canndrew: We could open a bug report for it? Though I think I've seen one before.

safe_core/src/ipc/req/ffi.rs
```diff
@@ -174,3 +174,54 @@ impl Drop for ContainerPermissions {
...
+/// For use in `ShareMDataReq`. Represents a specific `MutableData` that is being shared.
+pub struct ShareMData {
+    /// The mutable data type.
+    pub type_tag: u64,
+    /// The mutable data name.
+    pub name: XorName,
+    /// The metadata key.
+    pub metadata_key: *const c_char,
+    /// The permissions being requested.
+    pub perms: PermissionSet,
```
bochaco: I think we don't really need a new PermissionSet, and it would be better also for consistency to receive just a list of Permissions like the ones received for authorisation and containers permissions requests:
https://github.com/maidsafe/safe_client_libs/blob/master/safe_core/src/ipc/req/ffi.rs#L156

canndrew: What would the list imply? That every permission in it should be allowed and everything else denied?

bochaco: I guess so, I think that's how it works for the authorisation and container permissions requests.

nbaksalyar: @bochaco, actually it's a bit different permission set - authorisation and container perms refer to permissions stored in the access container, while this `PermissionSet` is about permissions stored in Mutable Data itself (e.g. compare them - the former has the `Read` permission while the latter doesn't). As we don't deal with the access container while sharing Mutable Data objects, I think this should be a separate structure (despite being very similar)

bochaco: Ok, from implementation side I guess you are right, but from API perspective, shouldn't we be consistent in the way he permissions are requested by just providing a list of permissions to set, e.g. ['Insert', 'Update'].

canndrew: On the Rust side the two kinds of requests use `PermissionSet` and `BTreeSet<Permission>`. A `PermissionSet` stores, for every permission, whether the permission should be allowed, denied, or unchanged whereas a `BTreeMap<Permission>` is just a set of permissions.

So if we wanted `ffi::ShareMData` to use a list, it would be a list of `(Permission, bool)` (or equivalent), not a list of `Permission`. That is, unless we wanted to change the Rust `ShareMData` to use a set of permissions rather than a `PermissionSet`.

nbaksalyar: @bochaco - @canndrew is right, it's still a bit different from the API perspective as well. You can think about it as `ManagePermissions` via a proxy (the authenticator/owner). In other words, it's still different from the containers IPC request, as here you should be able to _deny_ permissions too, not only 'allow' them.

bochaco: Why is that different from the containers? So you want more granularity for revoking permissions in MD but not in containers?

bochaco: My understanding is that this woukd be used by an app to request permission on a MD, in that use case I don't imagine an app requesting for permissions to be revoked. Now if what you are saying is that this will also be used by the authenticator to revoke permissions, I presume you want the same for containers so the authenticator can manage it at the permissions level. I am no taking about implementation details, I am talking that from the user's perspective there is an inconsistency and ut's not clear to me why there actually is, in fact, a container is a MD.

nbaksalyar: Yes @bochaco because containers are different beast. It's confusing a bit, but adding & revoking permissions through the access container works differently. It's high-level permissions management as opposed to low-level, Mutable Data, permissions, so in a general case apps should go the `IpcReq::Containers` way, but if they need a more fine-grained control, they can manage permissions through this new IPC, `ShareMData` (getting that MDataInfo out-of-band in the first place - which, again, is different to the containers perms management, because the access container also stores `MDataInfo` with encryption keys along with the permissions).

bochaco: So me as a user or app, why do I reqiest permissions for containers in a different way tjat for MDs? me as a user, I use the MD API to interact with a container so it is the same thing for me.

bochaco: You keep thinking implementation  wise, just think it from a user's perspective, as a black box, there is no difference in containers and MD, or there is no need to be for the user.

nbaksalyar: Yeah, but what I want to say is that it's different from the user's (app's) perspective as well. Because this `ShareMData` thing is similar to more low-level APIs that we have - e.g. you can technically skip the NFS part altogether and just do the same thing on the front-end by using the `immutable_data` module and all that stuff. We just provide a high-level interface to do the same, but technically it's the same thing. So, to answer this question:

> why do I reqiest permissions for containers in a different way tjat for MDs?

It's because of revocation. So remember that when you revoke an app's access, the authenticator actually re-encrypts data stored in the shared Mutable Data objects - and to do that, the authenticator should have access to the encryption keys, so they're stored in the access container itself, along with the permissions. _That's_ what an app should know from the API perspective - that if we're dealing with `IpcReq::Containers`, then the _authenticator_ is an owner and manager of those MD/containers.

Now, compare that to a different case, when an app creates an MD object on the network, completely bypassing the authenticator. The authenticator _does not know_ about this MD or its encryption keys - it's an app's responsibility, as now _the app_ is technically an "owner" of that MD (in the sense that an app manages this MD by itself). So how do you suppose to revoke the app's access here API-wise? Because the authenticator does not know and does not care about that MD, it only knows that de-facto the user is always the owner, therefore a user _always_ can manage MD permissions, adding or revoking them - so that's why this request should go through the authenticator, with a different API, and with a different struct to pass permission sets (because otherwise we could have just used that `IpcReq::Containers` thing). Does it make sense?

PR #520, opened 2017-07-31
test/nfs: test nfs functions after deleting then updating a file

m-cat:


nbaksalyar:
Looks neat, great work! :+1:

PR #526, opened 2017-08-04
test/login: add tests for logging in with low balance (MAID-2140)

m-cat:


safe_core/src/client/mod.rs
```diff
@@ -554,6 +577,33 @@ impl<T: 'static> Client<T> {
         }))
     }
 
+
+    #[cfg(any(all(test, feature = "use-mock-routing"),
+                all(feature = "testing", feature = "use-mock-routing")))]
+    /// Allows customising the mock Routing client before logging in using client keys
```
nbaksalyar: Same thing about the comment - it should go above `#[cfg(...)]`

safe_authenticator/src/tests.rs
```diff
@@ -681,6 +681,72 @@ fn containers_access_request() {
     compare_access_container_entries(&authenticator, app_sign_pk, access_container, expected);
 }
 
+// Ensure that users can log in with low account balance.
+#[test]
+fn login_with_low_balance() {
```
nbaksalyar: Please add `#[cfg(feature = "use-mock-routing")]` above this function and make sure this test suite builds without the `use-mock-routing` feature flag too (i.e. it should be possible to run tests against the real network, and the real network Routing doesn't have an option to override responses).

safe_core/src/client/mock/tests.rs
```diff
@@ -964,12 +964,31 @@ fn auth_keys() {
...
+    ));
+    let nae_mgr = Authority::NaeManager(*data.name());
+
+    let msg_id = MessageId::new();
+    unwrap!(routing.put_mdata(client_mgr, data, msg_id, owner_key));
+    expect_success!(routing_rx, msg_id, Response::PutMData);
+
     // Exhaust the account balance.
     loop {
         let data = ImmutableData::new(unwrap!(utils::generate_random_vector(10)));
```
nbaksalyar: Create `data` outside of the loop

PR #527, opened 2017-08-07
test/nfs: ensure file created/modified timestamps are set correctly

m-cat:


safe_app/src/ffi/nfs.rs
```diff
@@ -591,6 +599,10 @@ mod tests {
 
         let f: *const File = unsafe { unwrap!(call_1(|ud, cb| file_close(&app, read_h, ud, cb))) };
         assert!(f.is_null());
+
+        // Check timestamps again after append and update
+        assert_eq!(created_time, *written_file.created_time());
+        assert!(created_time <= *written_file.modified_time());
```
nbaksalyar: It would be better to compare this value with the previous `modified_time()` (i.e. store it as `let old_modified = *written_file.modified_time()` before writing to file again)

safe_app/src/ffi/nfs.rs
```diff
@@ -483,14 +485,16 @@ mod tests {
...
+
         // Insert file into container.
         unsafe {
             unwrap!(call_0(|ud, cb| {
                 dir_insert_file(
                     &app,
                     container_info_h,
                     ffi_file_name1.as_ptr(),
-                    &written_file.into_repr_c(),
+                    &written_file.clone().into_repr_c(),
```
nbaksalyar: Why `clone()`? Probably all relevant values can be stored in variables before this value is moved here.

PR #528, opened 2017-08-08
MAID-2203 revocation improvements

madadam:


safe_authenticator/src/tests.rs
```diff
@@ -1409,6 +1410,229 @@ fn app_authentication_recovery() {
...
+        app_container: false,
+        containers: create_containers_req(),
+    };
+
+    let _ = unwrap!(register_app(&auth, &auth_req));
+    let app_id_1 = auth_req.app.id.clone();
+
+    // Simulate network failure that prevents any subsequent revocation from
+    // successful completion.
+    let auth = unwrap!(Authenticator::login_with_hook(
```
nbaksalyar: Probably would be better to move this entire block into a separate helper function (e.g. `simulate_revocation_failure`)? I believe it's used in the exact same way in the revocation test, so it'd be nice if we could get rid of code duplication and improve tests clarity.

PR #529, opened 2017-08-08
Add ffi encoding/decoding functions for ShareMData requests

canndrew:
Jira task MAID-2224. This builds on the earlier #519 work.

Also depends on https://github.com/maidsafe/routing/pull/1505


canndrew:
This is now blocked on https://github.com/maidsafe/routing/pull/1507 being published.

PR #532, opened 2017-08-10
Maid 2209 - fix app dedicated container creation

m-cat:


safe_authenticator/src/app_auth.rs
```diff
@@ -75,7 +76,70 @@ pub fn app_state(client: &Client<()>, apps: &Apps, app_id: String) -> Box<AuthFu
...
+                            CoreError::RoutingClientError(
+                                ClientError::NoSuchEntry))) => 0,
+                    // Error has occurred while trying to get an existing entry
+                    Err(e) => return Err(e),
+                };
+                Ok((version, app_info, app_keys, dir, permissions))
+            })
+        })
+        .and_then(move |(version, app_info, app_keys, dir, perms)| {
+            put_access_container_entry(&c4, &dir, &app_info.id, &app_keys, &perms, version)
```
madadam: Nitpick, but I'd either rename `perms` to `permissions` here, or rename `permissions` to `perms` on line 132 - the code would be slightly easier to follow.

safe_authenticator/src/app_auth.rs
```diff
@@ -110,13 +174,13 @@ pub fn authenticate(client: &Client<()>, auth_req: AuthReq) -> Box<AuthFuture<Au
                         keys: keys,
                     };
                     config::insert_app(&c3, app.clone())
-                        .map(move |_| (app, app_state))
+                        .map(move |_| (app, app_state, app_id))
                         .into_box()
                 }
                 AppState::Authenticated | AppState::Revoked => {
-                    let app_entry_name = sha3_256(app_id.as_bytes());
+                    let app_entry_name = sha3_256(app_id.clone().as_bytes());
```
madadam: Is `clone` necessary here? `as_bytes` isn't consuming `self`, is it?

safe_app/src/lib.rs
```diff
@@ -662,4 +647,80 @@ mod tests {
...
+
+    // Test app container creation under the following circumstances:
+    // 1. An app is authorised for the first time with `app_container` set to `true`.
+    // 2. If an app is authorised for the first time with `app_container` set to `false`,
+    // then any subsequent authorisation with `app_container` set to `true` should trigger
+    // the creation of the app's own container.
+    // 3. Make sure that the app's own container is also created when it's re-authorised
+    // with `app_container` set to `true` after it's been revoked.
+    #[test]
+    fn app_container_creation() {
```
madadam: Is this test supposed to be runnable even without `use-mock-rouitng`?

safe_authenticator/src/app_auth.rs
```diff
@@ -75,7 +76,70 @@ pub fn app_state(client: &Client<()>, apps: &Apps, app_id: String) -> Box<AuthFu
...
+    client: &Client<()>,
+    app: &AppInfo,
+    permissions: AccessContainerEntry,
+) -> Box<AuthFuture<AccessContainerEntry>> {
+    let sign_pk = app.keys.sign_pk;
+    let app_info = app.info.clone();
+    let app_id = app_info.id.clone();
+
+    app_container::fetch(client.clone(), app_id, sign_pk)
+        .map(move |mdata_info| (mdata_info, permissions))
```
madadam: This line is not necessary, you can simply capture `permissions` by the closure in the following `and_then` block.

safe_authenticator/src/app_auth.rs
```diff
@@ -75,7 +76,70 @@ pub fn app_state(client: &Client<()>, apps: &Apps, app_id: String) -> Box<AuthFu
     }
 }
 
-/// Authenticates an app request
+/// Create app container and update access container.
+fn create_or_update_app_container(
```
madadam: This is just a suggestion and feel free to ignore if you disagree, but I'd prefer to split this function into two:
1. just `app_container::fetch`
2. something like this:
```rust
fn insert_app_container(permissions: &mut AccessContainerEntry, app_id: &str, app_container_info: MDataInfo) {
    let access = btree_set![Permission::Read, ...]; // all the permissions.
    let _ = permissions.insert(format!("apps/{}", app_id), (app_container_info, access));
}
```
because currently it feels like two concerns mixed into one function.

safe_authenticator/src/app_auth.rs
```diff
@@ -155,7 +207,51 @@ pub fn authenticate(client: &Client<()>, auth_req: AuthReq) -> Box<AuthFuture<Au
...
+    let bootstrap_config = fry!(Client::<()>::bootstrap_config());
+
+    // Check whether we need to create/update dedicated container
+    if app_container {
+        access_container(&c2)
+            .and_then(move |dir| {
+                access_container_entry(&c3, &dir, &app_id, app_keys.clone())
+            })
+            .and_then(move |(_, perms)| {
+                let perms = unwrap!(perms);
```
madadam: We should avoid `unwrap` in production code as much as possible. In this case, perhaps use something like this instead:
```rust
let perms = perms.unwrap_or_else(AccessContainerEntry::default);
```
(`None` would mean that the access container entry has been deleted, so using a empty one seems like decent approach in that case)

safe_authenticator/src/app_auth.rs
m-cat: Ah my bad -- that naming scheme is a survivor from before the refactor.

safe_app/src/lib.rs
m-cat: I believe so. This test doesn't use any `use-mock-routing`-only functionality, as far as I can tell.

safe_authenticator/src/app_auth.rs
ustulation: good spot

safe_app/src/lib.rs
madadam: If it compiles without `use-mock-routing` then it's probably good. :+1:

safe_app/src/lib.rs
```diff
@@ -662,4 +647,80 @@ mod tests {
...
+
+        assert_eq!(num_containers(&app), 0); // should be empty
+
+        // Re-authorise the app with `app_container` set to `true`.
+        let app = authorise_app(&auth, &app_info, &app_id, true);
+
+        assert_eq!(num_containers(&app), 1); // should only contain app container
+
+        // Revoke the app
+        let _ = unwrap!(try_revoke(&auth, &app_id));
```
nbaksalyar: Should have these lines above (or, alternatively, add an extra test case for it):
```
let app_info = gen_app_exchange_info();
let app_id = app_info.id.clone();
let app = authorise_app(&auth, &app_info, &app_id, false);
```

We need to test that an app's own container is created even if the app was revoked because the app's keys remain in the config dir even after the revocation, and re-authorising a revoked app is a special case.

safe_authenticator/src/app_auth.rs
```diff
@@ -155,7 +197,57 @@ pub fn authenticate(client: &Client<()>, auth_req: AuthReq) -> Box<AuthFuture<Au
...
+            })
+            .and_then(move |(mdata_info, perms, app_id)| {
+                insert_app_container(perms, &app_id, mdata_info).and_then(
+                    move |perms| update_access_container(&c5, &app, perms),
+                )
+            })
+            .into_box()
+    } else {
+        access_container(&c4)
+    }.and_then(move |dir| {
```
nbaksalyar: This looks a bit confusing. Probably would be better to store the result of the `if` expression in a variable, like `let app_cont_fut = if app_container { .. } else { .. }` and then use it as `app_cont_fut.and_then(move |dir| ...)`

safe_authenticator/src/app_auth.rs
```diff
@@ -155,7 +197,57 @@ pub fn authenticate(client: &Client<()>, auth_req: AuthReq) -> Box<AuthFuture<Au
...
+    let c5 = client.clone();
+
+    let app_keys = app.keys.clone();
+    let app_keys_auth = app.keys.clone();
+    let sign_pk = app.keys.sign_pk;
+    let bootstrap_config = fry!(Client::<()>::bootstrap_config());
+
+    // Check whether we need to create/update dedicated container
+    if app_container {
+        access_container(&c2)
```
nbaksalyar: Doesn't need to be `&c2`, just use `&client` here.

safe_authenticator/src/app_auth.rs
```diff
@@ -155,7 +197,57 @@ pub fn authenticate(client: &Client<()>, auth_req: AuthReq) -> Box<AuthFuture<Au
...
+                )
+            })
+            .and_then(move |(mdata_info, perms, app_id)| {
+                insert_app_container(perms, &app_id, mdata_info).and_then(
+                    move |perms| update_access_container(&c5, &app, perms),
+                )
+            })
+            .into_box()
+    } else {
+        access_container(&c4)
```
nbaksalyar: Same about `&c4` -> `&client`

safe_authenticator/src/app_auth.rs
```diff
@@ -172,70 +264,42 @@ fn authenticate_new_app(
...
+        .and_then(move |(perms, sign_pk)| {
+            if app_container {
+                app_container::fetch(c4, app_id.clone(), sign_pk)
+                    .and_then(move |mdata_info| {
+                        insert_app_container(perms, &app_id, mdata_info)
+                    })
+                    .into_box()
+            } else {
+                ok!(perms)
+            }.map(move |perms| (perms, app))
```
nbaksalyar: Just use `ok!((perms, app))` in the else branch and move the `map` call to the `if app_container { .. }` branch.

safe_app/src/lib.rs
nbaksalyar: Also, IIRC instead of `unwrap!(try_revoke(...))` you can just use `revoke(...)`

madadam: We might also consider removing `revoke`, since it's literally just `unwrap!(try_revoke(...))`.

PR #533, opened 2017-08-10
chore/update-deps: update to be compatible with routing master

canndrew:
This allows safe_client_libs to build without the `routing::Event::ProxyRateLimitExceeded` event which is no longer emitted by routing.


safe_core/src/client/mod.rs
```diff
@@ -1540,22 +1494,17 @@ fn create_empty_dir(
...
-        }
-    }
+    let msg_id = MessageId::new();
+    routing
+        .put_mdata(dst, dir_md.clone(), msg_id, owner_key)
+        .map_err(CoreError::from)
+        .and_then(|_| {
+            wait_for_response!(routing_rx, Response::PutMData, msg_id)
+        })
+        .map_err(|e| {
```
nbaksalyar: I think this should be `.or_else` - notice the part about ops recovery & `ClientError::DataExists` above.

PR #535, opened 2017-08-10
MAID-2226: improve share mdata

madadam:


safe_app/src/ffi/mutable_data/metadata.rs
```diff
@@ -0,0 +1,56 @@
...
+        let metadata = UserMetadata::clone_from_repr_c(metadata)?;
+        let encoded = serialise(&metadata)?;
+        o_cb(user_data, FFI_RESULT_OK, encoded.as_ptr(), encoded.len());
+        Ok(())
+    })
+}
+
+/// Get pointer and length of the key under which the metadata should be stored.
+#[no_mangle]
+pub unsafe extern "C" fn mdata_metadata_key(
```
nbaksalyar: I think there's no need for a function and probably `METADATA_KEY` can be just exposed as `#[no_mangle] pub static METADATA_KEY: *const c_char = "_meta_data"`. It could also be a null-terminated string instead of a byte array, as it's not intended to be encrypted.

safe_core/src/ipc/resp/ffi.rs
```diff
@@ -105,12 +108,37 @@ pub struct AccessContInfo {
...
+    fn drop(&mut self) {
+        unsafe {
+            if !self.name.is_null() {
+                let _ = CString::from_raw(self.name as *mut _);
+                self.name = ptr::null();
+            }
+
+            if !self.description.is_null() {
+                let _ = CString::from_raw(self.description as *mut _);
+                self.description = ptr::null();
```
nbaksalyar: I don't think you need this and `self.name = ptr::null()` - these vars will be dropped anyway.

safe_authenticator/src/revocation.rs
```diff
@@ -436,28 +433,19 @@ fn reencrypt_containers(
...
-                    }
+                    let plain_key = mdata_info.decrypt(&old_key)?;
+                    let new_key = mdata_info.enc_entry_key(&plain_key)?;
+
+                    let plain_content = mdata_info.decrypt(&value.content)?;
+                    let new_content = mdata_info.enc_entry_value(&plain_content)?;
+
+                    // Delete the old entry with the old key and
+                    // insert the re-encrypted entry with a new key
+                    actions = actions.del(old_key, value.entry_version + 1).ins(
```
nbaksalyar: Would this work if the `old_key` and the `new_key` are the same?
E.g. when a previous re-encryption attempt failed but the `new_enc_info` in MDataInfo has been replaced.
If I understand it correctly, then the `old_key` will be decrypted successfully using the new_enc_info, and then it will be re-encrypted to the exact same value with `enc_entry_key`. Then how it will work with `del` + `ins`, if both refer to the same key?

safe_app/src/ffi/mutable_data/metadata.rs
madadam: Problem with this is that it might lead to subtle errors. If `METADATA_KEY` was a C-style nul terminated string, then someone on the frontend might be tempted to get it's length by calling `strlen` or something like that, but that function excludes the trailing nul byte as far as I know. So the length would be incorrect.

madadam: So maybe we can make it not nul-terminated, but expose another constant `METADATA_KEY_LEN`?

safe_authenticator/src/revocation.rs
madadam: It would still work, but only because we are using `recovery::mutate_mdata_entries`. We might want to add a test for this though.

safe_core/src/ipc/resp/ffi.rs
madadam: :+1:

safe_app/src/ffi/mutable_data/metadata.rs
nbaksalyar: >expose another constant METADATA_KEY_LEN
Yes, sounds reasonable.
Because this is a constant value, it'd be better & more efficient to have it exposed as such.

PR #545, opened 2017-08-18
New API in Authenticator to fetch a list of apps accessing arbitrary MD (MAID-2247)

m-cat:
1. Created type definitions for FFI arrays and refactored out common ffi types.
2. Created a new API to fetch a list of apps that have access to an arbitrary MD.

### DIFF
--- a/safe_core/src/ipc/resp/ffi.rs
+++ b/safe_core/src/ipc/resp/ffi.rs
@@ -68,7 +68,7 @@ pub struct AppKeys {
     /// Asymmetric sign private key.
     pub sign_sk: SignSecretKey,
     /// Asymmetric enc public key.
-    pub enc_pk: PublicKey,
+    pub enc_pk: AsymPublicKey,
     /// Asymmetric enc private key.
     pub enc_sk: AsymSecretKey,
 }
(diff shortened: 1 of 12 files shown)


safe_core/src/ipc/resp/mod.rs
```diff
@@ -258,33 +262,94 @@ pub fn access_container_enc_key(
...
+        Ok(ffi::MetadataResponse {
+            name: match self.name {
+                Some(name) => CString::new(name)?.into_raw(),
+                None => ptr::null(),
+            },
+            description: match self.description {
+                Some(description) => CString::new(description)?.into_raw(),
+                None => ptr::null(),
+            },
+            xor_name: xor_name.0.as_ptr() as *const [u8; 32],
```
nbaksalyar: Can be just `&xor_name.0`, I think (Rust references are converted into pointers automatically)

safe_core/src/ipc/resp/ffi.rs
```diff
@@ -108,26 +109,48 @@ pub struct AccessContInfo {
...
+pub struct AppAccess {
+    /// App's or user's public key
+    pub sign_key: *const SignPublicKey,
+    /// A list of permissions
+    pub permissions: FfiPermissionSet,
+    /// App's user-facing name
+    pub name: *const c_char,
+    /// App id.
+    /// This is u8, as the app-id can contain non-printable characters.
+    pub id: *const c_uchar,
```
nbaksalyar: Would be better to name it `app_id`

PR #550, opened 2017-08-23
Fix safe_authenticator build with real routing (MAID-2274)

nbaksalyar:


.travis.yml
```diff
@@ -74,9 +74,9 @@ script:
       (
         set -x;
         cd ffi_utils && cargo clippy && cargo clippy --profile=test &&
-        cd ../safe_core && cargo clippy && cargo clippy --profile=test --features=$FEATURES &&
-        cd ../safe_authenticator && cargo clippy --profile=test --features=$FEATURES &&
-        cd ../safe_app && cargo clippy --profile=test --features=$FEATURES;
+        cd ../safe_core && cargo clippy && cargo clippy --profile=test && cargo clippy --profile=test --features=$FEATURES &&
```
ustulation: ah no, the reason why this was orchestrated such was so that things are not rebuilt everytime. This is now the worst case. So what needs to be done is - `cargo clippy --profile=test` for all 3 crates (core, auth and app in that order) followed by `cargo clippy --profile=test --features=$FEATURES` for all 3 etc so that build of safe-auth is not triggering complete rebuild of core and then safe-app triggering complete rebuild of both core and safe-auth etc.

PR #555, opened 2017-08-24
fix/mock-vault: fix mock vault write mode

nbaksalyar:
When mock-vault saved the changes to disk it didn't truncate the previous file contents. This could cause different hard-to-catch bugs, including the one reported in MAID-2262.


nbaksalyar:
@ustulation, yes, I tested it on the email-app with MAID-2262 in mind - works correctly.

PR #567, opened 2017-09-05
refactor/all: refactor ffi modules, generate c headers with cheddar

m-cat:
- refactor FFI modules in each crate into a single module
- add build scripts utilizing moz_cheddar to generate C FFI header files


include/safe_app.h
```diff
@@ -0,0 +1,59 @@
...
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include <safe_app/access_container.h>
```
m-cat: Should be `#include "foo.h"` rather than `<foo.h>` -- already fixed this on my branch.

safe_app/src/ffi/mutable_data/permissions.rs
```diff
@@ -32,7 +32,7 @@ use std::os::raw::c_void;
 pub static USER_ANYONE: u64 = 0;
 
 /// Permission actions.
-#[repr(C)]
+#[repr(u8)]
```
m-cat: These should revert back to `repr(C)` as cheddar can't handle any other ABI constraint and these enums are not appearing in the header files.

ffi_utils/src/header_gen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    Ok(())
+}
+
+// Replace occurrences of "c_char" and "c_void" in the file with "char" and "void".
+// Remove this if it ever gets fixed in cheddar.
+fn file_contents_replace(fname: &str) -> io::Result<()> {
+    let contents = read_file_str(fname)?;
+
+    let contents = contents.replace("c_char", "char");
```
madadam: This would replace all occurrences of "c_char" and "c_void", even if they are part of a longer identifier. It would be safer to use whole-word replace for this. On the other hand, we probably don't have such identifiers anywhere in out codebase and bringing in regular expression just to cover this might be overkill. I'll leave this up to you to decide.

safe_app/build.rs
```diff
@@ -0,0 +1,49 @@
...
+// relating to use of the SAFE Network Software.
+
+//! Build script for generating C header files from FFI modules.
+
+extern crate ffi_utils;
+#[macro_use]
+extern crate unwrap;
+
+static MODULES: [&'static str; 13] = [
+    "ffi::access_container",
```
madadam: Consider auto-generating this list (by recursively traversing the "ffi" folder perhaps?), so we don't have to keep updating it in mind when we change something.

safe_app/src/ffi/crypto.rs
```diff
@@ -22,7 +22,7 @@ use ffi_utils::{FFI_RESULT_OK, FfiResult, OpaqueCtx, catch_unwind_cb};
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use object_cache::{EncryptPubKeyHandle, EncryptSecKeyHandle, SignKeyHandle};
 use rust_sodium::crypto::{box_, sealedbox, sign};
-use safe_core::ffi::{AsymNonce, AsymPublicKey, AsymSecretKey};
+use safe_core::arrays::{AsymNonce, AsymPublicKey, AsymSecretKey};
```
madadam: Should the `arrays` module be part of `ffi` (`safe_core::ffi::arrays`)?

safe_authenticator/build.rs
```diff
@@ -0,0 +1,35 @@
...
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! Build script for generating C header files from FFI modules.
+
+extern crate ffi_utils;
+#[macro_use]
+extern crate unwrap;
+
+static MODULES: [&'static str; 3] = ["ffi::apps", "ffi::ipc", "ffi::logging"];
```
madadam: Again, try to auto-generate these.

safe_core/build.rs
```diff
@@ -0,0 +1,35 @@
...
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! Build script for generating C header files from FFI modules.
+
+extern crate ffi_utils;
+#[macro_use]
+extern crate unwrap;
+
+static MODULES: [&'static str; 3] = ["ffi::nfs", "ffi::ipc::req", "ffi::ipc::resp"];
```
madadam: ...same here: try to auto-generate these.

safe_app/build.rs
m-cat: I wanted to do something like this (just look at what is `pub use`d in lib.rs and include it if it starts with `ffi::`). I was just wondering if we are going to change anything when Rust’s new module system comes out (not sure if that’s anytime soon, but now that I think about it we can just write a task to change how we detect the ffi modules later)

ffi_utils/src/header_gen.rs
```diff
@@ -0,0 +1,159 @@
...
+// Parse given root file (such as lib.rs) and return all `pub use`d modules starting with `ffi::`.
+fn parse_root(fname: &str) -> io::Result<Vec<String>> {
+    let contents = read_file_str(fname)?;
+    let mut modules = Vec::new();
+    let mut found = false;
+    let re_pub_use = unwrap!(regex::Regex::new(r"^pub use ffi::.*\*;$"));
+    let re_module = unwrap!(regex::Regex::new(r"ffi::.*\*"));
+
+    for line in contents.lines() {
+        if re_pub_use.is_match(line) {
```
nbaksalyar: Probably it would have been better to rewrite this entire section using [`Regex::captures`](https://doc.rust-lang.org/regex/regex/struct.Regex.html#method.captures).
E.g.:
```
let re_pub_use = unwrap!(regex::Regex::new(r"^pub use ffi::(.*)\*;$"));
for line in contents.lines() {
  let captures = re_pub_use.captures(line);
  // captures.get(1) will contain the module part, and you won't need to slice the string.
}
```

safe_app/src/ffi/crypto.rs
```diff
@@ -22,7 +22,7 @@ use ffi_utils::{FFI_RESULT_OK, FfiResult, OpaqueCtx, catch_unwind_cb, vec_clone_
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use object_cache::{EncryptPubKeyHandle, EncryptSecKeyHandle, SignKeyHandle};
 use rust_sodium::crypto::{box_, sealedbox, sign};
-use safe_core::ffi::{AsymNonce, AsymPublicKey, AsymSecretKey};
+use safe_core::ffi::arrays::*;
```
nbaksalyar: Better to use fully-qualified names in such cases because sometimes it's hard to tell which names are imported from a module.
So probably better to leave it as
```
use safe_core::ffi::arrays::{AsymNonce, AsymPublicKey, AsymSecretKey};
```

PR #572, opened 2017-09-14
MAID-1612: Do not clone secret keys

madadam:
Replace all secret keys (`rust_sodium::crypto::box_::SecretKey`, `rust_sodium::crypto::secretbox::Key` and `rust_sodium::crypto::sign::SecretKey`) with drop-in equivalents that implement "secure cloning" - they don't actually clone the underlying data but instead implicitly share it. This is supposed to prevent some attack vectors.


nbaksalyar:
Superseded by #574

PR #575, opened 2017-09-18
MAID-2349 + MAID-2352: Improve account creation error reporting + more tests for NFS

madadam:


safe_authenticator/src/lib.rs
```diff
@@ -171,27 +171,26 @@ impl Authenticator {
...
+            let client = try_tx!(create_client_fn(el_h, core_tx.clone(), net_tx), tx);
 
             unwrap!(core_tx.unbounded_send(CoreMsg::new(move |client, &()| {
                 std_dirs::create(client)
-                    .map(move |()| {
-                        unwrap!(tx.send(Ok(core_tx2)));
-                    })
-                    .map_err(move |e| {
-                        unwrap!(tx2.send(Err((Some(core_tx3), AuthError::from(e)))));
+                    .map_err(|_| AuthError::AccountContainersCreation)
```
nbaksalyar: Probably it'd be worth to include the underlying error info as well.
So I guess change the error type definition to `AccountContainersCreation(String)` and then create it as
```
.map_err(|e| {
  AuthError::AccountContainersCreation(format!("{}", e))
})
```
then, for `Display` impl something like:
```
AuthError::AccountContainersCreation(reason) => {
  write!(
    formatter,
    "Account containers creation error due to {:?}. Login to attempt recovery.",
    reason
  )
}
```

PR #586, opened 2017-09-21
refactor/app: remove MDataPermissionSetHandle and related functions

m-cat:
Also fixes a bug with handles beginning at 0.

### DIFF
--- a/safe_app/src/ffi/mutable_data/permissions.rs
+++ b/safe_app/src/ffi/mutable_data/permissions.rs
@@ -21,8 +21,10 @@ use App;
 use errors::AppError;
 use ffi::helper::send_sync;
 use ffi::mutable_data::helper;
-use ffi_utils::{FfiResult, OpaqueCtx, catch_unwind_cb};
-use object_cache::{MDataPermissionsHandle, SignKeyHandle};
+use ffi_utils::{FfiResult, catch_unwind_cb};
+use ffi_utils::SafePtr;
+use object_cache::{MDataPermissionsHandle, NULL_OBJECT_HANDLE, SignKeyHandle};
+use permissions;
 use routing::{Action, User};
 use safe_core::ffi::ipc::req::PermissionSet as FfiPermissionSet;
 use safe_core::ipc::req::{permission_set_clone_from_repr_c, permission_set_into_repr_c};
@@ -30,7 +32,7 @@ use std::os::raw::c_void;
 
 /// Special value that represents `User::Anyone` in permission sets.
 #[no_mangle]
-pub static USER_ANYONE: u64 = 0;
+pub static USER_ANYONE: u64 = NULL_OBJECT_HANDLE;
 
 /// Permission actions.
 #[repr(C)]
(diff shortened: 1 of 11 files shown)


safe_app/src/object_cache.rs
```diff
@@ -270,7 +258,7 @@ struct HandleGenerator(Cell<ObjectHandle>);
 
 impl HandleGenerator {
     fn new() -> Self {
-        HandleGenerator(Cell::new(u64::MAX))
+        HandleGenerator(Cell::new(0))
```
madadam: Consider defining a public constant, say `NULL_OBJECT_HANDLE` and use it here instead of `0`. Then also use it everywhere where we previously used `0` to signify special / null handles.

safe_app/src/ffi/mutable_data/permissions.rs
```diff
@@ -266,20 +130,24 @@ pub unsafe extern "C" fn mdata_permissions_for_each(
     user_data: *mut c_void,
     o_each_cb: extern "C" fn(user_data: *mut c_void,
                              sign_key_h: SignKeyHandle,
-                             perm_set_h: MDataPermissionSetHandle),
+                             perm_set: FfiPermissionSet),
```
nbaksalyar: I think it's better to rename `mdata_permissions_for_each` to `mdata_list_permissions` and just return the results as `perm_sets: *const FfiPermissionSet, len: usize`, similarly to the other replaced `for_each` functions

safe_app/src/ffi/mutable_data/permissions.rs
```diff
@@ -311,7 +179,7 @@ pub unsafe extern "C" fn mdata_permissions_insert(
             let mut permissions = context.object_cache().get_mdata_permissions(permissions_h)?;
             let _ = permissions.insert(
                 helper::get_user(context.object_cache(), user_h)?,
-                permission_set_h,
+                unwrap!(permission_set_clone_from_repr_c(&permission_set)),
```
nbaksalyar: We generally don't use `unwrap!` in production code (except for locking mutexes and tests).
Pls replace this with `permission_set_clone_from_repr_c(...)?`

safe_app/src/ffi/mutable_data/tests.rs
```diff
@@ -92,86 +95,13 @@ fn permissions_crud_ffi() {
...
-                    ud,
-                    cb,
-                )
-            }))
-        };
-        assert_eq!(permission_value, PermissionValue::Allowed);
-    }
+    assert_eq!(Some(true), perm_set.is_allowed(Action::Insert));
+    assert_eq!(Some(true), perm_set.is_allowed(Action::ManagePermissions));
+    assert_eq!(None, perm_set.is_allowed(Action::Update));
```
nbaksalyar: If the `PermissionSet` struct is tested somewhere else (I think in `routing` at least?), there'd be no need to double-check that it works - so I'd suggest removing these 3 lines

safe_app/src/ffi/mutable_data/tests.rs
```diff
@@ -372,22 +254,24 @@ fn entries_crud_ffi() {
...
-    let perm_set_h: MDataPermissionSetHandle =
-        unsafe { unwrap!(call_1(|ud, cb| mdata_permission_set_new(&app, ud, cb))) };
+    let perm_set = PermissionSet::new().allow(Action::Insert);
 
-    unsafe {
-        unwrap!(call_0(|ud, cb| {
-            mdata_permission_set_allow(&app, perm_set_h, MDataAction::Insert, ud, cb)
-        }))
-    };
+    assert_eq!(Some(true), perm_set.is_allowed(Action::Insert));
```
nbaksalyar: Same here for testing the basic `PermissionSet` functionality

nbaksalyar:
Integrated into #589

PR #588, opened 2017-09-22
refactor/ffi: remove MDataKeysHandle/MDataValuesHandle and related functions

m-cat:
Add callback handling for vectors of vectors of u8.
Add `mdata_get_all_keys` function.
Add `mdata_get_all_values` function.

### DIFF
--- a/safe_app/src/ffi/mutable_data/mod.rs
+++ b/safe_app/src/ffi/mutable_data/mod.rs
@@ -34,8 +34,9 @@ use object_cache::{MDataEntriesHandle, MDataEntryActionsHandle, MDataPermissionS
 use routing::MutableData;
 use safe_core::{CoreError, FutureExt, MDataInfo};
 use safe_core::ffi::MDataInfo as FfiMDataInfo;
-use safe_core::ffi::ipc::resp::Value as FfiValue;
-use safe_core::ipc::resp::Value;
+use safe_core::ffi::ipc::resp::MDataKey as FfiMDataKey;
+use safe_core::ffi::ipc::resp::MDataValue as FfiMDataValue;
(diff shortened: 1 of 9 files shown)


safe_core/src/ipc/resp.rs
```diff
@@ -421,6 +422,60 @@ impl ReprC for UserMetadata {
...
+        ffi::Value {
+            content_ptr: self.content.as_ptr(),
+            content_len: self.content.len(),
+            entry_version: self.entry_version,
+        }
+    }
+
+    /// Converts the `Value` into its routing representation.
+    pub fn to_routing(&self) -> RoutingValue {
+        let Self {
```
m-cat: Can remove this and use `self.content` and `self.entry_version` below.

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -239,57 +240,72 @@ pub unsafe extern "C" fn mdata_list_entries(
...
-pub unsafe extern "C" fn mdata_list_keys(
+pub unsafe extern "C" fn mdata_get_all_keys(
     app: *const App,
     info: *const FfiMDataInfo,
     user_data: *mut c_void,
     o_cb: extern "C" fn(user_data: *mut c_void,
                         result: FfiResult,
-                        keys_h: MDataKeysHandle),
+                        keys: *const *const u8,
+                        key_lens: *const usize,
```
nbaksalyar: This signature looks complicated. Probably better to just have `struct MDataKey { val: *const u8, len: usize }`? Or more generic one (IIRC we had something similar to this struct).

nbaksalyar:
Integrated into #589

PR #590, opened 2017-09-26
[alpha-2] Make concurrent tests deterministic and fix rare bugs in app revocation

madadam:


nbaksalyar:
The `alpha-2` branch is deprecated and soon will be removed.

PR #591, opened 2017-09-27
feat/mock: use config files instead of env var, add option for unlimited mutations in mock

m-cat:


safe_core/src/lib.rs
```diff
@@ -15,7 +15,21 @@
...
+//!
+//! # Environment variables
+//!
+//! Switch off mutations limit on mock-vault.
+//!
+//! ```rust,ignore
+//! SAFE_MOCK_UNLIMITED_MUTATIONS
+//! ```
+//!
+//! Use memory store.
```
ustulation: This is very short and confusing description - a new person looking at this will have no idea what it means and is likely to skip it. So pls expand what this is all about. Also check other new documentation introduced here and elsewhere. Wherever possible pls make it abundantly clear and worthy of being a documentation, else it just looks like an afterthought or a placeholder.

safe_core/src/client/mock/tests.rs
```diff
@@ -1098,7 +1100,16 @@ fn auth_keys() {
 
 // Exhaust the account balance and ensure that mutations fail.
 #[test]
-fn balance_check() {
+fn low_balance_check() {
+    // Turn off unlimited mutations.
+    let env_var = if let Ok(val) = env::var(SAFE_MOCK_UNLIMITED_MUTATIONS) {
```
Fraser999: Wouldn't it be safer to use RAII here; make `env_var` a struct with a custom `drop` implementation which resets `SAFE_MOCK_UNLIMITED_MUTATIONS`?  Then if the test fails after we call `env::set_var(SAFE_MOCK_UNLIMITED_MUTATIONS, "1");` it still resets the environment.

PR #601, opened 2017-10-25
[alpha-2] Add config file feature to alpha-2

m-cat:


nbaksalyar:
The `alpha-2` branch is deprecated and soon will be removed.

PR #604, opened 2017-10-26
Serialisation compatibility tests (master)

madadam:


safe_core/src/client/mock/vault.rs
```diff
@@ -48,13 +47,13 @@ pub struct Vault {
 // 1. `SAFE_MOCK_VAULT_PATH` env var
 // 2. DevConfig `mock_vault_path` option
 // 3. None
```
nbaksalyar: I think it should say "default temp dir" instead of `None` now

PR #607, opened 2017-10-31
[master] feat/app: add crypto API for signing/verifying

m-cat:


safe_app/src/ffi/crypto.rs
```diff
@@ -267,6 +357,105 @@ pub unsafe extern "C" fn enc_secret_key_free(
...
+                    context.object_cache().get_sec_sign_key(sign_sk_h),
+                    user_data,
+                    o_cb
+                );
+                sign_sk.clone()
+            };
+
+            let signed_text = sign::sign(&plaintext, &sign_sk);
+
+            match serialise(&signed_text) {
```
nbaksalyar: How do you think, maybe we can return the result in a raw form without serialising it? From what I understand, `sign::sign` should return a `Vec<u8>`, which we can pass as the result as it is, just by using `signed_text.as_safe_ptr()`.

safe_app/src/ffi/crypto.rs
```diff
@@ -267,6 +357,105 @@ pub unsafe extern "C" fn enc_secret_key_free(
...
+                Ok(result) => o_cb(user_data.0, FFI_RESULT_OK, result.as_ptr(), result.len()),
```
nbaksalyar: `FFI_RESULT_OK` should be passed by reference now, please update it to be compatible with latest `master` changes.

Also, it's best to return pointers to vectors with `.as_safe_ptr()` (`.as_ptr()` in Rust returns `0x01` if the vec is empty, which doesn't work quite right with node-ffi)

safe_app/src/ffi/crypto.rs
```diff
@@ -267,6 +357,105 @@ pub unsafe extern "C" fn enc_secret_key_free(
...
+        let signed = vec_clone_from_raw_parts(signed_data, len);
+
+        (*app).send(move |_, context| {
+            let sign_pk = try_cb!(
+                context.object_cache().get_pub_sign_key(sign_pk_h),
+                user_data,
+                o_cb
+            );
+
+            match deserialise::<Vec<u8>>(&signed) {
```
nbaksalyar: Same thing here, I'd suggest trying to rewrite it without deserialisation and see if it works

safe_app/src/ffi/crypto.rs
```diff
@@ -267,6 +357,105 @@ pub unsafe extern "C" fn enc_secret_key_free(
...
+                Ok(signed_data) => {
+                    let verified =
+                        try_cb!(sign::verify(&signed_data, &sign_pk)
+                                .map_err(|()| AppError::EncodeDecodeError), user_data, o_cb);
+                    o_cb(
+                        user_data.0,
+                        FFI_RESULT_OK,
```
nbaksalyar: Same thing about `&FFI_RESULT_OK` (btw, maybe it'd make sense to change it into `pub static FFI_RESULT_OK: &'static FfiResult = struct { ... }`?)

safe_app/src/ffi/crypto.rs
```diff
@@ -493,9 +682,66 @@ mod tests {
     use super::*;
     use ffi_utils::test_utils::{call_0, call_1, call_2, call_vec_u8};
     use rust_sodium::crypto::box_;
-    use safe_core::arrays::{AsymNonce, AsymPublicKey, SignPublicKey};
+    use safe_core::arrays::{AsymNonce, AsymPublicKey, SignPublicKey, SignSecretKey};
     use test_utils::{create_app, run_now};
 
+    // Test signing and verifying messages between apps.
+    #[test]
+    fn sign_verify() {
```
nbaksalyar: Neat test, nothing to change here :)

safe_app/src/ffi/crypto.rs
m-cat: I agree with this, but wouldn't it have been better to do that in https://github.com/maidsafe/safe_client_libs/pull/594? It will just generate a lot of merge conflicts and unnecessary work if I do it here. I can also do it in my next PR.

nbaksalyar: @m-cat, sure, feel free to ignore it for now, that's just an idea/suggestion for future consideration.

PR #613, opened 2017-11-20
feat/app: add FFI function to safe_app to get app's container name

m-cat:


safe_app/src/ffi/mod.rs
```diff
@@ -48,6 +48,7 @@ use config_file_handler;
 use ffi_utils::{FFI_RESULT_OK, FfiResult, OpaqueCtx, ReprC, catch_unwind_cb, from_c_str};
 use futures::Future;
 use maidsafe_utilities::serialisation::deserialise;
+use safe_authenticator::app_container;
```
nbaksalyar: I think it'd better to move this module to `safe_core` or some other common place, as `safe_authenticator` is an [optional dependency](https://github.com/maidsafe/safe_client_libs/blob/d5d9253bdd4a166f59be978c86c3b2c4ee2571f5/safe_app/Cargo.toml#L24) that's used mostly in testing.

PR #616, opened 2017-11-27
test/integration: add integration test that connects to real network

m-cat:


safe_authenticator/src/ffi/ipc.rs
```diff
@@ -391,7 +391,7 @@ pub unsafe extern "C" fn encode_auth_resp(
     user_data: *mut c_void,
     o_cb: extern "C" fn(user_data: *mut c_void,
                         result: *const FfiResult,
-                        repsonse: *const c_char),
+                        response: *const c_char),
```
nbaksalyar: Nice catch :)

tests/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+
+#![cfg_attr(feature="cargo-clippy", deny(clippy, unicode_not_nfc, wrong_pub_self_convention,
+                                         option_unwrap_used))]
+#![cfg_attr(feature="cargo-clippy", allow(use_debug, too_many_arguments))]
+
+extern crate ffi_utils;
+extern crate safe_app;
+extern crate safe_authenticator;
+extern crate safe_core;
+// extern crate serde;
```
nbaksalyar: I'd suggest to remove dead code

tests/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+
+// Gets account credentials from the env vars "TEST_ACC_LOCATOR" and "TEST_ACC_PASSWORD".
+// If not found, reads the `tests.config` config file and returns it or panics if this fails.
+fn get_config() -> TestConfig {
+    match std::env::var("TEST_ACC_LOCATOR") {
+        Ok(acc_locator) => {
+            TestConfig {
+                test_account: AccountConfig {
+                    acc_locator,
+                    acc_password: unwrap!(std::env::var("TEST_ACC_PASSWORD")),
```
nbaksalyar: Minor thing: I think rewriting this as
```
let env = env::var("TEST_ACC_LOCATOR")
  .and_then(|acc_locator| Ok((acc_locator, env::var("TEST_ACC_PASSWORD"))));

match env {
  Ok((acc_locator, acc_password)) => { ... },
  Err(_) => { ... },
}
```
will make the code easier to reason about and there'll be less unwraps.
Feel free to ignore this comment though.

Also, it'd have been nice to put the env var names as the module-level static vars.

tests/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+        }))
+    };
+
+    // Get list of revoked apps.
+    let revoked_apps: Vec<AppExchangeInfo> =
+        unsafe { unwrap!(call_vec(|ud, cb| auth_revoked_apps(auth_h, ud, cb))) };
+    assert!(revoked_apps.iter().any(
+        |revoked_app| revoked_app.id == app_id,
+    ));
+
```
nbaksalyar: Would be nice to have the app re-authorised again here.
That way, we can also check that any data put to containers that an app had access to can still be read after re-encryption. It'd make sense to check the opposite too, actually: after the app is revoked it should not be able to read containers it had access to.

PR #665, opened 2018-08-22
[MAID-2990] refactor/ffi_utils: remove ffi_utils from workspace

m-cat:
Despite being a standalone crate, ffi_utils currently shares a repository and workspace with Safe Client Libs. As system_uri and parsec have it as a dependency, we are decoupling ffi_utils from Safe Client Libs. Benefits of doing this include making Client Libs smaller and more focused and being able to modify ffi_utils without building and testing all of Client Libs in CI.


nbaksalyar:
It's important to also fix the `bindgen` feature which depends on structs from `ffi_utils`.

PR #671, opened 2018-09-13
Use case insensitive URIs for IPC [MAID-3149]

nbaksalyar:
Using case sensitive base64 encoding caused some issues on different platforms. Base32 is case insensitive and more versatile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,18 @@
 vars:
   - &stable "1.28.0"
-  - &nightly "nightly-2018-07-07"
 env:
   global:
     - RUST_BACKTRACE=1
     - RUSTFLAGS="-C opt-level=2 -C codegen-units=8"
     - PATH=$PATH:$HOME/.cargo/bin
     - RUST_STABLE=1.28.0
-    - RUST_NIGHTLY=nightly-2018-07-07
-    - RUST_RUSTFMT=0.99.2
-    - RUST_CLIPPY=0.0.212
 language: rust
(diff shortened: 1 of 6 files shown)


bochaco:
Just for the record here, as discussed with @nbaksalyar we believe it will be good to support backward compatibility with base64 to prevent from having incompatible applications, e.g. a browser with latest safe_authenticator lib and an old web hosting manager app. He also suggested using something like the multiformats for the URI which I think could be good to allow backward compatible changes in the future as well.

bochaco:
One thing I noticed is that when an invalid URI is being decoded now the error message is different from `Serialisation error`, to 'IPC error: InvalidMsg', is this correct? if so, we would just need to update our tests.

I tried it with the safe browser and WHM apps, and when both apps use this branch of safe_app and safe_authenticator it seems to work all ok, but here is a problem if I try with previous version of safe_app (v0.8.0) on the WHM app, it obviously fails to decode the base32 auth uri.
So It seems that to correctly provide backward compatibility we should have a way of encoding a response with previous algo (base64 in this case) if the request received was encoded with previous algo. Is this something that has to be done in the Authenticator app within the browser or it can be done in the safe_authenticator lib? I'm starting to wonder if it worthwhile trying to support this now, or perhaps make sure we release new apps altogether with the new encoding, and hopefully if you are introducing multiformats then next time it will be easier to do?

nbaksalyar:
Superseded by #675

PR #672, opened 2018-09-20
Fixes for Java/JNI bugs [MAID-2659]

nbaksalyar:
By default Android JNI derives the class loader from the last Java class available in the call stack. This fails in case if we call from non-Java context (as we do when we call Java callbacks from the event loop thread
in SAFE Client Libs). This PR fixes this by caching the class loader during the library load and using it afterwards to look up Java classes.

Also, this PR uses fixed & updated `ffi_utils` and `safe_bindgen` to generate more robust JNI code and also use proper JNI threads detachment (using [`EnvGuard`](https://github.com/maidsafe/ffi_utils/blob/f7567e63f7258b18268b889af542a2d65274bb1c/src/java.rs#L21-L32) from `ffi_utils`).

### DIFF
--- a/safe_authenticator_jni/src/lib.rs
+++ b/safe_authenticator_jni/src/lib.rs
@@ -29,10 +29,10 @@ extern crate unwrap;
 use ffi_utils::java::{convert_cb_from_java, object_array_to_java, JniResult};
 use ffi_utils::*;
 use jni::errors::{Error as JniError, ErrorKind};
-use jni::objects::{GlobalRef, JClass, JObject, JString};
+use jni::objects::{AutoLocal, GlobalRef, JClass, JMethodID, JObject, JString, JValue};
 use jni::strings::JNIStr;
 use jni::sys::{jbyte, jbyteArray, jint, jlong, jobject, jsize};
-use jni::{JNIEnv, JavaVM};
+use jni::{signature::JavaType, JNIEnv, JavaVM};
 use safe_authenticator::*;
 use safe_core::arrays::*;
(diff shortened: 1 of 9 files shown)


safe_authenticator_jni/src/lib.rs
```diff
@@ -53,14 +53,55 @@ use std::{cmp, mem, slice};
...
+    FIND_CLASS_METHOD = Some(unwrap!(env.get_method_id(
+        "java/lang/ClassLoader",
+        "findClass",
+        "(Ljava/lang/String;)Ljava/lang/Class;",
+    )));
+
     jni::sys::JNI_VERSION_1_4
 }
 
+pub(crate) fn find_class<'a>(env: &'a JNIEnv, class_name: &str) -> AutoLocal<'a> {
```
m-cat: Looks like safe_app_jni has a similar (identical?) function to this, and other functionality. Is there a way we could avoid replicating code between the two crates?

lionel1704:
@m-cat I have built the libs and tried them out. They work as expected on mobile, but, however there seems to be something wrong on desktop. I've raised this with @nbaksalyar already. Just putting this in here too:
On desktop while calling the APIs this error is thrown:
```
java.lang.NoClassDefFoundError: IllegalName: net/maidsafe/safe_authenticator/FfiResult
```
The error occurs in multiple places whenever the callback(`FfiResult`) is handled

nbaksalyar:
> there seems to be something wrong on desktop.

This will be addressed in a separate PR, along with other bug fixes.
Thanks for testing Lionel!

PR #676, opened 2018-10-15
[MAID-3220] feat/app: add `GET_NEXT_VERSION` constant 

m-cat:


hunterlester:
@m-cat A couple of things we noticed:
- We need a way to get the latest file version when `GET_NEXT_VERSION` is passed to `dir_update_file`. Can [dir_update_file](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/nfs.rs#L101) return a file entry version?
- Should [dir_delete_file](https://github.com/maidsafe/safe_client_libs/blob/master/safe_app/src/ffi/nfs.rs#L123) also support `GET_NEXT_VERSION`?

m-cat:
@hunterlester 
* We can make `dir_update_file` return the latest version, good idea!
* We can surely make `dir_delete_file` support `GET_NEXT_VERSION` but then the name of the constant doesn't make sense anymore. I'm not sure whether it's better to rename the constant to `GET_VERSION` or use a separate one for `dir_delete_file`, like `GET_CURRENT_VERSION`. I'm leaning towards the first option since it would get confusing which constants can be used where.

Also, if we reuse the same constant, its value should be changed from `0` to the maximum possible value, as we may want to delete a file whose version is `0` and avoid the unnecessary network GET.

nbaksalyar:
@m-cat for the Rust API I'd probably even use an enum like `pub enum Version { Get, Custom(u64) }` (with better names of course :)). This will be a breaking change, but if we document it in the changelog it can go with the 0.9.0 release. What do you think?

For the FFI API I agree it's better to use two separate constants.

bochaco:
@m-cat , I'm confused by this:
>  I'm thinking we use two separate constants for clarity. It should be obvious to the user whether the next version will be obtained (for update) or the current one (for delete)
I think for both cases you need the next version, and in both cases it would make sense to have the function to automatically realise the next version and return it back, am I right?

m-cat:
I see, you're right @bochaco. You do need to increment the version when deleting a file. In that case it makes perfect sense to reuse the same constant. 👍  Sorry for the confusion there.

m-cat:
@hunterlester @bochaco The latest commit adds `GET_NEXT_VERSION` support to `dir_delete_file`, and both `dir_delete_file` and `dir_update_file` now return the latest file entry version.

PR #679, opened 2018-11-01
 test/nfs: test writing to files in append mode with different sizes 

m-cat:
+ added a test for appending with both small and medium SE instances
+ some minor documentation fixes

### DIFF
--- a/safe_core/src/nfs/tests.rs
+++ b/safe_core/src/nfs/tests.rs
@@ -17,6 +17,7 @@ use nfs::writer::Writer;
 use nfs::{create_dir, file_helper, File, Mode, NfsError, NfsFuture};
 use rand::{self, Rng};
 use rust_sodium::crypto::secretbox;
+use self_encryption::MIN_CHUNK_SIZE;
 use std;
 use utils::test_utils::random_client;
 use utils::FutureExt;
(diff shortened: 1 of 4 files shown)


safe_core/src/nfs/tests.rs
```diff
@@ -404,37 +411,42 @@ fn file_update_overwrite() {
...
-                let size = reader.size();
-                println!("reading {} bytes", size);
-                reader.read(0, size)
-            }).map(move |data| {
-                assert_eq!(data.len(), ORIG_SIZE + APPEND_SIZE);
-                assert_eq!(data[0..ORIG_SIZE].to_owned(), vec![0u8; ORIG_SIZE]);
-                assert_eq!(&data[ORIG_SIZE..], [2u8; APPEND_SIZE]);
-            })
-    });
+    for i in 0..3 {
```
nbaksalyar: Probably better to move it inside `random_client`: this way we can run the test for different sizes in parallel instead of waiting for all of them sequentially.

So it'd be more like
```rust
random_client(move |client| {
  let futures = vec![];
  for i in 0..3 {
    futures.push(create_test_file_with_size(client, size).then(...));
  }
  future::join(futures)
});
```

PR #717, opened 2018-12-20
test/nfs: test opening files in both read and write modes

m-cat:


safe_app/src/ffi/tests/nfs.rs
```diff
@@ -757,6 +757,120 @@ fn open_close_file() {
...
+        unwrap!(call_vec_u8(|ud, cb| file_read(
+            &app,
+            read_h,
+            0,
+            FILE_READ_TO_END,
+            ud,
+            cb
+        )))
+    };
+    assert_eq!(retrieved_content, b"hello world");
```
nbaksalyar: Probably this should be `assert_eq!(retrieved_content, content)`?
For the case if you'd want to change or randomise the written content.

safe_app/src/ffi/tests/nfs.rs
```diff
@@ -757,6 +757,120 @@ fn open_close_file() {
...
+            &written_file.into_repr_c(),
+            OPEN_MODE_OVERWRITE | OPEN_MODE_APPEND | OPEN_MODE_READ,
+            ud,
+            cb,
+        )))
+    };
+
+    // Can query the size since the file is opened in read mode.
+    let size: u64 = unsafe { unwrap!(call_1(|ud, cb| file_size(&app, read_write_h, ud, cb))) };
+    assert_eq!(size, 0);
```
nbaksalyar: Maybe replace it `assert_eq!(size, initial_content.len())`? Should be easier to understand it this way.

safe_app/src/ffi/tests/nfs.rs
```diff
@@ -757,6 +757,120 @@ fn open_close_file() {
...
+    let (app, container_info) = setup();
+
+    // Create empty file.
+    let file = NativeFile::new(Vec::new());
+    let ffi_file = file.into_repr_c();
+
+    let initial_content = b"";
+    let content = b"hello world";
+
+    // Write to the file first or we can't open it for reading.
```
nbaksalyar: This bit looks somewhat vague to me. Maybe expand it a little bit, e.g. `We should write to the file first because we can't open a non-existent file for reading`?

PR #762, opened 2019-03-27
doc/readme: add About section

m-cat:


README.md
```diff
@@ -17,6 +17,12 @@ Crate graph ([explanation](https://github.com/maidsafe/safe_client_libs/wiki#cra
 
 ![safe_app dependencies](safe-client-libs.png)
 
+## About
```
nbaksalyar: The contents looks good :+1: 

I'd suggest putting this section at the top of the file, below the "[ MaidSafe website | ... ]" block, it would look even better there. What do you think?

PR #763, opened 2019-03-28
Allow client_stress_test to login using flags

iancoleman:
This enables repeat runs of the client_stress_test.

eg ./client_stress_test --locator abcd --password 0123
will run the client_stress_test using the supplied login credentials

This commit also renames secret_0 and secret_1 to their conventional names.

### DIFF
--- a/safe_app/examples/client_stress_test.rs
+++ b/safe_app/examples/client_stress_test.rs
@@ -136,6 +136,20 @@ fn main() {
                 .requires("seed")
                 .help("Only Get the data, don't Put it. Logs in to an existing account."),
         ).arg(
+            Arg::with_name("locator")
+                .short("l")
+                .long("locator")
+                .takes_value(true)
+                .requires("password")
+                .help("Use the given Locator for login."),
+        ).arg(
+            Arg::with_name("password")
+                .short("p")
+                .long("password")
+                .takes_value(true)
+                .requires("locator")
+                .help("Use the given Password for login."),
+        ).arg(
             Arg::with_name("invite")
                 .long("invite")
                 .takes_value(true)
@@ -161,31 +175,40 @@ fn main() {
     let get_only = matches.is_present("get-only");
 
     // Create account
-    let secret_0: String = rng.gen_ascii_chars().take(20).collect();
-    let secret_1: String = rng.gen_ascii_chars().take(20).collect();
+    let (locator, password) = if let (Some(locator), Some(password)) =
+        (matches.value_of("locator"), matches.value_of("password"))
+    {
+        (locator.to_string(), password.to_string())
+    } else {
+        let new_locator = rng.gen_ascii_chars().take(20).collect();
+        let new_password = rng.gen_ascii_chars().take(20).collect();
+        println!(
+            "A new account will be created.\nLocator: {}\nPassword: {}",
+            new_locator, new_password
+        );
+        (new_locator, new_password)
+    };
 
-    let auth = if get_only {
+    let try_login = matches.is_present("locator") && matches.is_present("password");
+    let auth = if get_only || try_login {
         println!("\n\tAccount Login");
         println!("\t================");
         println!("\nTrying to login to an account ...");
 
-        unwrap!(Authenticator::login(secret_0, secret_1, || ()))
+        unwrap!(Authenticator::login(locator, password, || ()))
     } else {
         println!("\n\tAccount Creation");
         println!("\t================");
         println!("\nTrying to create an account ...");
 
         let auth = unwrap!(Authenticator::create_acc(
-            secret_0.as_str(),
-            secret_1.as_str(),
+            locator.as_str(),
+            password.as_str(),
             invitation.as_str(),
             || ()
         ));
 
-        println!(
-            "Created an account!\nLocator: {}\nPassword: {}",
-            secret_0, secret_1
-        );
+        println!("Account created!");
 
         auth
     };


safe_app/examples/client_stress_test.rs
```diff
@@ -135,6 +135,16 @@ fn main() {
                 .long("get-only")
                 .requires("seed")
                 .help("Only Get the data, don't Put it. Logs in to an existing account."),
+        ).arg(
+            Arg::with_name("locator")
```
m-cat: This won't work as-is. It requires `takes_value(true)` to be set; same with `password`.

iancoleman:
Thanks. I've resolved the argument parsing.

Also it was trying to create an account rather than do a login, so if `locator` and `password` are both present it will login instead of try to create an account.

safe_app/examples/client_stress_test.rs
```diff
@@ -135,6 +135,18 @@ fn main() {
...
+                .short("l")
+                .long("locator")
+                .takes_value(true)
```
m-cat: It doesn't make sense to provide a locator without a password and vice versa. Can you add `.requires("password")` here (and also make password require location)?

safe_app/examples/client_stress_test.rs
```diff
@@ -161,30 +173,40 @@ fn main() {
...
+        i.to_string()
+    } else {
+        rng.gen_ascii_chars().take(20).collect()
+    };
+    let password: String = if let Some(i) = matches.value_of("password") {
+        i.to_string()
+    } else {
+        rng.gen_ascii_chars().take(20).collect()
+    };
+
```
m-cat: It would be nice to print the locator and password, I found it useful when testing the changes locally but an end user can also find it useful to see what locator and password were used.

safe_app/examples/client_stress_test.rs
```diff
@@ -161,30 +173,40 @@ fn main() {
     let get_only = matches.is_present("get-only");
 
     // Create account
-    let secret_0: String = rng.gen_ascii_chars().take(20).collect();
-    let secret_1: String = rng.gen_ascii_chars().take(20).collect();
+    let locator: String = if let Some(i) = matches.value_of("locator") {
```
m-cat: You can simplify this a bit with something like:

```rust
let (locator, password) = if let (Some(locator), Some(password) = (matches.value_of("locator"), matches.value_of("password")) { 
    (locator, password)
} else { ... }
```

This makes the code a bit cleaner and also makes it semantically more clear that *both* the locator and the password have to be set.

Also you can get `try_login` from this if-statement here instead of computing it later by checking `matches.is_present`. Again, nothing wrong with the code as-is, this would just make it cleaner.

m-cat:
> Also just curious for your views about force-push; I basically never do force-push but in this PR wanted to keep the commit history clean rather than two commits for one unit of work. What's your views on this?

We encourage amending the most recent commit if there are simple fixes to be made, and force-pushing the commit. GitHub now shows diffs for force-pushed commits, making them easy to review, and it keeps the final commit history clean.

Good question btw, will add a note about this to our "Guide to contribution" page.

m-cat:
Added a section about code review :) https://github.com/maidsafe/safe_client_libs/wiki/Guide-to-contributing#code-review

safe_app/examples/client_stress_test.rs
iancoleman: When the account is created it does already print the new credentials:

safe_app/examples/client_stress_test.rs
iancoleman: Thanks for pointing this out, I really appreciate guidance on code style in rust.

I couldn't get the values returned in the `else` clause to be of the right type, I would have thought this would work:

```
let (locator, password) = if let (Some(locator), Some(password)) = (matches.value_of("locator"), matches.value_of("password")) {
    (locator, password)
} else {
    (rng.gen_ascii_chars().take(20).collect(), rng.gen_ascii_chars().take(20).collect())
};
```

but it throws an error

```
a collection of type `&str` cannot be built from `std::iter::Iterator<Item=char>`
```

I tried turbofish with the collect method (and a bunch of other things) but couldn't get this to work.

From my existing/old code habits, keeping a similar pattern to the `invitation` seemed overall easier to read (same visual structure, avoids long line length, less syntactic sugar), but I can see how it communicates less about the relation between `locator` and `password` compared to the suggested change.

m-cat: Yeah, I didn't try the code and left something out. The return value should be:

```rust
(locator.to_string(), password.to_string())
```

The `collect()` calls will now be able to infer the type to collect into (`String`).

> From my existing/old code habits, keeping a similar pattern to the invitation seemed overall easier to read (same visual structure, avoids long line length, less syntactic sugar)

If you run rustfmt I imagine it would shorten these lines and make them more readable.

It still is the same structure as `invitation`, except that you're using pattern-matching on a tuple (quite common in Rust-land). Personally I find this way to be cleaner and easier to read, though I can see it being confusing for a non-Rust programmer. I'll leave this up to you.

safe_app/examples/client_stress_test.rs
m-cat: But it would be nice to see the locator/password again even if they were passed in (could help catch typos). Also, if account creation fails, the locator/password never get printed. Just justifying my opinion, the final call is yours.

iancoleman: Yes I agree, it's important to show the new credentials before the account is created in case it fails. I added this.

safe_app/examples/client_stress_test.rs
iancoleman: Thanks for the help. I see my problem was I was trying to get the inference to happen from the `else` condition but it needed to be in the initial condition.

safe_app/examples/client_stress_test.rs
m-cat: Won't the credentials be printed twice now if an account is created successfully? Maybe remove the credentials from the "Created an account!" message.

PR #787, opened 2019-04-05
refactor/safe_authenticator: provide Rust API for application management and revocation

### LINKED ISSUE #771: Provide Rust API for apps management in SAFE Authenticator
Currently, `safe_authenticator::ffi::apps` is the only way to access some of high-level functions of the Authenticator. We should move code from that module to Rust functions and change the FFI functions to be thin wrappers around native code.

The new module should provide the following functions:

* Remove a revoked app.

```rust
pub fn remove_revoked_app<S: Into<String>>(
  client: &impl Client,
  app_id: S
) -> impl Future<(), AuthError>
```

(move code from [`auth_rm_revoked_app`](https://github.com/maidsafe/safe_client_libs/blob/93154e0fa8bccad18fd6b26457cd80071092b386/safe_authenticator/src/ffi/apps.rs#L58-L60))

* List revoked apps

```rust
pub fn revoked_apps(
  client: &impl Client
) -> impl Future<Vec<AppExchangeInfo>, AuthError>;
```

(move code from [`auth_revoked_apps`](https://github.com/maidsafe/safe_client_libs/blob/93154e0fa8bccad18fd6b26457cd80071092b386/safe_authenticator/src/ffi/apps.rs#L98-L100))

* List registered apps

```rust
pub fn registered_apps(
  client: &impl Client
) -> impl Future<Vec<RegisteredApp>, AuthError>;
```

(move code from [`auth_registered_apps`](https://github.com/maidsafe/safe_client_libs/blob/93154e0fa8bccad18fd6b26457cd80071092b386/safe_authenticator/src/ffi/apps.rs#L157-L159))

The `RegisteredApp` struct is defined below.

* List apps that have access to a certain Mutable Data object on the network

```
use safe_core::ipc::response::AppAccess;

pub fn apps_accessing_mutable_data(
  client: &impl Client,
  name: XorName,
  type_tag: u64,
) -> impl Future<Vec<AppAccess>, AuthError>
```

(move code from [`auth_apps_accessing_mutable_data`](https://github.com/maidsafe/safe_client_libs/blob/93154e0fa8bccad18fd6b26457cd80071092b386/safe_authenticator/src/ffi/apps.rs#L231-L234)

---

The module should also provide a native struct `RegisteredApp`:

```rust
use safe_core::ipc::req::{AppExchangeInfo, Permission};

struct RegisteredApp {
    /// Unique application identifier
    pub app_info: AppExchangeInfo,
    /// List of containers that this application has access to
    /// Maps from the container name to the set of permissions
    pub containers: BTreeMap<String, BTreeSet<Permission>>,
}
```

the [FFI struct](https://github.com/maidsafe/safe_client_libs/blob/93154e0fa8bccad18fd6b26457cd80071092b386/safe_authenticator/src/ffi/apps.rs#L32-L44) should be a wrapper around it.

### LINKED ISSUE #786: Make revocation module in SAFE Authenticator public
Public functions in the revocation module, `revoke_app` and `flush_app_revocation_queue`, should be available as part of the authenticator API. Make [`mod revocation` in lib.rs](https://github.com/maidsafe/safe_client_libs/blob/0190e76a40c437b7caf9072fbe39b5b233e273fa/safe_authenticator/src/lib.rs#L109) public.

lionel1704:
- remove logic for application management from the FFI layer and expose them as native Rust API
- refactor their respective FFI functions as wrappers to the native API
- add the `RegisteredApp` native RUST structure and implement a `into_repr_c()` function to convert it to the FFI struct representation
- make revocation functions publicly accessible in the API

closes #771, closes #786

### DIFF
--- a/safe_authenticator/src/apps.rs
+++ b/safe_authenticator/src/apps.rs
@@ -1,0 +1,237 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+//! App management functions
+
+use super::{config, AuthFuture};
+use app_auth::{app_state, AppState};
+use app_container;
+use client::AuthClient;
+use ffi::apps as ffi;
+use ffi::apps::RegisteredApp as FfiRegisteredApp;
+use ffi_utils::{vec_into_raw_parts, ReprC};
+use futures::future::Future;
+use maidsafe_utilities::serialisation::deserialise;
+use routing::User::Key;
+use routing::XorName;
+use safe_core::client::Client;
+use safe_core::ipc::req::{containers_from_repr_c, containers_into_vec, ContainerPermissions};
+use safe_core::ipc::resp::{AccessContainerEntry, AppAccess};
+use safe_core::ipc::{access_container_enc_key, AppExchangeInfo, IpcError};
+use safe_core::utils::symmetric_decrypt;
+use safe_core::FutureExt;
+use std::collections::HashMap;
+use AuthError;
+
+/// Represents an application that is registered with the Authenticator.
+#[derive(Debug)]
+pub struct RegisteredApp {
+    /// Unique application identifier.
+    pub app_info: AppExchangeInfo,
+    /// List of containers that this application has access to.
+    /// Maps from the container name to the set of permissions.
+    pub containers: HashMap<String, ContainerPermissions>,
+}
+
+impl RegisteredApp {
+    /// Construct FFI wrapper for the native Rust object, consuming self.
+    pub fn into_repr_c(self) -> Result<FfiRegisteredApp, IpcError> {
+        let RegisteredApp {
+            app_info,
+            containers,
+        } = self;
+
+        let container_permissions_vec = containers_into_vec(containers.into_iter())?;
+
+        let (containers_ptr, containers_len, containers_cap) =
+            vec_into_raw_parts(container_permissions_vec);
+
+        Ok(FfiRegisteredApp {
+            app_info: app_info.into_repr_c()?,
+            containers: containers_ptr,
+            containers_len,
+            containers_cap,
+        })
+    }
+}
+
+impl ReprC for RegisteredApp {
+    type C = *const ffi::RegisteredApp;
+    type Error = IpcError;
+
+    unsafe fn clone_from_repr_c(repr_c: Self::C) -> Result<Self, Self::Error> {
+        Ok(Self {
+            app_info: AppExchangeInfo::clone_from_repr_c(&(*repr_c).app_info)?,
+            containers: containers_from_repr_c((*repr_c).containers, (*repr_c).containers_len)?,
+        })
+    }
+}
+
+/// Removes an application from the list of revoked apps.
+pub fn remove_revoked_app(client: &AuthClient, app_id: String) -> Box<AuthFuture<()>> {
+    let client = client.clone();
+    let c2 = client.clone();
+    let c3 = client.clone();
+    let c4 = client.clone();
+
+    let app_id = app_id.clone();
+    let app_id2 = app_id.clone();
+    let app_id3 = app_id.clone();
+
+    config::list_apps(&client)
+        .and_then(move |(apps_version, apps)| {
+            app_state(&c2, &apps, &app_id).map(move |app_state| (app_state, apps, apps_version))
+        })
+        .and_then(move |(app_state, apps, apps_version)| match app_state {
+            AppState::Revoked => Ok((apps, apps_version)),
+            AppState::Authenticated => Err(AuthError::from("App is not revoked")),
+            AppState::NotAuthenticated => Err(AuthError::IpcError(IpcError::UnknownApp)),
+        })
+        .and_then(move |(apps, apps_version)| {
+            config::remove_app(&c3, apps, config::next_version(apps_version), &app_id2)
+        })
+        .and_then(move |_| app_container::remove(c4, &app_id3).map(move |_res| ()))
+        .into_box()
+}
+
+/// Returns a list of applications that have been revoked.
+pub fn list_revoked(client: &AuthClient) -> Box<AuthFuture<Vec<AppExchangeInfo>>> {
+    let c2 = client.clone();
+    let c3 = client.clone();
+
+    config::list_apps(client)
+        .map(move |(_, auth_cfg)| (c2.access_container(), auth_cfg))
+        .and_then(move |(access_container, auth_cfg)| {
+            c3.list_mdata_entries(access_container.name, access_container.type_tag)
+                .map_err(From::from)
+                .map(move |entries| (access_container, entries, auth_cfg))
+        })
+        .and_then(move |(access_container, entries, auth_cfg)| {
+            let mut apps = Vec::new();
+            let nonce = access_container
+                .nonce()
+                .ok_or_else(|| AuthError::from("No nonce on access container's MDataInfo"))?;
+
+            for app in auth_cfg.values() {
+                let key = access_container_enc_key(&app.info.id, &app.keys.enc_key, nonce)?;
+
+                // If the app is not in the access container, or if the app entry has
+                // been deleted (is empty), then it's revoked.
+                let revoked = entries
+                    .get(&key)
+                    .map(|entry| entry.content.is_empty())
+                    .unwrap_or(true);
+
+                if revoked {
+                    apps.push(app.info.clone());
+                }
+            }
+            Ok(apps)
+        })
+        .into_box()
+}
+
+/// Return the list of applications that are registered with the Authenticator.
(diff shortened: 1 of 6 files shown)


safe_authenticator/src/lib.rs
```diff
@@ -108,6 +108,7 @@ mod errors;
 mod ipc;
 mod revocation;
 mod std_dirs;
+mod apps;
```
nbaksalyar: Should be either `pub mod apps`, or alternatively we can have `pub use apps::{registered_apps, ...};` below.
Also, what do you think if we rename functions to be more succinct?
I.e. compare:
```
use safe_authenticator::apps::{registered_apps, revoked_apps};
```
and
```
use safe_authenticator::apps::{registered, revoked};
```

m-cat: Suggestion: I think `list_registered` and `list_revoked` are more descriptive.

lionel1704: Sounds like a good idea. Since they are already in the `apps` module, we need not mentioned it again.

nbaksalyar: >`list_registered` and `list_revoked` are more descriptive
+1 on that :+1:

safe_authenticator/src/apps.rs
```diff
@@ -0,0 +1,236 @@
+// Copyright 2018 MaidSafe.net limited.
```
nbaksalyar: For new files a year should be updated I believe:
```suggestion
// Copyright 2019 MaidSafe.net limited.
```

safe_authenticator/src/apps.rs
```diff
@@ -0,0 +1,236 @@
...
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+//! SAFE Authenticator
```
nbaksalyar: This should contain module-level documentation. E.g. something like:
```suggestion
//! App management functions.
```

safe_authenticator/src/apps.rs
```diff
@@ -0,0 +1,236 @@
...
+            AppState::NotAuthenticated => Err(AuthError::IpcError(IpcError::UnknownApp)),
+        })
+        .and_then(move |(apps, apps_version)| {
+            config::remove_app(&c3, apps, config::next_version(apps_version), &app_id2)
+        })
+        .and_then(move |_| app_container::remove(c4, &app_id3).map(move |_res| ()))
+        .into_box()
+}
+
+/// Returns a list of applicatons that have been revoked.
```
m-cat: Typo!! "applicatons"

safe_authenticator/src/lib.rs
```diff
@@ -108,6 +108,7 @@ mod errors;
 mod ipc;
 mod revocation;
 mod std_dirs;
+pub mod apps;
```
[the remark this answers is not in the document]
lionel1704: @m-cat Should I add this right below `pub mod ffi;` or to a separate group? I feel that the former may imply that the comment applies to the entire group.

safe_authenticator/src/apps.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// You're now responsible for freeing this memory once you're done.
+    pub fn into_repr_c(self) -> Result<FfiRegisteredApp, IpcError> {
+        let RegisteredApp {
+            app_info,
+            containers,
+        } = self;
+
+        let container_permissions_vec = containers
+            .into_iter()
+            .map(|(container_name, permission_set)| FfiContainerPermissions {
```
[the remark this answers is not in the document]
lionel1704: Isn't `permission_set` already a `ContainerPermissions` object here? Are you suggesting that we have an alias for `BTreeMap<String, ContainerPermissions>` which has an implementation with `into_repr_c` ?

safe_authenticator/src/lib.rs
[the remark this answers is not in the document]
m-cat: I would put it before `pub mod ffi` to keep things alphabetical. The `FFI routines` doc string only applies to the `ffi` module, it's the same as putting `//! FFI routines` inside `ffi/mod.rs`.

Also, can you bring up `pub mod test_utils` to that group while you're at it?

safe_authenticator/src/apps.rs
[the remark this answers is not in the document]
m-cat: Exactly, and maybe `FfiContainerPermissions` should be renamed as it doesn't correspond exactly to the native `ContainerPermissions`. And yeah, I was thinking we would move this conversion logic over to a new `into_repr_c` function. I'm not entirely sure if this is worth doing, it would make things a bit easier to think about but I'll leave it for you to decide.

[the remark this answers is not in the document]
lionel1704: @m-cat it makes sense to do this, but, I think that it's better to create a separate issue for this and change _all_ the occurrences. What do you think?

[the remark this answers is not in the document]
m-cat: Sure, you can leave this code as-is for now.

safe_authenticator/src/ffi/apps.rs
```diff
@@ -111,42 +87,18 @@ pub unsafe extern "C" fn auth_revoked_apps(
...
-
-                        if revoked {
-                            apps.push(app.info.clone().into_repr_c()?);
-                        }
-                    }
-
-                    o_cb(user_data.0, FFI_RESULT_OK, apps.as_safe_ptr(), apps.len());
+            list_revoked(client)
+                .and_then(move |apps| {
+                    let app_list :Vec<_> = apps.iter()
```
m-cat: This syntax (with the colon right before the type) looks really strange to me, can you change it to `let app_list: Vec<_> ...`? There are a couple other instances of this that I saw :)

safe_authenticator/src/lib.rs
```diff
@@ -91,14 +91,20 @@ extern crate unwrap;
...
+#[cfg(any(test, feature = "testing"))]
+#[macro_use]
+pub mod test_utils;
 
-pub use ffi::apps::*;
 pub use ffi::ipc::*;
 pub use ffi::logging::*;
 pub use ffi::*;
 
+pub use ffi::apps::*;
```
m-cat: Did you mean to move this into its own group?

safe_authenticator/src/apps.rs
```diff
@@ -0,0 +1,237 @@
...
+                        Some(app_info) => AppAccess {
+                            sign_key: public_key,
+                            permissions: perm_set,
+                            name: Some(app_info.name.clone()),
+                            app_id: Some(app_info.id.clone()),
+                        },
+                        None => {
+                            // If an app is listed in the MD permissions list, but is not
+                            // listed in the registered apps list in Authenticator, then set
+                            // the app_id and app_name fields to ptr::null(), but provide
```
nbaksalyar: `ptr::null()` should be `None` here I suppose:
```suggestion
                            // the app_id and app_name fields to None, but provide
```

safe_authenticator/src/ffi/apps.rs
```diff
@@ -111,42 +87,18 @@ pub unsafe extern "C" fn auth_revoked_apps(
...
-                        if revoked {
-                            apps.push(app.info.clone().into_repr_c()?);
-                        }
-                    }
-
-                    o_cb(user_data.0, FFI_RESULT_OK, apps.as_safe_ptr(), apps.len());
+            list_revoked(client)
+                .and_then(move |apps| {
+                    let app_list: Vec<_> = apps.iter()
+                        .map(|app| app.clone().into_repr_c())
```
nbaksalyar: I think `app.clone()` is redundant here. Could we maybe try it like this?
```rust
let app_list: Vec<_> = apps.into_iter()
    .map(ReprC::into_repr_c)
    .collect::<Result<_, _>>()?;
```

safe_authenticator/src/ffi/apps.rs
```diff
@@ -170,51 +122,11 @@ pub unsafe extern "C" fn auth_registered_apps(
...
-                                containers_cap: cap,
-                            };
-
-                            apps.push(reg_app);
-                        }
-                    }
-
+            list_registered(client)
+                .and_then(move |registered_apps| {
+                    let apps: Vec<_> = registered_apps.iter()
```
nbaksalyar: Same thing about `.into_iter()`

lionel1704:
@nbaksalyar I'm not sure what you meany by `ReprC::into_repr_c`.
`into_repr_c` is not defined in the `ReprC` trait. Perhaps it should be? That way any struct that implements it would have proper conversion functions between it's native and C representations. If you agree, I can raise an issue for this.

I've updated the PR using `into_iter()` to get mutable references of the objects.

m-cat:
> @nbaksalyar I'm not sure what you meany by `ReprC::into_repr_c`.
> `into_repr_c` is not defined in the `ReprC` trait. Perhaps it should be? That way any struct that implements it would have proper conversion functions between it's native and C representations. If you agree, I can raise an issue for this.

I agree, and I've already written an issue about this:

https://github.com/maidsafe/ffi_utils/issues/21

It would be great if you could chime in with your thoughts!

nbaksalyar:
>`into_repr_c` is not defined in the ReprC trait

Ah you're right @lionel1704 👍 This can be replaced with `AppExchangeInfo::into_repr_c` for now (i.e. we still don't need that `clone` I think).

PR #799, opened 2019-04-22
refactor/feature: rename the `use-mock-routing` feature to `mock-network`

### LINKED ISSUE #793: Rename the `use-mock-routing` feature to `mock-network`
The `use-` prefix is redundant and `mock-routing` does not convey too much meaning to newcomers to the library.

We also should conform to standards used in our other projects: e.g. routing uses `mock-crust` and `mock-parsec` instead of `use-mock-...` too.

lionel1704:
closes #793

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,9 +65,9 @@ after_build:
 
 test_script:
   - |-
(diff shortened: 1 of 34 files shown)


nbaksalyar:
Great stuff, thanks :slightly_smiling_face:

PR #822, opened 2019-05-29
Implement new flow for mock routing for new data types

lionel1704:
- Add safe-nd depdendency
- Implement send() in mock routing for all client-vault interaction
- Add functions to the mock vault to handle a new data type
- Add put and get functions for the new type to the Client trait with tests
- Introduce BLS key-pair for clients

### DIFF
--- a/safe_core/src/client/account.rs
+++ b/safe_core/src/client/account.rs
@@ -131,6 +131,8 @@ pub struct ClientKeys {
     pub bls_sk: threshold_crypto::SecretKey,
 }
 
+// threshold_crypto::SecretKey cannot be serialised directly,
+// hence this trait is implemented
 impl Serialize for ClientKeys {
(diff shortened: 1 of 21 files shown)


safe_core/src/client/account.rs
```diff
@@ -123,6 +125,27 @@ pub struct ClientKeys {
     pub enc_sk: shared_box::SecretKey,
     /// Symmetric encryption key
     pub enc_key: shared_secretbox::Key,
+    /// BLS public key
+    pub bls_pk: threshold_crypto::PublicKey,
+    /// BLS private key
+    pub bls_sk: threshold_crypto::SecretKey,
+}
+
+impl Serialize for ClientKeys {
```
m-cat: Why implement Serialize here instead of deriving it? Maybe add a comment why this is necessary.

PR #833, opened 2019-06-07
feat/mdata: impl Mutable data DELETE RPC function and test

Yoga07:
feat/mdata: impl Mutable data DELETE RPC function and test

- Fixes breakage of experimental branch by altering function parameters as needed and fixing dependencies.
- Implements converter trait for XorName to facilitate conversion from `routing::XorName` to `safe_nd::XorName`
- Implements required functions for `DELETE` RPC in Mutable Data

### DIFF
--- a/safe_authenticator/src/errors.rs
+++ b/safe_authenticator/src/errors.rs
@@ -17,6 +17,7 @@ use routing::ClientError;
 use safe_core::ipc::IpcError;
 use safe_core::nfs::NfsError;
 use safe_core::CoreError;
+use safe_nd::Error as SndError;
 use std::error::Error;
 use std::ffi::NulError;
 use std::fmt::{self, Display, Formatter};
@@ -88,6 +89,12 @@ mod codes {
     pub const ERR_ACCOUNT_CONTAINERS_CREATION: i32 = -1014;
     pub const ERR_NO_SUCH_CONTAINER: i32 = -1015;
     pub const ERR_UNEXPECTED: i32 = -2000;
+
+    // BLS errors.
+    pub const ERR_INVALID_OWNERS_SUCCESSOR: i32 = -3001;
+    pub const ERR_INVALID_PERMISSIONS_SUCCESSOR: i32 = -3002;
+    pub const ERR_SIGN_KEYTYPE_MISMATCH: i32 = -3003;
+    pub const ERR_INVALID_SIGNATURE: i32 = -3004;
 }
 
 /// Authenticator errors.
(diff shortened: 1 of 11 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -168,6 +169,55 @@ impl Vault {
...
+        match *dst {
+            Authority::NaeManager(name) if name == *data_name => Ok(()),
+            x => {
+                debug!("Unexpected authority for delete: {:?}", x);
+                Err(ClientError::InvalidOperation)
+            }
+        }
+    }
+
+    pub fn authorised_delete(
```
lionel1704: The delete operation can be done only by _owners_.
The `Action::Delete` permission applies only to delete operations for entries in mutable data.

safe_authenticator/src/errors.rs
```diff
@@ -17,6 +17,7 @@ use routing::ClientError;
 use safe_core::ipc::IpcError;
 use safe_core::nfs::NfsError;
 use safe_core::CoreError;
+use safe_nd::Error as SndError;
```
m-cat: Calling this `SndError` can be a bit confusing. There's many error types in this file, so maybe call this something more specific than "second error"?

Yoga07: Oh damn! I meant that to be "Safe-nd Errors" :smile:

m-cat: Haha! 😄

safe_core/src/client/mock/vault.rs
```diff
@@ -150,21 +163,70 @@ impl Vault {
...
+        let req = if let Requester::Key(req) = requester {
+            req
+        } else {
+            return Err(Error::AccessDenied);
+        };
+        match self.get_data(&data_name) {
+            Some(data_type) => match data_type {
+                Data::NewMutable(kind) => match kind {
+                    MutableDataKind::Unsequenced(data) => {
+                        if data.owners() == &req {
```
nbaksalyar: This would not be correct.
`Requester::Key` is reserved for apps and other users, so the check should go the other way around - i.e.
```rust
let signature = if let Requester::Owner(signature) = requester {
  signature
} else {
  return Err(Error::AccessDenied);
};
```
Then, checking the ownership as this:
```rust
if data.owners() == &req {
```
wouldn't really work either because here we're just comparing the key, which wouldn't work that simply for BLS.
Instead, we need to check the signature:
```rust
if safe_nd::verify_signature(signature, data, &original_request) {
  ...
}
```
So we also need to have this `verify_owner` function in `safe_nd::mutable_data`.

PR #835, opened 2019-06-12
feat/idata: implement functions for unpublished idata

### LINKED ISSUE #810: Implement mock unpublished Immutable Data
* Define new methods in `trait Client`:
```rust
pub fn put_unpub_idata(&self, data: UnpubImmutableData) -> Box<CoreFuture<()>>;
pub fn get_unpub_idata(&self, name: XorName) -> Box<CoreFuture<UnpubImmutableData>>;
pub fn delete_unpub_idata(&self, name: XorName) -> Box<CoreFuture<()>>;
```
* Implement mock routing & mock vault handling of unpublished Immutable Data RPCs.
* Make sure only the owner(s) can get and delete unpublished Immutable Data objects. Verify the requester's public key is the same as the owner key stored in the `UnpubImmutableData` object itself. `DataManagers` must reject these requests from anyone but `MaidManagers`.
* It should be impossible to create name conflicts between Unpublished and Published Immutable Data.
* Implement tests:
    * Put `UnpubImmutableData`. The request should succeed.
    * Knowing the `XorName` of unpublished Immutable Data, try to fetch it from the network as an owner. The request should succeed.
    * Knowing the `XorName` of unpublished Immutable Data, try to fetch it from a `DataManagers` group directly. The request should fail regardless of whether an owner sent this request or not.
    * Knowing the `XorName` of unpublished Immutable Data, try to fetch it from the network while not being an owner (e.g. as an unregistered client). The request should be denied.
    * Knowing the `XorName` of unpublished Immutable Data, try to delete it from the network while not being an owner. The request should be denied.

m-cat:
Closes #810

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -20,7 +20,7 @@ use safe_nd::mutable_data::{
 };
 use safe_nd::request::{Request, Requester};
 use safe_nd::response::Response;
-use safe_nd::Error;
+use safe_nd::{Error, UnpubImmutableData, XorName as NewXorName};
 use std::collections::HashMap;
 use std::env;
 use std::fs::{File, OpenOptions};
@@ -156,6 +156,7 @@ impl Vault {
             Requester::Key(key) => Ok(key),
         };
         let public_key = unwrap!(public_key_res);
+
         match self.get_data(&data_name) {
             Some(data_type) => match data_type {
                 Data::NewMutable(data) => match data.clone() {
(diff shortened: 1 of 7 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -174,6 +175,20 @@ impl Vault {
                         }
                     }
                 },
+                Data::Immutable(data) => match data {
+                    ImmutableDataKind::Unpublished(idata) => {
+                        // Only the owners can read unpub idata.
+                        if public_key == *idata.owners() {
```
nbaksalyar: That will need to be changed, but we can leave it at that for now as there'll be a separate commit to handle the permissions properly.

safe_core/src/client/mod.rs
```diff
@@ -225,6 +225,80 @@ pub trait Client: Clone + 'static {
...
+        };
+        send(self, move |routing, msg_id| {
+            let request = Request::GetUnpubIData {
+                address: name,
+                requester: Requester::Key(requester),
+                message_id: msg_id.to_new(),
+            };
+            routing.send(
+                client,
+                Authority::NaeManager(XorName::from_new(name)),
```
nbaksalyar: This would be incorrect: requests for ImmutableData should always go through `ClientManager`, but that's OK for now as the client was slightly changed anyway in the `safecoin` branch, so it will be addressed within another PR.

lionel1704: Does this mean all _owned_ GETs go via the client manager?
I assumed that it was the role of the data managers to verify the requester against the permisson set / owner field?

PR #838, opened 2019-06-13
Pub idata

m-cat:

### DIFF
--- a/safe_core/src/client/mock/tests.rs
+++ b/safe_core/src/client/mock/tests.rs
@@ -22,10 +22,10 @@ use routing::{
     MutableData, PermissionSet, Request, Response, User, Value, TYPE_TAG_SESSION_PACKET,
 };
 use safe_nd::{
-    AppFullId, ClientFullId, Error, ImmutableData, MDataAction as NewAction, MDataAddress,
-    MDataPermissionSet as NewPermissionSet, Message, MessageId, MutableData as NewMutableData,
-    PublicKey, Request as RpcRequest, Response as RpcResponse, UnpubImmutableData,
-    UnseqMutableData, XorName,
+    AppFullId, ClientFullId, Error, IDataAddress, IDataKind, ImmutableData,
+    MDataAction as NewAction, MDataAddress, MDataPermissionSet as NewPermissionSet, Message,
+    MessageId, MutableData as NewMutableData, PublicKey, Request as RpcRequest,
+    Response as RpcResponse, UnpubImmutableData, UnseqMutableData, XorName,
 };
 use std::collections::BTreeMap;
 use std::sync::mpsc::{self, Receiver};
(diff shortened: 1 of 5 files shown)


safe_core/src/client/mod.rs
```diff
@@ -1362,9 +1388,43 @@ mod tests {
...
     use std::str::FromStr;
 
+    // Test putting and getting pub idata.
+    #[test]
+    fn pub_idata_test() {
+        random_client(move |client| {
+            let client2 = client.clone();
+            let client3 = client.clone();
+
+            let value: Vec<u8> = b"pub idata value".to_vec();
```
nbaksalyar: I think it's better to have a `rand()` content here because otherwise these tests would fail on a second run with `Pub idata should not exist` yet. I think we already have a utility function for generating a random vector of a specified size.

safe_core/src/client/mod.rs
```diff
@@ -1374,7 +1434,7 @@ mod tests {
             let client7 = client.clone();
             let client8 = client.clone();
 
-            let value: Vec<u8> = Default::default();
+            let value: Vec<u8> = b"unpub idata value".to_vec();
```
nbaksalyar: Same thing about the random content here.

safe_core/src/client/mock/tests.rs
```diff
@@ -1079,33 +1079,79 @@ fn mutable_data_ownership() {
...
+    let (mut routing, routing_rx, full_id) = setup();
+    let owner_key = PublicKey::from(*full_id.public_id().bls_public_key());
+    let _client_mgr = create_account(&mut routing, &routing_rx, owner_key);
+
+    let data = ImmutableData::new(Default::default());
+    let name = data.name().clone();
+
+    // Put pub idata. Should succeed.
+    {
+        let _ = routing.req_as_owner(&routing_rx, RpcRequest::PutIData(data.into()));
```
lionel1704: The RPC response is not validated here. Even though the next step would fail if this doesn't work as expected, maybe it's also worth validating this response as well.

safe_core/src/client/mod.rs
```diff
@@ -1357,14 +1391,49 @@ type TimeoutFuture = Either<
...
+        random_client(move |client| {
+            let client2 = client.clone();
+            let client3 = client.clone();
+
+            let value = unwrap!(generate_random_vector::<u8>(10));
+            let data = ImmutableData::new(value.clone());
+            let name = *data.name();
+
+            client
+                // Get inexistant idata
```
lionel1704: *inexistent :smile:

safe_core/src/client/mod.rs
```diff
@@ -1374,7 +1443,7 @@ mod tests {
             let client7 = client.clone();
             let client8 = client.clone();
 
-            let value: Vec<u8> = Default::default();
+            let value = unwrap!(generate_random_vector::<u8>(10));
             let data = UnpubImmutableData::new(value.clone(), unwrap!(client.public_bls_key()));
             let data2 = UnpubImmutableData::new(value.clone(), unwrap!(client.public_bls_key()));
```
lionel1704: Along with your changes, do you think we should also change this block to the following?
```
let data2 = data.clone();
let data3 = data.clone();
```

PR #839, opened 2019-06-13
WIP: Implement AppendOnly Data RPCs and functions in the mock vault

lionel1704:
Authored by @Yoga07

### DIFF
--- a/mock_routing/Cargo.toml
+++ b/mock_routing/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-safe-nd = { git = "https://github.com/maidsafe/safe-nd" }
+safe-nd = { git = "https://github.com/lionel1704/safe-nd/", branch = "pub-field" }
 maidsafe_utilities = "~0.18.0"
 rust_sodium = "~0.10.2"
 rand = "~0.3.18"
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -13,10 +13,15 @@ use crate::config_handler::{Config, DevConfig};
 use fs2::FileExt;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use routing::{Authority, ClientError, MutableData as OldMutableData};
+use safe_nd::appendable_data::{
+    Action as ADataAction, AppendOnlyData, AppendOnlyDataRef, AppendOnlyKind, Index, Indices,
+    PubPermissions, SeqAppendOnly, SeqAppendOnlyData, UnpubPermissions, UnseqAppendOnly,
+    UnseqAppendOnlyData,
+};
 use safe_nd::mutable_data::{
     MutableData as NewMutableData, MutableDataRef, SeqMutableData, UnseqMutableData,
 };
-use safe_nd::request::{Request, Requester};
+use safe_nd::request::{AppendOnlyData as AData, Request, Requester};
 use safe_nd::response::{Response, Transaction};
 use safe_nd::{
     Coins, Error, ImmutableData, Message, MessageId, PublicKey, UnpubImmutableData, XorName,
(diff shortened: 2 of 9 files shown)


safe_core/src/client/mod.rs
```diff
@@ -2028,4 +2379,604 @@ mod tests {
...
+                        AData::UnpubSeq(adata) => assert_eq!(adata.name(), name),
+                        _ => panic!("Unexpected data found"),
+                    })
+                })
+                .then(move |res| res)
+        });
+    }
+
+    #[test]
+    #[should_panic]
```
m-cat: Instead of panicking you could test that the proper error variant is returned, see e.g. `unpub_idata_test`.

safe_core/src/client/mod.rs
```diff
@@ -2028,4 +2379,604 @@ mod tests {
                 })
         });
     }
+
+    #[test]
+    pub fn put_adata_test() {
```
m-cat: What does this test cover that `del_adata_test` doesn't already?

safe_core/src/client/mod.rs
```diff
@@ -2028,4 +2379,604 @@ mod tests {
...
+                    client4
+                        .get_adata(adataref.name(), adataref.tag())
+                        .map(move |_data| {})
+                })
+                .then(move |res| res)
+        });
+    }
+
+    #[test]
+    pub fn get_adata_range_test() {
```
m-cat: Looks like between this function and the ones below there is a lot of duplicated test code.

mock_routing/Cargo.toml
```diff
@@ -11,7 +11,7 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-safe-nd = { git = "https://github.com/maidsafe/safe-nd" }
+safe-nd = { git = "https://github.com/lionel1704/safe-nd/", branch = "pub-field" }
```
m-cat: Did you mean to commit the safe-nd dependency changes?

safe_core/src/client/mock/vault.rs
```diff
@@ -865,10 +870,674 @@ impl Vault {
...
+            }
+            Request::GetADataRange {
+                kind,
+                address,
+                range,
+            } => {
+                let res = self
+                    .get_adata(dest, address, requester.clone())
+                    .and_then(move |data| match kind {
+                        AppendOnlyKind::PubSeq => match data {
```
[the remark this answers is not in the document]
m-cat: What I mean is, for each Request variant, we match on all four AData variants only to execute the same code.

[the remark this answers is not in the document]
Yoga07: Makes sense, but data _has_ to be fetched and matched against the expected `kind` for the `requester` to be verified(by the `verify_requester()` method). If we could find a way to perform the above actions, _only once_ for all four types, then LOC could be significantly reduced for the all the Requests.

safe_core/src/client/mod.rs
Yoga07: Just a test written initially to verify `PUT` for AData. Can be ripped off. (`get_adata_test` also covers this, hence renaming them to `put_and_get_adata_test` and `put_and_delete_adata_test` might sound to make sense)

safe_core/src/client/mod.rs
Yoga07: Most of it are initializations for datatype, permissions structs and data entries. Writing a function returning these three can reduce LOC.

mock_routing/Cargo.toml
Yoga07: Setting these dependencies to point to a different repo temporarily to make it pass CI checks. A bunch of changes were made to the safe-nd to fit this impl., which must be raised as a PR to the safe-nd master.

m-cat: Sounds good, will wait for the safe-nd changes 👍

safe_core/src/client/mod.rs
m-cat: I think a lot of these functions can be squashed as they have the same initialization and can test multiple things at once. Also, instead of re-writing the same test for each variant of adata, you could initialize the different adatas and do something like this:

```rust
for adata in &[unseq_unpub_adata, seq_unpub_adata, unseq_pub_adata, seq_pub_adata] {
    // Body of the test goes here
}
```

And where we currently have:

```rust
.get_adata_range(AppendOnlyKind::UnpubSeq, adataref, (idx_start, idx_end))
```

It would be cool if we had this instead:

```rust
.get_adata_range(adata.kind(), adataref, (idx_start, idx_end))
```

Or even better would be if `get_adata_range` was able to infer the kind from `adataref` (if possible?). Just some ideas :)

safe_core/src/client/mod.rs
m-cat: You could even have one test that puts, gets, and deletes. We already have a lot of tests like that in the existing implementation (e.g. `immutable_data_basics` and `mutable_data_basics`).

safe_core/src/client/mock/vault.rs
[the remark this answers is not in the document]
m-cat: Well you could do this with a function that takes a closure of the code to execute :)

I see that this is implemented in a different way to idata and mdata because the `kind` is included with the request. Seems like a better way of doing it!

lionel1704:
- These changes will be integrated with additional code required for the AppendOnly data implementation by @Yoga07.
- The review comments will also be addressed.

PR #862, opened 2019-06-21
Adapt to the API changes in safe-nd

### LINKED ISSUE #849: Adapt SCL to changes in safe-nd API
1. Modules like `request` and `response` in safe-nd has been made private now.
2. There are also several renamed request/response types.
3. Most of the `Request` enum variants now comprise of a single inline tuple rather than an inline struct (e.g. `GetMDataShell(Address)` instead of `GetMDataShell { address: Address }`).
4. Several structs have been refactored and renamed (e.g. `MutableDataRef` -> `Address`, etc.).

lionel1704:
- Adapt to the changes in safe-nd for data type structs and request/response types
- Add `FullId` variant to mock-routing from which the vaults fetch the requester field
- Implement proper approach to signature verification and permissions validation for mutable / immutable data
- Re-enable IData tests
Resolves #849

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -80,8 +80,10 @@ pub fn unlimited_muts(config: &Config) -> bool {
 pub struct Routing {
     vault: Arc<Mutex<Vault>>,
     sender: Sender<Event>,
+    /// old
     pub full_id: FullId,
-    full_id_new: FullIdentity,
+    /// new
+    pub full_id_new: FullIdentity,
     client_auth: Authority<XorName>,
     max_ops_countdown: Option<Cell<u64>>,
     timeout_simulation: bool,
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -80,9 +80,9 @@ pub fn unlimited_muts(config: &Config) -> bool {
 pub struct Routing {
     vault: Arc<Mutex<Vault>>,
     sender: Sender<Event>,
-    /// old
+    /// mock_routing::FullId for old types
     pub full_id: FullId,
-    /// new
+    /// safe_nd::FullIdentity for new types
     pub full_id_new: FullIdentity,
     client_auth: Authority<XorName>,
     max_ops_countdown: Option<Cell<u64>>,
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -11,7 +11,7 @@
 use super::vault::{self, Data, ImmutableDataKind, Vault, VaultGuard};
 use super::DataId;
 use crate::config_handler::{get_config, Config};
-use maidsafe_utilities::serialisation::{deserialise, serialise};
+use maidsafe_utilities::serialisation::serialise;
 use maidsafe_utilities::thread;
 use routing::{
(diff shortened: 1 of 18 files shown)


safe_core/src/client/mock/routing.rs
```diff
@@ -108,10 +122,14 @@ impl Routing {
             proxy_node_name: new_rand::random(),
         };
 
+        let bls_sk = BlsSecretKey::random();
```
nbaksalyar: I think we should have the BLS secret key in `mock_routing::FullId`? 
I.e. we can probably replace this with
```suggestion
        let bls_sk = id.map(|id| id.bls_key()).unwrap_or_else(BlsSecretKey::random);
```

safe_core/src/client/mock/tests.rs
```diff
@@ -75,44 +76,40 @@ macro_rules! expect_failure {
...
 impl Routing {
-    fn req_as_app(
-        &mut self,
-        rx: &Receiver<Event>,
-        client_mgr: Authority<XorName>,
-        app_key: PublicKey,
-        request: RpcRequest,
-    ) -> RpcResponse {
-        let msg_id = MessageId::new();
+    fn req_as_app(&mut self, rx: &Receiver<Event>, request: RpcRequest) -> RpcResponse {
```
nbaksalyar: We should be able to combine both `req_as_app` and `req_as_owner` into just `fn req(&mut self, rx: &Receiver<Event>, request: RpcRequest) -> RpcResponse` if we add a new function, `NewFullId::sign`:
```rust
impl NewFullId {
  fn sign(&self, msg: &[u8]) -> Signature {
    match self {
      NewFullId::App(app_full_id) => app_full_id.sign(msg),
      NewFullId::Client(client_full_id) => client_full_id.sign(msg),
    }
  }
}
```

Should remove some duplication and will make code easier :slightly_smiling_face:

safe_core/src/client/mock/vault.rs
```diff
@@ -835,63 +776,36 @@ impl Vault {
...
                         }
                     });
                 Response::MutateUnseqMDataEntries(result)
             }
-            _ => {
-                // Dummy return
-                // other requests to be handled by their data type impls
-                return Ok((dest, payload));
-            }
+            _ => panic!("RPC not handled"),
```
nbaksalyar: This should be OK to leave it here at this moment, but we should address this in the future and we should take note of which RPCs are not handled.

safe_core/src/client/mock/vault.rs
```diff
@@ -901,84 +815,65 @@ impl Vault {
...
-        let data_name = DataId::mutable(address.name(), address.tag());
-        // self.authorise_read(&dst, &address.name())
-        // .map_err(|err| Error::from(err.description()))
-        // .and_then(|_| match self.get_data(&data_name) {
+        let data_name = DataId::mutable(*address.name(), address.tag());
+        let sig = match signature {
+            Some(s) => s,
+            None => return Err(Error::InvalidSignature),
+        };
+        verify_signature(&sig, &requester_pk, &request, &msg_id)?;
```
nbaksalyar: I think we should be able to move this `verify_signature` call into the root function, `process_request`?
It should also make the code little bit simpler by requiring less function arguments here.

Same applies to `authorise_coin_operation`.

PR #864, opened 2019-06-22
WIP: AppendOnly Data Impl

Yoga07:

### DIFF
--- a/safe_core/src/client/mock/mod.rs
+++ b/safe_core/src/client/mock/mod.rs
@@ -69,7 +69,7 @@ impl DataId {
     }
 
     /// Create `DataId` for mutable data.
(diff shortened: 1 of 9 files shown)


safe_core/src/client/mock/mod.rs
```diff
@@ -56,11 +68,17 @@ impl DataId {
         DataId::Mutable(MutableDataId(name, tag))
     }
 
+    /// Create `DataId` for mutable data.
+    pub fn appendonly(name: XorName, tag: u64) -> Self {
```
nbaksalyar: This should be 
```suggestion
    pub fn append_only(name: XorName, tag: u64) -> Self {
```
to be consistent with the other names.

PR #867, opened 2019-06-24
Multiple improvements based on review comments

lionel1704:
- Use imports instead of mentioning the package in-line
- Use `PublicKey::from` instead of explicitly mentioning `PublicKey::Bls`
- Impl `Into` trait to convert `AppKeys` into `AppFullId`
- Re-use BLS key from `mock_routing::FullId`
- Combine `req_as_app` and `req_as_owner` functions for tests
- Use common signature verification for all requests

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -17,8 +17,10 @@ use routing::{
 };
 #[cfg(any(feature = "testing", test))]
 use safe_nd::Coins;
-use safe_nd::{AppFullId, ClientFullId};
-use safe_nd::{IDataKind, ImmutableData, Message, MessageId, PublicId, PublicKey, XorName};
+use safe_nd::{
+    AppFullId, ClientFullId, IDataKind, ImmutableData, Message, MessageId, PublicId, PublicKey,
+    Signature, XorName,
+};
 use std;
 use std::cell::Cell;
 use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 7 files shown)


safe_core/src/client/mock/routing.rs
```diff
@@ -120,7 +132,10 @@ impl Routing {
             proxy_node_name: new_rand::random(),
         };
 
-        let bls_sk = BlsSecretKey::random();
+        let bls_sk = id
+            .as_ref()
```
m-cat: Is `as_ref` needed here? Looks like you're just creating a clone of the internal key.

safe_core/src/client/mock/vault.rs
```diff
@@ -794,7 +750,7 @@ impl Vault {
                     });
                 Response::MutateUnseqMDataEntries(result)
             }
-            _ => panic!("RPC not handled"),
+            _other => panic!("{:?} RPC not handled", _other),
```
m-cat: Variables should only start with an underscore if they are not used, surprised clippy didn't complain about this.

safe_core/src/client/mock/routing.rs
lionel1704: I needed to use as_ref here as id is an `Option` and `unwrap_or_else` causes a `move`.

safe_core/src/client/mock/vault.rs
lionel1704: I just googled this and learnt that I can use any identifier as a wildcard :smile: 
I'll update this to `other`, thanks :+1:

nbaksalyar:
This PR addressed review comments on https://github.com/maidsafe/safe_client_libs/pull/862#pullrequestreview-252834306

PR #869, opened 2019-06-25
feat/adata: implement AppendOnly Data RPCs and support AppendOnly Data in Mock Vault

Yoga07:
- Implements feature complete AppendOnly Data RPCs
- Implements AppendOnly Data in Mock Vault
- Adds tests for AppendOnly Data RPCs

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -14,9 +14,11 @@ use fs2::FileExt;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use routing::{Authority, ClientError, MutableData as OldMutableData};
 use safe_nd::{
-    verify_signature, Coins, Error, IDataKind, MDataAddress, Message,
-    MutableData as NewMutableData, PublicId, PublicKey, Request, Response, SeqMutableData,
-    Transaction, UnseqMutableData, XorName,
+    verify_signature, AData, ADataAddress, ADataIndex, AppendOnlyData, Coins, Error, IDataKind,
+    MDataAddress, Message, MutableData as NewMutableData, PubSeqAppendOnlyData,
+    PubUnseqAppendOnlyData, PublicId, PublicKey, Request, Response, SeqAppendOnly, SeqMutableData,
+    Transaction, UnpubSeqAppendOnlyData, UnpubUnseqAppendOnlyData, UnseqAppendOnly,
+    UnseqMutableData, XorName,
 };
 use std::collections::HashMap;
 use std::env;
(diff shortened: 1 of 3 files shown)


Yoga07:
> A quick note here @Yoga07.
> The `safe-nd::AData` and the `AppendOnlyDataKind` enums are exactly the same. So I guess we can change the `Data::AppendOnly` variant to hold `safe-nd::AData` instead of `AppendOnlyKind`.
> 
> I'll leave it to you to change it in this PR or a separate one.

Ah nice catch @lionel1704 , I am thinking we could resolve this in the next PR along with the same improvisation for `MutableDataKind` as well.

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                                    id,
+                                    Data::AppendOnly(AppendOnlyDataKind::PubSeq(adata)),
+                                );
+                                Ok(())
+                            }
+                            _ => Err(Error::NoSuchData),
+                        },
+                        ADataAddress::PubUnseq { .. } => match data {
+                            AData::PubUnseq(mut adata) => {
+                                unwrap!(adata.append_owner(owner));
```
nbaksalyar: This should also not be `unwrap!`.

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                                    id,
+                                    Data::AppendOnly(AppendOnlyDataKind::PubUnseq(adata)),
+                                );
+                                Ok(())
+                            }
+                            _ => Err(Error::NoSuchData),
+                        },
+                        ADataAddress::UnpubSeq { .. } => match data.clone() {
+                            AData::UnpubSeq(mut adata) => {
+                                unwrap!(adata.append_owner(owner));
```
nbaksalyar: Same comment about `unwrap!`

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                                    id,
+                                    Data::AppendOnly(AppendOnlyDataKind::UnpubSeq(adata)),
+                                );
+                                Ok(())
+                            }
+                            _ => Err(Error::NoSuchData),
+                        },
+                        ADataAddress::UnpubUnseq { .. } => match data {
+                            AData::UnpubUnseq(mut adata) => {
+                                unwrap!(adata.append_owner(owner));
```
nbaksalyar: Same comment about `unwrap!`

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                        },
+                        _ => Err(Error::AccessDenied),
+                    });
+                Response::AddUnpubADataPermissions(res)
+            }
+            Request::SetADataOwner { address, owner } => {
+                let id = DataId::append_only(*address.name(), address.tag());
+                let res = self
+                    .get_adata(address, requester_pk, request)
+                    .and_then(move |data| match address {
```
nbaksalyar: We can try to get rid of code duplication here by moving the logic of checking whether the `address` matches the stored data type to `get_adata` - please see how it is implemented for `MutableData`.

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
         })
     }
+    //
+    // ======= Append Only Data =======
+    //
+    pub fn get_adata(
+        &mut self,
+        data: ADataAddress,
+        requester_pk: PublicKey,
+        request: Request,
```
nbaksalyar: This should be
```suggestion
        request: &Request,
```
to prevent unnecessary data copying.

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                Response::GetADataOwners(res)
+            }
         };
-
         Ok(Message::Response {
             response,
             message_id,
         })
     }
+    //
```
nbaksalyar: Could you please add a blank line above to have a uniform code style?

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+        data: ADataAddress,
+        requester_pk: PublicKey,
+        request: Request,
+    ) -> Result<AData, Error> {
+        let data_name = DataId::append_only(*data.name(), data.tag());
+        match self.get_data(&data_name) {
+            Some(data_type) => match data_type {
+                Data::AppendOnly(kind) => match kind {
+                    AppendOnlyDataKind::PubSeq(data) => {
+                        if data.check_permission(&request, requester_pk).is_ok() {
```
nbaksalyar: It is not very apparent that `get_adata` function also checks the permissions. I think it's better to decouple it and move to a separate function - but for now we can take a note of that and implement in a separate PR.

safe_core/src/client/mock/vault.rs
```diff
@@ -942,6 +1355,14 @@ impl MutableDataKind {
     }
 }
 
+#[derive(Clone, Deserialize, Serialize)]
+pub enum AppendOnlyDataKind {
```
nbaksalyar: Can't we use enum defined in safe-nd?
https://github.com/maidsafe/safe-nd/blob/c8c7fd792c92caae963e4d759194b8ca39a5b5a0/src/request.rs#L22

safe_core/src/client/mock/vault.rs
```diff
@@ -752,14 +754,424 @@ impl Vault {
...
+                Data::AppendOnly(kind) => match kind {
+                    AppendOnlyDataKind::PubSeq(data) => {
+                        if data.check_permission(&request, requester_pk).is_ok() {
+                            Ok(AData::PubSeq(data))
+                        } else {
+                            Err(Error::AccessDenied)
+                        }
+                    }
+                    AppendOnlyDataKind::UnpubSeq(data) => {
+                        if data.check_permission(&request, requester_pk).is_ok() {
```
nbaksalyar: I think we should have `check_permission` implemented for the `safe_nd::AppendOnlyData` enum to prevent code duplication.

safe_core/src/client/mock/vault.rs
m-cat: These lines strike me as not being very good Rust style. It seems like you are taking a `Result`, matching on an `if`, and returning another `Result` - so why not use `map` and `map_err` on the original `Result`? (`map_err` might not even be needed here)

safe_core/src/client/mod.rs
```diff
@@ -861,6 +863,362 @@ pub trait Client: Clone + 'static {
...
+
+    /// Get a Set of Entries for the requested range from an AData.
+    #[allow(clippy::type_complexity)]
+    fn get_adata_range(
+        &self,
+        address: ADataAddress,
+        range: (ADataIndex, ADataIndex),
+    ) -> Box<CoreFuture<Vec<(Vec<u8>, Vec<u8>)>>> {
+        trace!(
+            "Get Rage of entries from AppendOnly Data at {:?}",
```
m-cat: Typo: "Rage"

safe_core/src/client/mod.rs
```diff
@@ -861,6 +863,362 @@ pub trait Client: Clone + 'static {
...
+            match res {
+                Response::GetADataShell(res) => res.map_err(CoreError::from),
+                _ => Err(CoreError::ReceivedUnexpectedEvent),
+            }
+        })
+        .into_box()
+    }
+
+    /// Get a Set of Entries for the requested range from an AData.
+    #[allow(clippy::type_complexity)]
```
m-cat: Hmm, the return value of this function is very convoluted and hard to understand. I think it would be better to give it a name with a `type` statement, and then we can get rid of this clippy exception!

nbaksalyar:
I merged this now - the review comments will be addressed in a separate PR.

Issue #843, opened 2019-06-14
It's not possible to get the balance from a CoinBalance different than the default owned by the account


bochaco:
In `experimental` branch, It's not possible to get the balance from a `CoinBalance` different than the default owned by the account since the `get_balance` is missing an SK arg for arbitrary `CoinBalance` retrieval:
`fn get_balance(&self, destination: XorName) -> Box<CoreFuture<Coins>>`

https://github.com/maidsafe/safe_client_libs/blob/20b945dc6fc7b8684fd15fde78948240db86503d/safe_core/src/client/mod.rs#L299

lionel1704:
Merged into experimental in #871

PR #871, opened 2019-06-25
Add safecoin features and fix GetMDataValue

### LINKED ISSUE #842: Client function `get_seq_mdata_value` panics when trying to fetch a non-existing entry
In `experimental` branch, client function `get_seq_mdata_value` panics when trying to fetch a non-existing entry. Looking at the code the problem seems to be an `unwrap` here: https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_core/src/client/mock/vault.rs#L532

### LINKED ISSUE #850: [Experimental] Cannot transfer from arbitrary coin balance
At the moment, we cannot create a transfer from any coin balance as there's no `sk` field. Right now: https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_core/src/client/mod.rs#L275-L282

```rust
fn transfer_coins(
        &self,
        source: XorName,
        destination: XorName,
        amount: Coins,
        transaction_id: Option<u64>,
    ) -> Box<CoreFuture<()>> { 
//... 
}
```

But we need to be able to pass in a `sk` for wallet functionality. So we need something like:

```rust
fn transfer_coins(
        &self,
        from_pk: &PublicKey,
        from_sk: &SecretKey,
        to_pk: &PublicKey,
        transaction_id: ....
        amount: &str,
    ) { 
//... 
}
```

We were assuming the need to pass the keys themselves, if XorName works for Public Key that should be fine (Aside from for `sk`, where we'll need to pass it in)

### LINKED ISSUE #852: Allocate test coins
Currently we have no method to allocate test coins to an existing balance.

eg, current CLI impl: 
```rust
    pub fn allocate_test_coins(&mut self, to_pk: &PublicKeyMock, amount: &str) -> XorName {
```

lionel1704:
- Add API to load a wallet with test safecoin
- Add optional SecretKey field for safecoin operations
- Gracefully return an error when trying to fetch a non-existent entry in MData
resolves #842 #843 #852 #850

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -15,12 +15,12 @@ use routing::{
     Authority, BootstrapConfig, ClientError, EntryAction, Event, FullId, InterfaceError,
     MutableData, PermissionSet, Request, Response, RoutingError, User, TYPE_TAG_SESSION_PACKET,
 };
-#[cfg(any(feature = "testing", test))]
-use safe_nd::Coins;
 use safe_nd::{
     AppFullId, ClientFullId, IDataKind, ImmutableData, Message, MessageId, PublicId, PublicKey,
     Signature, XorName,
 };
+#[cfg(any(feature = "testing", test))]
+use safe_nd::{Coins, Error};
 use std;
 use std::cell::Cell;
 use std::collections::{BTreeMap, BTreeSet};
@@ -1065,6 +1065,16 @@ impl Routing {
     ) {
         let mut vault = self.lock_vault(true);
         vault.mock_create_balance(coin_balance_name, amount, owner);
+    }
+
+    /// Add some coins to a wallet's PublicKey
+    pub fn allocate_test_coins(
+        &self,
+        coin_balance_name: &XorName,
+        amount: Coins,
+    ) -> Result<(), Error> {
+        let mut vault = self.lock_vault(true);
+        vault.mock_increment_balance(coin_balance_name, amount)
     }
 }
 
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -515,10 +515,18 @@ impl Vault {
 
                 match (address, result) {
                     (MDataAddress::Seq { .. }, Ok(MutableDataKind::Sequenced(mdata))) => {
-                        Response::GetSeqMDataValue(Ok(mdata.get(&key).unwrap().clone()))
+                        let res = match mdata.get(&key) {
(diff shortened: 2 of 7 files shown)


safe_core/src/client/mod.rs
```diff
@@ -260,45 +260,61 @@ pub trait Client: Clone + 'static {
     fn transfer_coins(
         &self,
         source: XorName,
+        secret_key: Option<threshold_crypto::SecretKey>,
```
nbaksalyar: This should be `&threshold_crypto::SecretKey` - we don't want to copy private keys in memory.

safe_core/src/client/mod.rs
```diff
@@ -1493,6 +1509,33 @@ pub trait Client: Clone + 'static {
...
+    fn allocate_test_coins(&self, coin_balance_name: &XorName, amount: Coins) {
+        let inner = self.inner();
+        inner
+            .borrow_mut()
+            .routing
+            .allocate_test_coins(coin_balance_name, amount);
+    }
+}
+
+fn sign_request_with_key(request: Request, key: threshold_crypto::SecretKey) -> Message {
```
nbaksalyar: This should be
```suggestion
fn sign_request_with_key(request: &Request, key: &threshold_crypto::SecretKey) -> Message {
```

safe_core/src/client/mock/routing.rs
```diff
@@ -1066,6 +1066,12 @@ impl Routing {
         let mut vault = self.lock_vault(true);
         vault.mock_create_balance(coin_balance_name, amount, owner);
     }
+
+    /// Add some coins to a wallet's PublicKey
+    pub fn allocate_test_coins(&self, coin_balance_name: &XorName, amount: Coins) {
```
m-cat: Why `unwrap` here instead of returning an error from the function, e.g. if the given name doesn't exist?

safe_core/src/client/mock/vault.rs
```diff
@@ -170,6 +189,19 @@ impl Vault {
...
             }
         };
+        // Check if the request was signed by owner of the wallet
+        match verify_signature(
+            &signature,
+            &PublicKey::from(*balance.owner()),
+            request,
+            &message_id,
+        ) {
+            Ok(()) => return Ok(()),
```
m-cat: Maybe instead of matching here, you could do something like:

```rust
if verify_signature(...).is_ok() {
    return Ok(());
}

// Continue with verification
verify_signature(...);
```

This would be a bit cleaner and easier to understand, imo.

Issue #844, opened 2019-06-16
Remove authorities from the Mock Vault implementation


nbaksalyar:
We won't have src/dst Authorities anymore in the real Vault.
Therefore, we also need to remove them from the Mock Vault.
When handling requests in Mock Vault, we should always assume they come from the clients (i.e. what was `Authority::Client(..)` before), and check the permissions accrodingly. Similarly, what was the `dst` authority will be assumed to be src Elders now (comprising the functionality from `Authority::ClientManager` and `Authority::NaeManager`).
This makes mock Routing unnecessary, but we should transition it incrementally rather than in a single task; thus, the `MockRouting::send` function should remain, but it doesn't need `dst` and `src` arguments anymore - it will be just `send(Vec<u8>)`, roughly corresponding to the `send` function in quic-p2p.

nbaksalyar:
Implemented in #873

PR #873, opened 2019-06-26
Implement CreateCoinBalance RPC and remove src parameter for safecoin transactions

### LINKED ISSUE #851: Create Coin Balance. Preload from another balance missing
For the CLI we need the ability to create arbitrary coin balances.

These should be able to be preloaded from another balance. OR with `mock` preload `test-coins`.

Currently we can only create balances to fill with test coins.

Wanted:

- Ability to create arbitrary coin balances outside of `mock` mode
- Ability to preload coins from another existing coin balance address.
- Ability to choose to preload `test-coins`

eg (current sig in CLI):
```rust
pub fn create_balance(
        &mut self,
        from_pk: &PublicKeyMock,
        from_sk: &SecretKeyMock,
        new_balance_owner: &PublicKeyMock,
        amount: &str,
    ) 
```

Our current impl is super simple re: coins, so there's no real 'switch' for test vs real coins. Not sure how best this would look for this api.

lionel1704:
- Implement create_coin_balace that creates a new coin balance for a
given address and public key, preloaded with some safecoin from an
existing wallet. If the existing wallet does not use the client's owner
key then the secret key must be specified
- Remove source fields from safecoin transactions, the wallet address is
derived from the Public Id of the client
- Replace the dest parameter in routing::send with a requester field to
send an optional PublicKey in case of arbitrary wallet transactions
- Refactor the handling of wallet transactions in the mock vault to
adapt to the above changes
Resolves #851, #844

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -148,7 +148,7 @@ impl Vault {
         &mut self,
         coin_balance_name: &XorName,
         amount: Coins,
-        owner: threshold_crypto::PublicKey,
+        owner: PublicKey,
     ) {
         let _ = self
             .cache
@@ -190,19 +190,12 @@ impl Vault {
             }
         };
         // Check if the request was signed by owner of the wallet
-        if verify_signature(
-            &signature,
-            &PublicKey::from(*balance.owner()),
-            request,
-            &message_id,
-        )
-        .is_ok()
-        {
+        if verify_signature(&signature, balance.owner(), request, &message_id).is_ok() {
             return Ok(());
         }
         verify_signature(&signature, &requester_pk, request, &message_id)?;
-        let owner_account = XorName::from(PublicKey::from(*balance.owner()));
-        if PublicKey::from(*balance.owner()) == requester_pk {
+        let owner_account = XorName::from(*balance.owner());
+        if *balance.owner() == requester_pk {
             Ok(())
         } else {
             let account = match self.get_account(&owner_account) {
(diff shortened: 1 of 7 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -190,19 +190,12 @@ impl Vault {
...
         // Check if the request was signed by owner of the wallet
-        if verify_signature(
-            &signature,
-            &PublicKey::from(*balance.owner()),
-            request,
-            &message_id,
-        )
-        .is_ok()
-        {
+        if verify_signature(&signature, balance.owner(), request, &message_id).is_ok() {
```
m-cat: Looks like we call `balance.owner` several times, is it possible to store the return value in a variable?

safe_core/src/client/mod.rs
```diff
@@ -287,6 +287,42 @@ pub trait Client: Clone + 'static {
         .into_box()
     }
 
+    /// Creates a new coin balance on the network.
+    fn create_coin_balance(
+        &self,
+        source: XorName,
```
nbaksalyar: Probably we don't need the `source` here; it can be derived from the `secret_key` parameter.
I.e., if `secret_key` is None, it's the default user's account (identity stored in `FullId`). If it's `Some(secret_key)`, the source is going to be the coin balance corresponding to the pub key (`XorName::from(secret_key.public_key())`).

We can then set the `requester` to the value we need when calling `vault.process_request(..)`.

Same applies for other coin-related functions (`transfer_balance` and `get_balance`).

PR #874, opened 2019-06-26
Remove authorities from mock_routing::send

### LINKED ISSUE #844: Remove authorities from the Mock Vault implementation

lionel1704:
resolves #844

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -152,7 +152,7 @@ impl Routing {
     }
 
     /// Send a routing message
-    pub fn send(&mut self, dst: Authority<XorName>, payload: &[u8]) -> Result<(), InterfaceError> {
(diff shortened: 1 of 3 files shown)


lionel1704:
These changes have been incorporated in #873

PR #876, opened 2019-07-02
fix/mock-vault: fix Mutable data EntryActions and refactor AppendOnlyData impl 

Yoga07:


mock_routing/Cargo.toml
```diff
@@ -11,7 +11,7 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-safe-nd = { git = "https://github.com/maidsafe/safe-nd" }
+safe-nd = { git = "https://github.com/maidsafe/safe-nd"}
```
nbaksalyar: I reckon we don't need to make this change?
This applies to the rest of `Cargo.toml` files.

safe_core/src/client/mod.rs
```diff
@@ -2529,16 +2529,12 @@ mod tests {
...
-                    let _ = entry_actions.insert(
-                        b"key1".to_vec(),
-                        SeqEntryAction::Update(Val::new(b"newValue".to_vec(), 1)),
-                    );
-                    let _ = entry_actions.insert(b"key2".to_vec(), SeqEntryAction::Del(1));
-                    let _ = entry_actions.insert(
-                        b"key3".to_vec(),
-                        SeqEntryAction::Ins(Val::new(b"value".to_vec(), 0)),
-                    );
+                    let entry_actions: MDataSeqEntryActions = MDataSeqEntryActions::new()
```
lionel1704: Looks like the logic of the test is changed a bit here?
If this is the case can you please update the test decription if required?

mock_routing/Cargo.toml
Yoga07: Ah a copy paste error 😄 fixing it!

safe_core/src/client/mod.rs
```diff
@@ -2529,16 +2529,12 @@ mod tests {
...
+                        .ins(b"key1".to_vec(), b"newValue".to_vec(), 0)
```
lionel1704: Here we are performing an insert using a key that already exists.
This should fail. This might be an issue with the mutate function in safe-nd, can you please check this?

Yoga07: Hmm, the safe_app tests are failing in CI. Looks like the safe-nd part is correct and the test is wrong. I'll get that updated now. Thanks.

safe_core/src/client/mod.rs
Yoga07: Reverted the test logic during the fix. Therefore this change can be resolved.

safe_core/src/client/mod.rs
```diff
@@ -2529,16 +2529,10 @@ mod tests {
...
-                        b"key1".to_vec(),
-                        SeqEntryAction::Update(Val::new(b"newValue".to_vec(), 1)),
-                    );
-                    let _ = entry_actions.insert(b"key2".to_vec(), SeqEntryAction::Del(1));
-                    let _ = entry_actions.insert(
-                        b"key3".to_vec(),
-                        SeqEntryAction::Ins(Val::new(b"value".to_vec(), 0)),
-                    );
+                    let entry_actions: MDataSeqEntryActions = MDataSeqEntryActions::new()
+                        .update(b"key1".to_vec(), b"newValue".to_vec(), 0)
```
lionel1704: Version should be incremented when updating an entry

safe_core/src/client/mod.rs
```diff
@@ -2550,17 +2544,17 @@ mod tests {
                         .map(move |fetched_entries| {
                             let mut expected_entries: BTreeMap<_, _> = Default::default();
                             let _ = expected_entries
-                                .insert(b"key1".to_vec(), Val::new(b"newValue".to_vec(), 1));
+                                .insert(b"key1".to_vec(), Val::new(b"newValue".to_vec(), 0));
                             let _ = expected_entries
-                                .insert(b"key3".to_vec(), Val::new(b"value".to_vec(), 0));
+                                .insert(b"key3".to_vec(), Val::new(b"value".to_vec(), 1));
```
lionel1704: Version should be 0, since the entry was not updated after insertion.

safe_core/src/client/mod.rs
```diff
@@ -2550,17 +2544,17 @@ mod tests {
...
                             assert_eq!(fetched_entries, expected_entries);
                         })
                 })
                 .and_then(move |_| {
                     client5
-                        .get_seq_mdata_value(name, tag, b"key1".to_vec())
+                        .get_seq_mdata_value(name, tag, b"key3".to_vec())
                         .and_then(|fetched_value| {
-                            assert_eq!(fetched_value, Val::new(b"newValue".to_vec(), 1));
+                            assert_eq!(fetched_value, Val::new(b"updatedValue".to_vec(), 1));
```
lionel1704: `key3`s value field should be `value`. The test should fail :confused:

Yoga07: Hmm, Your are right, but the CI's are passing. Something's fishy with the tests themselves. I'll have a look at it now. 🙂

PR #886, opened 2019-07-05
 refactor/data: avoid namespace collision and type erasure in DataId 

### LINKED ISSUE #883: Published and unpublished variants of the data types should not use the same namespace
In the current implementation, for mutable data and append-only data published and unpublished variants use the same namespace. So if a user tries to store published mutable data with an address and tag and then another (or the same) client tries to store unpublished mutable data at the same address space there will be a conflict.

This should not be the case, published and unpublished variants of mutable and append-only data should be stored in their own address spaces.

As a result, the `UnexpectedDataReturned` would not be required anymore and should be removed.

EDIT: When sequenced / unsequenced types do not match `Error:NoSuchData` is returned

### LINKED ISSUE #884: Avoid type erasure from the mock vault
Currently, the `DataId` enum in the mock vault has only Mutable, AppendOnly and Immutable types. As mentioned in #883, published and unpublished variants will not reuse the same address space. To facilitate this and to also avoid type erasure the DataId enum should be:
```
enum DataId {
    PubImmutable(address),
    UnpubImmutable(address),
    UnPubMutable(address, tag),
    PubAppendOnly(address, tag),
    UnpubAppendOnly(address, tag),
    ...
}
```
The `Data` enum in the mock vault should also follow the same pattern.

m-cat:
Fixes #883, fixes #884.

### DIFF
--- a/safe_core/src/client/mock/mod.rs
+++ b/safe_core/src/client/mock/mod.rs
@@ -17,25 +17,29 @@ pub use self::account::{Account, CoinBalance, DEFAULT_MAX_MUTATIONS};
 pub use self::routing::{NewFullId, RequestHookFn, Routing};
 
 use ::routing::XorName;
+use safe_nd::{ADataAddress, IDataAddress, MDataAddress};
 
 /// Identifier for a data.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
 pub enum DataId {
+    /// Identifier of old mutable data.
+    OldMutable { name: XorName, tag: u64 },
     /// Identifier of immutable data.
-    Immutable(XorName),
+    Immutable(IDataAddress),
     /// Identifier of mutable data.
-    Mutable { name: XorName, tag: u64 },
+    Mutable(MDataAddress),
     /// Identifier of appendonly data.
-    AppendOnly { name: XorName, tag: u64 },
+    AppendOnly(ADataAddress),
 }
 
 impl DataId {
     /// Get name of this identifier.
     pub fn name(&self) -> &XorName {
         match *self {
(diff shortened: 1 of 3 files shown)


mock_routing/src/lib.rs
```diff
@@ -1046,19 +1046,19 @@ pub enum Response {
         msg_id: MsgId,
     },
 
-    // --- ImmutableData ---
+    // --- PubImmutableData ---
```
m-cat: This line should still say `ImmutableData`

safe_app/src/errors.rs
```diff
@@ -438,7 +438,6 @@ fn core_error_code(err: &CoreError) -> i32 {
             SndError::InvalidEntryActions(_) => ERR_INVALID_ENTRY_ACTIONS,
             SndError::NoSuchKey => ERR_NO_SUCH_KEY,
             SndError::DuplicateMessageId => ERR_DUPLICATE_MSG_ID,
-            SndError::UnexpectedDataReturned => ERR_UNEXPECTED_DATA,
```
lionel1704: Can you please add this line back. We can remove it along when adapting to the other changes in safe-nd.

safe_authenticator/src/errors.rs
```diff
@@ -325,7 +325,6 @@ fn core_error_code(err: &CoreError) -> i32 {
             SndError::InvalidEntryActions(_) => ERR_INVALID_ENTRY_ACTIONS,
             SndError::NoSuchKey => ERR_NO_SUCH_KEY,
             SndError::DuplicateMessageId => ERR_DUPLICATE_MSG_ID,
-            SndError::UnexpectedDataReturned => ERR_UNEXPECTED_DATA,
```
lionel1704: Same as above.

PR #888, opened 2019-07-10
Implement login packet operations

### LINKED ISSUE #858: Implement account handling as per the Safecoin RFC
1. Introduce a notion of `Account` data type (not to be confused with the `Account` structure held by Vaults). This special data type replaces MD with type tag 0 from the previous design: an account is just a location on the network that holds arbitrary data (`Vec<u8>`), with the max. size of 1 MB. It also has a `XorName` as the rest of data types.
2. Support `CreateAccount` and `UpdateAccount` RPCs in mock Vaults. It should respectively create or update the account with the specified destination address.
3. Support `GetAccount` RPC in mock Vault. It should return the `Vec<u8>` data stored within the account.

lionel1704:
- Adapt to the latest changes in safe-nd
- Handle RPCs for login packet operations
- Add `create_account` and `get_account` public API to establish new routing instances and perform the required operations

closes #858

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -9,7 +9,7 @@
 use super::vault::{self, Data, Vault, VaultGuard};
 use super::DataId;
 use crate::config_handler::{get_config, Config};
-use maidsafe_utilities::serialisation::serialise;
+use maidsafe_utilities::serialisation::{deserialise, serialise};
 use maidsafe_utilities::thread;
 use routing::{
     Authority, BootstrapConfig, ClientError, EntryAction, Event, FullId, InterfaceError,
@@ -17,7 +17,8 @@ use routing::{
 };
 use safe_nd::{
     AppFullId, ClientFullId, ClientPublicId, IData, IDataAddress, Message, MessageId,
-    PubImmutableData, PublicId, PublicKey, Signature, XorName,
+    PubImmutableData, PublicId, PublicKey, Request as RpcRequest, Response as RpcResponse,
+    Signature, XorName,
 };
 #[cfg(any(feature = "testing", test))]
 use safe_nd::{Coins, Error};
@@ -25,7 +26,7 @@ use std;
 use std::cell::Cell;
 use std::collections::{BTreeMap, BTreeSet};
 use std::env;
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{Receiver, Sender};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use threshold_crypto::SecretKey as BlsSecretKey;
@@ -58,6 +59,28 @@ const CHANGE_MDATA_OWNER_DELAY_MS: u64 = DEFAULT_DELAY_MS;
 
 lazy_static! {
     static ref VAULT: Arc<Mutex<Vault>> = Arc::new(Mutex::new(Vault::new(get_config())));
+}
+
+/// Helper macro to receive a routing event and assert it's a response
+/// success.
+#[macro_export]
+macro_rules! expect_success {
+    ($rx:expr, $msg_id:expr, $res:path) => {
+        match unwrap!($rx.recv_timeout(Duration::from_secs(10))) {
+            Event::Response {
+                response: $res { res, msg_id },
+                ..
+            } => {
+                assert_eq!(msg_id, $msg_id);
+
+                match res {
+                    Ok(value) => value,
+                    Err(err) => panic!("Unexpected error {:?}", err),
+                }
+            }
+            event => panic!("Unexpected event {:?}", event),
+        }
+    };
 }
 
 /// Creates a thread-safe reference-counted pointer to the global vault.
@@ -189,6 +212,24 @@ impl Routing {
         self.send_response(DEFAULT_DELAY_MS, dummy_authority, dummy_authority, response);
 
         Ok(())
+    }
+
+    /// Send a request and get a response
+    pub fn req(&mut self, rx: &Receiver<Event>, request: RpcRequest) -> RpcResponse {
+        let message_id = MessageId::new();
+        let signature = self
+            .full_id_new
+            .sign(&unwrap!(bincode::serialize(&(&request, message_id))));
+        unwrap!(self.send(
+            None,
+            &unwrap!(serialise(&Message::Request {
+                request,
+                message_id,
+                signature: Some(signature),
+            }))
+        ));
+        let response = expect_success!(rx, message_id, Response::RpcResponse);
+        unwrap!(deserialise(&response))
     }
 
     /// Sets the vault for this routing instance.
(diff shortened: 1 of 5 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -173,7 +173,7 @@ impl Vault {
             Some(balance) => balance,
             None => {
                 debug!("Account not found for {:?}", coin_balance_name);
```
nbaksalyar: Should be
```suggestion
                 debug!("Coin balance not found for {:?}", coin_balance_name);
```
for consistency?

safe_core/src/client/mock/vault.rs
```diff
@@ -405,7 +394,7 @@ impl Vault {
                         account.version(),
                     )))
                 } else {
-                    return Err(SndError::NoSuchAccount);
+                    return Err(SndError::NoSuchData);
```
nbaksalyar: In `put_data` (line 1203) we return `AccessDenied` if an account has not been found.
Probably `AccessDenied` should be returned here too?

safe_core/src/client/mock/vault.rs
```diff
@@ -417,15 +406,15 @@ impl Vault {
                 if let Some(account) = self.get_account_mut(&name) {
                     Response::Mutation(account.ins_auth_key(key, permissions, version))
                 } else {
-                    return Err(SndError::NoSuchAccount);
+                    return Err(SndError::NoSuchData);
```
nbaksalyar: Same thing about `AccessDenied`.

safe_core/src/client/mock/vault.rs
```diff
@@ -417,15 +406,15 @@ impl Vault {
...
                 }
             }
             Request::DelAuthKey { key, version } => {
                 let name = requester.name();
                 if let Some(account) = self.get_account_mut(&name) {
                     Response::Mutation(account.del_auth_key(&key, version))
                 } else {
-                    return Err(SndError::NoSuchAccount);
+                    return Err(SndError::NoSuchData);
```
nbaksalyar: Same thing about `AccessDenied`.

safe_core/src/client/mod.rs
```diff
@@ -2287,15 +2279,19 @@ mod tests {
...
+                    assert_eq!(transaction.amount, unwrap!(Coins::from_str("100")));
                     client2
                         .transfer_coins(
                             Some(&bls_sk),
                             wallet1,
                             unwrap!(Coins::from_str("5.0")),
                             None,
                         )
-                        .and_then(move |_| Ok(()))
+                        .and_then(move |transaction| {
```
nbaksalyar: `.and_then` block can be replaced with
```rust
.map(move |transaction| {
  assert_eq!(transaction.amount, unwrap!(Coins::from_str("5.0")));
})
```

safe_core/src/client/mock/account.rs
```diff
@@ -163,3 +163,32 @@ impl Account {
         }
     }
 }
+
+#[derive(Deserialize, Serialize)]
+pub struct LoginData {
```
nbaksalyar: This struct looks almost the same as [`LoginPacket`](https://github.com/maidsafe/safe-nd/blob/master/src/request/login_packet.rs#L18) in safe-nd (apart from the `location` field).
Can't we use `LoginPacket` here instead?

safe_core/src/client/mock/tests.rs
```diff
@@ -76,25 +75,6 @@ macro_rules! expect_failure {
     };
 }
 
-impl Routing {
```
nbaksalyar: I think now we can import the `expect_success` macro from mock-routing and remove it here too?

safe_core/src/client/mock/vault.rs
```diff
@@ -417,6 +434,7 @@ impl Vault {
                     return Err(SndError::AccessDenied);
                 }
             }
+            // ===== Safecoin =====
```
nbaksalyar: Probably would be better to replace it with just `Coins`.

safe_core/src/client/mock/vault.rs
```diff
@@ -463,6 +481,92 @@ impl Vault {
...
+                if let Err(e) = self.authorise_coin_operation(&source, requester_pk) {
+                    Response::Mutation(Err(e))
+                } else if self.get_login_packet(account_data.destination()).is_some() {
+                    Response::Mutation(Err(SndError::LoginPacketExists))
+                } else {
+                    let res = self
+                        .get_balance(&source)
+                        .and_then(|source_balance| {
+                            if source_balance
+                                .checked_sub(unwrap!(Coins::from_nano(1)))
```
nbaksalyar: I would suggest defining it as a constant, `pub const COST_OF_PUT: Coins = <..>`, similarly to [how it works in the Vaults](https://github.com/maidsafe/safe_vault/blob/436c37a72b56541f7f89d3eff8c84d1bc6532775/src/source_elder.rs#L43).

safe_core/src/client/mod.rs
```diff
@@ -1578,6 +1616,93 @@ pub trait Client: Clone + 'static {
     }
 }
 
+/// Create an account given a locator and passsword and the secret key of an exitsing wallet
```
nbaksalyar: Typo: exitsing -> _existing_

Probably also worth replacing `wallet` with `coin balance`?

safe_core/src/client/mock/account.rs
m-cat: Style question: instead of having a constructor and getters that seem not to do much, why not make the member fields public?

safe_core/src/client/mock/routing.rs
```diff
@@ -60,6 +61,25 @@ lazy_static! {
     static ref VAULT: Arc<Mutex<Vault>> = Arc::new(Mutex::new(Vault::new(get_config())));
 }
 
+macro_rules! expect_success {
```
m-cat: This looks like the same macro as in mock/tests.rs. We shouldn't be duplicating it, and perhaps it shouldn't be used outside of tests since it contains `unwrap` and `panic` calls.

safe_core/src/client/mock/vault.rs
```diff
@@ -172,8 +188,8 @@ impl Vault {
         let balance = match self.get_coin_balance_mut(coin_balance_name) {
             Some(balance) => balance,
             None => {
-                debug!("Account not found for {:?}", coin_balance_name);
-                return Err(SndError::NoSuchAccount);
+                debug!("Balance found for {:?}", coin_balance_name);
```
m-cat: Should be "Balance not found"

safe_core/src/client/mock/routing.rs
nbaksalyar: I'd suggest moving it to `test_utils` probably, but I would also consider `mock::routing` module used exclusively in tests, so it's fine to use `unwrap!`s there.

safe_core/src/client/mock/vault.rs
```diff
@@ -474,13 +481,92 @@ impl Vault {
                     Response::GetBalance(res)
                 }
             }
-            Request::GetTransaction {
-                coins_balance_id,
+            // ===== Account =====
+            Request::CreateLoginPacketFor {
```
m-cat: I like the code here and it's easy to follow, but I think that adding some comments would be helpful.

PR #891, opened 2019-07-10
feat/adata: add GetADataValue RPC in Mock-Vault

### LINKED ISSUE #879: Add GetADataValue RPC
As mentioned in maidsafe/rfcs#315, it would be good to support fetching the value of a specific key in an AData. For this,

- Add the required request and response types in safe-nd
- Include a `get_data_value` function to the client trait with relevant tests

Yoga07:
Resolves #879

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -173,7 +173,7 @@ impl Vault {
             Some(balance) => balance,
             None => {
                 debug!("Account not found for {:?}", coin_balance_name);
-                return Err(SndError::NoSuchAccount);
+                return Err(SndError::NoSuchBalance);
             }
         };
         balance.credit_balance(amount, new_rand::random())
@@ -190,7 +190,7 @@ impl Vault {
             Some(balance) => balance,
             None => {
                 debug!("Coin balance {:?} not found", coin_balance_name);
-                return Err(SndError::NoSuchAccount);
+                return Err(SndError::NoSuchBalance);
             }
         };
         let owner_account = XorName::from(*balance.owner());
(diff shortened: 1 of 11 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -819,6 +801,16 @@ impl Vault {
                     });
                 Response::GetADataRange(res)
             }
+            Request::GetADataValue { address, key } => {
+                let res = self
+                    .get_adata(address, requester_pk, request)
+                    .and_then(move |data| {
+                        data.get(&key)
+                            .map(move |data| data.clone())
```
lionel1704: Can we please change this to `value` or `entry_value` to differentiate it from the entire AData object.

lionel1704: There's also a clippy error here. It's a simple change :)

safe_core/src/client/mock/vault.rs
```diff
@@ -819,6 +801,17 @@ impl Vault {
                     });
                 Response::GetADataRange(res)
             }
+            Request::GetADataValue { address, key } => {
+                let res = self
+                    .get_adata(address, requester_pk, request)
+                    .and_then(move |data| {
+                        data.get(&key)
+                            .cloned()
+                            .map(move |value| value)
```
lionel1704: Now that we are using `cloned()` we don't need to use `map` again.
```suggestion
```

safe_core/src/client/mock/vault.rs
```diff
@@ -934,13 +927,13 @@ impl Vault {
                     .get_adata(append.address, requester_pk, request)
                     .and_then(move |data| match data {
                         AData::PubSeq(mut adata) => {
-                            unwrap!(adata.append(&append.values, index));
+                            unwrap!(adata.append(append.values, index));
```
lionel1704: This can be changed to
```suggestion
                            adata.append(append.values, index)?;
```
There are a few more occurrences. Could you please change them

PR #894, opened 2019-07-11
Add functions for safecoin operations without a client instance

lionel1704:
Users with a wallet and without an account should be able to perform coin operations. The functions added will allow coin operations to be performed, provided the secret key of the wallet is provided.

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -300,8 +300,8 @@ pub trait Client: Clone + 'static {
         .into_box()
     }
 
-    /// Creates a new coin balance on the network.
-    fn create_coin_balance(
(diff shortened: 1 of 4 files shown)


safe_core/src/client/mod.rs
```diff
@@ -1655,6 +1655,90 @@ pub trait Client: Clone + 'static {
     }
 }
 
+/// Get the balance at the given key's location
+pub fn get_balance(wallet_sk: &threshold_crypto::SecretKey) -> Result<Coins, SndError> {
+    let (mut routing, routing_rx) = setup_routing(
```
m-cat: Can you add a `trace` call here for consistency with the other new functions?

Issue #889, opened 2019-07-10
Rename mock vault file


lionel1704:
To avoid the corruption of the mock vault file for developers working with alpha-2, we can use a different name for the mock vault file. Rename the file to `SCL-Mock`

Yoga07:
Resolved in #895

PR #895, opened 2019-07-12
fix/all: fix validation of Owners before PUT for all data types and rename mock vault file

### LINKED ISSUE #885: Owners field must be validated before PUT
The owner(s) field of the immutable / mutable / append-only data should be validated before storing it in the mock vault.

The owner field in the data struct should be validated against the public_key of the PublicId that is held by routing(soon to be replaced by quic-p2p).
For requests by an App, it should be validated against the owner field of the public Id.
This is already extracted from the PublicId [here](https://github.com/maidsafe/safe_client_libs/blob/c999a9a4028271dc885b2600974959b481b03675/safe_core/src/client/mock/vault.rs#L346)

Yoga07:
Resolves #885 and #889 

- Fixes new mutable data safe_app tests
- Adds tests for validation of owners before PUTs
- Renames mock vault file name to 'SCL-Mock'

### DIFF
--- a/safe_app/src/tests/unpublished_mutable_data.rs
+++ b/safe_app/src/tests/unpublished_mutable_data.rs
@@ -36,8 +36,9 @@ fn md_created_by_app_1() {
     let (name_tx, name_rx) = mpsc::channel();
     unwrap!(app.send(move |client, _app_context| {
         let sign_pk = unwrap!(client.public_signing_key());
-        let bls_pk = unwrap!(client.public_bls_key());
-        unwrap!(app_keys_tx.send((sign_pk, bls_pk)));
+        let bls_pk = unwrap!(client.owner_key());
+        let app_bls_key = unwrap!(client.public_bls_key());
+        unwrap!(app_keys_tx.send((sign_pk, PublicKey::from(app_bls_key))));
         let name: XorName = unwrap!(name_rx.recv());
         let entry_actions = MDataSeqEntryActions::new().ins(vec![1, 2, 3, 4], vec![2, 3, 5], 0);
         let cl2 = client.clone();
@@ -57,7 +58,7 @@ fn md_created_by_app_1() {
                     Err(CoreError::NewRoutingClientError(Error::AccessDenied)) => (),
                     Err(x) => panic!("Expected Error::AccessDenied. Got {:?}", x),
                 }
-                let user = PublicKey::from(bls_pk);
+                let user = bls_pk;
                 let permissions = MDataPermissionSet::new().allow(MDataAction::Update);
                 cl3.set_mdata_user_permissions_new(
                     MDataAddress::Seq {
@@ -94,8 +95,6 @@ fn md_created_by_app_1() {
                     .allow(MDataAction::Read),
             );
 
-            let owners = unwrap!(client.public_bls_key());
-
             let name: XorName = XorName(rng.gen());
 
             let mdata = SeqMutableData::new_with_data(
@@ -103,7 +102,7 @@ fn md_created_by_app_1() {
                 DIR_TAG,
                 BTreeMap::new(),
                 permissions,
-                PublicKey::from(owners),
+                unwrap!(client.owner_key()),
             );
             let cl2 = client.clone();
             let cl3 = client.clone();
@@ -138,8 +137,8 @@ fn md_created_by_app_2() {
     let (name_tx, name_rx) = mpsc::channel();
     unwrap!(app.send(move |client, _app_context| {
         let sign_pk = unwrap!(client.public_signing_key());
-        let bls_pk = unwrap!(client.public_bls_key());
-        unwrap!(app_keys_tx.send((sign_pk, bls_pk)));
+        let app_bls_key = unwrap!(client.public_bls_key());
+        unwrap!(app_keys_tx.send((sign_pk, PublicKey::from(app_bls_key))));
         let name: XorName = unwrap!(name_rx.recv());
         let entry_actions = MDataUnseqEntryActions::new().ins(vec![1, 2, 3, 4], vec![2, 3, 5]);
         let cl2 = client.clone();
@@ -169,7 +168,7 @@ fn md_created_by_app_2() {
                     Err(CoreError::NewRoutingClientError(Error::AccessDenied)) => (),
                     Err(x) => panic!("Expected Error::AccessDenied. Got {:?}", x),
                 }
-                let user = PublicKey::from(bls_pk);
+                let user = PublicKey::from(app_bls_key);
                 let permissions = MDataPermissionSet::new()
                     .allow(MDataAction::Insert)
                     .allow(MDataAction::Delete);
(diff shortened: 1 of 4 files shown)


safe_app/src/tests/unpublished_mutable_data.rs
```diff
@@ -230,7 +232,7 @@ fn md_created_by_app_2() {
                 DIR_TAG,
                 data,
                 permissions,
-                PublicKey::from(unwrap!(client.public_bls_key())),
+                PublicKey::from(unwrap!(client.owner_key())),
```
lionel1704: `client.owner_key()` will return a `safe_nd::PublicKey` so this might return a clippy error

safe_core/src/client/mock/vault.rs
```diff
@@ -37,10 +37,7 @@ pub enum Data {
     AppendOnly(AData),
 }
 
-const FILE_NAME: &str = "MockVault";
-lazy_static! {
```
lionel1704: Looks like this block has been removed by mistake :smile_cat:

safe_core/src/client/mod.rs
```diff
@@ -2034,9 +2039,20 @@ mod tests {
                     // Put idata
                     client2.put_pub_idata(data.clone())
                 })
+                .and_then(move |_| {
+                    client3
+                        .put_unpub_idata(test_data.clone())
+                        .then(|res| match res {
+                            Ok(_) => panic!("Verifying owners failed"),
```
lionel1704: Maybe also mention `Unexpected Success` in the panic message

safe_core/src/client/mod.rs
```diff
@@ -2965,13 +3001,33 @@ mod tests {
...
             client
                 .put_adata(AData::UnpubSeq(data.clone()))
                 .map(move |res| {
                     assert_eq!(res, ());
                 })
                 .and_then(move |_| {
                     client1
+                        .put_adata(AData::UnpubSeq(test_data.clone()))
+                        .then(|res| match res {
+                            Ok(_) => panic!("Verifying owner failed"),
```
lionel1704: One more here.

safe_core/src/client/mock/vault.rs
```diff
@@ -390,12 +390,25 @@ impl Vault {
                 Response::GetIData(result)
             }
             Request::PutIData(idata) => {
-                let result = self.put_data(
-                    DataId::Immutable(*idata.address()),
-                    Data::Immutable(idata),
-                    requester,
-                );
-                Response::Mutation(result)
+                if let IData::Unpub(data) = idata.clone() {
```
lionel1704: This can be refactored a bit. Something like:
```
                let result = {
                    if let IData::Unpub(data) = idata.clone() {
                        if owner_pk != *data.owner() {
                            return Err(SndError::InvalidOwners);
                        }
                    }
                    self.put_data(
                        DataId::Immutable(*idata.address()),
                        Data::Immutable(idata),
                        requester,
                    )
                };
                Response::Mutation(result)
```
Reduces duplicate code :smile:

safe_core/src/client/mod.rs
```diff
@@ -2583,15 +2600,33 @@ mod tests {
...
                 .put_seq_mutable_data(data.clone())
                 .and_then(move |res| {
                     assert_eq!(res, ());
                     Ok(())
                 })
+                .and_then(move |_| {
+                    client1
+                        .put_seq_mutable_data(test_data.clone())
+                        .then(|res| match res {
+                            Ok(_) => panic!("Verifying owners failed"),
```
lionel1704: Here too.

safe_core/src/client/mod.rs
```diff
@@ -2034,9 +2039,20 @@ mod tests {
                     // Put idata
                     client2.put_pub_idata(data.clone())
                 })
+                .and_then(move |_| {
+                    client3
+                        .put_unpub_idata(test_data.clone())
+                        .then(|res| match res {
+                            Ok(_) => panic!("Unexpected Success: Validating owners has failed"),
```
m-cat: Shouldn't this say "Validating owners should fail" instead of "has failed"?

safe_core/src/client/mod.rs
```diff
@@ -2583,15 +2606,33 @@ mod tests {
...
                 .put_seq_mutable_data(data.clone())
                 .and_then(move |res| {
                     assert_eq!(res, ());
                     Ok(())
                 })
+                .and_then(move |_| {
+                    client1
+                        .put_seq_mutable_data(test_data.clone())
+                        .then(|res| match res {
+                            Ok(_) => panic!("Unexpected Success: Validating owners has failed"),
```
m-cat: Same here.

safe_core/src/client/mod.rs
```diff
@@ -2965,13 +3035,33 @@ mod tests {
...
             client
                 .put_adata(AData::UnpubSeq(data.clone()))
                 .map(move |res| {
                     assert_eq!(res, ());
                 })
                 .and_then(move |_| {
                     client1
+                        .put_adata(AData::UnpubSeq(test_data.clone()))
+                        .then(|res| match res {
+                            Ok(_) => panic!("Unexpected Success: Validating owners has failed"),
```
m-cat: Same here.

safe_core/src/client/mock/vault.rs
```diff
@@ -586,13 +596,17 @@ impl Vault {
...
-                );
-                Response::Mutation(result)
+                if data.owner() == owner_pk {
+                    let address = *data.address();
+                    let result = self.put_data(
+                        DataId::Mutable(address),
+                        Data::NewMutable(data.clone()),
+                        requester,
+                    );
+                    Response::Mutation(result)
```
m-cat: Instead of returning `Response::Mutation` in two different spots, maybe we should follow the pattern used elsewhere and only return it once. So something like:

```rust
let result = if ... {
    ...
} else {
    ...
};
Response::Mutation(result)
```

safe_core/src/client/mock/vault.rs
```diff
@@ -857,13 +871,18 @@ impl Vault {
...
-                Response::Mutation(result)
+                let owner_idx = adata.owners_index();
+                if unwrap!(adata.get_owners(owner_idx - 1)).public_key == owner_pk {
+                    let address = *adata.address();
+                    let result = self.put_data(
+                        DataId::AppendOnly(address),
+                        Data::AppendOnly(adata),
+                        requester,
+                    );
+                    Response::Mutation(result)
```
m-cat: Same comment as above.

safe_core/src/client/mock/vault.rs
```diff
@@ -857,12 +871,17 @@ impl Vault {
...
             //
             Request::PutAData(adata) => {
+                let owner_idx = adata.owners_index();
                 let address = *adata.address();
-                let result = self.put_data(
-                    DataId::AppendOnly(address),
-                    Data::AppendOnly(adata),
-                    requester,
-                );
+                let result = if unwrap!(adata.get_owners(owner_idx - 1)).public_key != owner_pk {
```
lionel1704: Can you remove this `unwrap!` please?

Issue #900, opened 2019-07-12
Unregistered clients should be able to read published data


lionel1704:
For registered clients / apps, the requests are signed by their respective keys and verified at the source elders.

For unregistered clients, since there are no keys a signature cannot be created and an `OperationForbidden` error is thrown. This should not be the case.

*Solution:* We can solve this by creating a random keypair using which unregistered clients can sign requests. This key is destroyed with the session.

lionel1704:
Resolved by #903

PR #903, opened 2019-07-12
Allow GET requests to be sent by unregistered clients

### LINKED ISSUE #899: Unable to `get_adata_indices` with aunauthed connection (Operation Forbidden)
Using https://github.com/joshuef/safe-cli/tree/Fix-Integration 

Attempting to fetch latest published append only data, we're seeing `Operation Forbidden` when calling `get_adata_indices` via an aunauthed connection.

To replicate. Pull the above branch and run

`RUST_LOG=safe=trace cargo test --features="mock-network" -- --test-threads=1 calling_safe_files_put_recursive_and_set_dest_path`

## Currently:

`[Error] NetDataError - Failed to get Sequenced Appendable Data indices: CoreError(Operation forbidden - CoreError::OperationForbidden)`

## Wanted: 

All published data should be accessible to unauthed connections.

lionel1704:
- Create a random keypair for every unregistered session and init routing using a ClientFullId generated from those keys
- Add tests for unregistered access to published AData

Resolves #899, #900

### DIFF
--- a/safe_app/src/tests/mod.rs
+++ b/safe_app/src/tests/mod.rs
@@ -31,8 +31,8 @@ use safe_core::utils::test_utils::random_client;
 use safe_core::MockRouting;
 use safe_core::{Client, CoreError};
 use safe_nd::{
-    AData, ADataAddress, Error as SndError, PubImmutableData, PubUnseqAppendOnlyData,
-    UnpubUnseqAppendOnlyData, XorName,
+    ADataAddress, ADataOwner, AppendOnlyData, Error as SndError, PubImmutableData,
+    PubUnseqAppendOnlyData, UnpubUnseqAppendOnlyData, XorName,
 };
 use std::collections::HashMap;
 use std::rc::Rc;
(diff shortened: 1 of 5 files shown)


safe_app/src/tests/mod.rs
```diff
@@ -272,3 +279,59 @@ fn app_container_creation() {
...
+
+// Test unregistered clients.
+// 1. Have a registered clients put published immutable and published append-only data on the network.
+// 2. Try to read them as unregistered.
+#[test]
+fn unregistered_client() {
+    let addr: XorName = new_rand::random();
+    let tag = 15002;
+    let pub_idata = PubImmutableData::new(unwrap!(utils::generate_random_vector(30)));
+    let pub_adata = AData::PubUnseq(PubUnseqAppendOnlyData::new(addr, tag));
```
m-cat: Might be cleaner to use `.into()` here instead of explicitly specifying the `AData::PubUnseq` etc variants

Issue #897, opened 2019-07-12
Apps should not be allowed to send DeleteMData


lionel1704:
Similar to #896, apps should not be able to send requests to delete a mutable data.
This can be moved into `AuthClient`

Yoga07:
Resolved in #904

PR #904, opened 2019-07-12
fix/all: disable AppClients from accessing Auth Keys

### LINKED ISSUE #896: Apps should not be able to send requests to get / mutate auth keys
Currently, 
- `list_auth_keys`
- `delete_auth_key`
- `insert_auth_key`
are all functions of the `Client` trait. Apps should not be allowed to perform such operations. So this can be removed from the `Client` trait and into the `AuthClient` and `CoreClient` for testing.

Yoga07:
Resolves #896 and #897 

- Introduces a new trait implementing functions which are not supposed to be called by AppClients
- Adapts safe_authenticator to the new impl
- Fixes tests for auth keys

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -593,13 +593,6 @@ pub trait Client: Clone + 'static {
             .into_box()
     }
 
-    /// Delete MData from network
-    fn delete_mdata(&self, address: MDataAddress) -> Box<CoreFuture<()>> {
-        trace!("Delete entire Mutable Data at {:?}", address);
-
-        send_mutation_new(self, Request::DeleteMData(address))
-    }
-
     /// Mutates `MutableData` entries in bulk.
     fn mutate_mdata_entries(
         &self,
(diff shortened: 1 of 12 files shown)


safe_core/src/client/mod.rs
```diff
@@ -1735,6 +1687,66 @@ pub fn wallet_transfer_coins(
...
 }
 
+/// Trait for manipulating Auth Keys
+pub trait AuthKeys: Clone + 'static {
+    /// Fetches a list of authorised keys and version in MaidManager.
+    fn list_auth_keys_and_version(
+        &self,
+    ) -> Box<CoreFuture<(BTreeMap<PublicKey, AppPermissions>, u64)>>
+    where
+        Self: Client,
```
m-cat: I'm confused, is this saying that `Self` must implement the `Client` trait? Why not have `AuthKeys` depend on `Client` (i.e. `AuthKeys: Client + Clone + 'static`)? Also, it is not immediately clear why this is separate from `Client` in the first place (unless one reads the associated issue). Can you add some documentation in this module addressing these design decisions?

Issue #783, opened 2019-04-04
The NFS `file_open` doesn't support creating a file without passing empty File struct


m-cat:
In order to create a new NFS file, the `file_open` function doesn't support receiving a `null` as the File struct param, it requires the client to crate and empty File struct for such purpose. It's herein proposed to support receiving a `null` param so the client doesn't need to bother with creating an empty File struct.

sprucely:
Thinking of taking this on. If the ffi File is null, that means there is no length or capacity information for user_metadata. So when initializing a native File, does that mean its user_metadata should be initialized to an empty Vec<u8> with 0 capacity? Should the into_repr_c() in native File also be modified to set user_metadata_ptr to null when it encounters an empty Vec<u8> with 0 capacity? I could see a client developer easily making the mistake of only checking if user_metadata_ptr is not null while ignoring len and cap, and then trying to dereference it.

m-cat:
> Thinking of taking this on.

Awesome! Thank you for the assistance, and great questions! 

> If the ffi File is null, that means there is no length or capacity information for user_metadata. So when initializing a native File, does that mean its user_metadata should be initialized to an empty Vec with 0 capacity?

I would say so; simply `Vec::new()` would suffice. This field is information provided by the user, and if he wishes to fill it out, he can always create his own file. So in the case where he does not provide a file, just initialize it to an empty vec. I would thoroughly document these behaviors and assumptions in the doc-string for this function.

> Should the into_repr_c() in native File also be modified to set user_metadata_ptr to null when it encounters an empty Vec with 0 capacity? I could see a client developer easily making the mistake of only checking if user_metadata_ptr is not null while ignoring len and cap, and then trying to dereference it.

I don't see any need to modify this function. There is nothing (in this case) wrong with dereferencing a non-null pointer - a properly-written client-side function will just see an empty vec of length 0.

sprucely:
OK. I'm finding some good info in the wiki, but I want to make sure I'm clear on the difference between File::user_metadata and the user_data param in fn file_open(). My guess is that user_metadata gets stored with the file bytes, while user_data is simply passed to the callback as an alternative to closures for languages that don't support closures?

m-cat:
[...]

PR #906, opened 2019-07-13
chore/ffi: allow file_open to accept null file arg #783

sprucely:
First contribution. I think I crossed all t's and dotted all i's.

### DIFF
--- a/safe_app/src/ffi/nfs.rs
+++ b/safe_app/src/ffi/nfs.rs
@@ -159,6 +159,12 @@ pub unsafe extern "C" fn dir_delete_file(
 }
 
 /// Open the file to read or write its contents.
+///
+/// # Notes
+///
+/// When creating an empty file with no metadata, a null pointer can be passed
+/// in for the `file` argument. In such cases, the library will internally
+/// create a default File with a 0 length `user_metadata_ptr`
 #[no_mangle]
 pub unsafe extern "C" fn file_open(
     app: *const App,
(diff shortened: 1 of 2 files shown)


sprucely:
Sorry for breaking your mock_windows build o0

jacderida:
@sprucely 

Not your fault - looks like the Windows slave in our Jenkins environment ran out of disk space.

I tried to run the build again for you but I'm getting a weird error on the Linux builds.

safe_app/src/ffi/nfs.rs
```diff
@@ -159,6 +159,12 @@ pub unsafe extern "C" fn dir_delete_file(
 }
 
 /// Open the file to read or write its contents.
+///
+/// # Notes
+///
+/// When creating an empty file with no metadata, a null pointer can be passed
+/// in for the `file` argument. In such cases, the library will internally
+/// create a default File with a 0 length `user_metadata_ptr`
```
m-cat: > with a 0 length `user_metadata_ptr`

This isn't quite accurate -- the pointer itself is not 0-length, as it will be a valid pointer. Better would be to just say "a default File with no user metadata" which does not carry any assumptions about details such as variable names, which are subject to change in future refactors.

sprucely: Okay, I added a commit with the correction. For future reference, do you prefer new commits for one-liners like this, or commit amendments? I guess it doesn't matter if you squash merge your pull requests.

m-cat: We usually prefer amending the last commit and force-pushing as it keeps the commit history cleaner and GitHub now shows diffs for force-pushes. Substantial changes can happen in new commits, but addressing review comments usually results in only minor changes :)

PR #917, opened 2019-07-18
refactor/all: refactor mock-vault based on lastest safe-nd changes

Yoga07:


safe_core/src/client/mock/vault.rs
```diff
@@ -339,6 +340,114 @@ impl Vault {
         }
     }
 
+    fn check_pub_perms_adata(
```
lionel1704: This function can be clubbed with `check_unpub_perms`

safe_core/src/client/mock/vault.rs
```diff
@@ -339,6 +340,114 @@ impl Vault {
...
+                        Err(SndError::AccessDenied)
+                    }
+                },
+            ),
+            _ => Err(SndError::InvalidOperation),
+        }
+    }
+
+    fn check_unpub_perms_adata(
+        &self,
```
lionel1704: Do you need the `self` parameter?

safe_core/src/client/mock/vault.rs
```diff
@@ -339,6 +340,114 @@ impl Vault {
...
+                        }
+                    },
+                )
+            }
+            _ => Err(SndError::InvalidOperation),
+        }
+    }
+
+    fn check_perms_mdata(
+        &self,
```
lionel1704: Here too. Is this needed?

safe_core/src/client/mock/vault.rs
```diff
@@ -1221,9 +1314,7 @@ impl Vault {
             message_id,
         })
     }
-    //
```
lionel1704: Also was this removed on purpose?

safe_core/src/client/mock/vault.rs
```diff
@@ -954,60 +1063,54 @@ impl Vault {
...
-                        };
-                        match address {
-                            ADataAddress::PubSeq { .. } => {
-                                let res = match data {
-                                    AData::PubSeq(adata) => {
-                                        match adata.fetch_permissions_at_index(idx as u64) {
-                                            Some(perm) => Ok(perm.clone()),
-                                            None => Err(SndError::NoSuchEntry),
-                                        }
+                    .and_then(move |data| match address {
```
m-cat: It's not a part of your PR, but I don't understand the `Macro cannot be used here` comment a few lines up, above `address,`. This comment seems meaningless to me, can it be removed?

safe_core/src/client/mock/vault.rs
Yoga07: This did not make much sense as there was only one function under AData, and the rest of the types did not have these kinda banners too :)

safe_app/src/tests/mod.rs
```diff
@@ -326,15 +326,16 @@ fn unregistered_client() {
...
+                        assert_eq!(data.tag(), pub_adata.tag());
                     })
             })
             .then(move |_| {
                 client3
                     .get_adata(ADataAddress::UnpubUnseq { name: addr, tag })
                     .then(|res| {
                         match res {
-                            Err(CoreError::NewRoutingClientError(SndError::AccessDenied)) => (),
+                            Err(CoreError::NewRoutingClientError(SndError::NoSuchData)) => (),
```
lionel1704: This error type should not be changed.
If this error type is returned from safe_nd we can use `map_err` to return the error we need.

Issue #918, opened 2019-07-18
Replace no. of PUTs with coin balance


lionel1704:
Currently, the practise of using PUTs for the account's balance is being followed. This needs to be replaced with coin balance deductions for every mutation.

lionel1704:
Resolved in #927

Issue #914, opened 2019-07-17
Refactor safe_auth internals to use the new login packet mechanism


lionel1704:
In the safe_authenticator mutable data is used to store the account packet (now called login packet). Once create account and login is refactored to use the new mechanism for login packets, API such as update_account_packet should also be updated to use the `UpdateAccessPacket` request / response variants.

lionel1704:
Resolved in #927

PR #921, opened 2019-07-22
Adapt internals of safe_authenticator to use the new LoginPacket approach

### LINKED ISSUE #859: Change account packet handling in the SAFE Authenticator
Access container and other structures related to a user's account should now be stored using the new functions and RPCs defined in #858.

lionel1704:
Resolves #859, #914, #918


safe_authenticator/src/errors.rs
```diff
@@ -211,6 +211,12 @@ impl From<IoError> for AuthError {
     }
 }
 
+impl From<safe_nd::Error> for AuthError {
+    fn from(error: safe_nd::Error) -> AuthError {
+        AuthError::from(CoreError::NewRoutingClientError(error))
```
m-cat: I think `AuthError` should have its own variant for `safe_nd::Error`, wrapping it in `CoreError` may mislead someone into believing an error originated in safe_core.

safe_core/src/client/mock/vault.rs
```diff
@@ -491,35 +502,49 @@ impl Vault {
...
+                        let name = requester.name();
+                        if self.get_account(&name).is_none() {
+                            self.insert_account(*name);
+                        }
+                        let account = unwrap!(self.get_account(&name));
+                        Ok((account.auth_keys().clone(), account.version()))
+                    }
                 };
-                Response::ListAuthKeysAndVersion(res)
+                Response::ListAuthKeysAndVersion(result)
```
m-cat: Agreed with renaming `res` to `result` 👍

safe_app/examples/client_stress_test.rs
```diff
@@ -162,7 +162,7 @@ fn main() {
     };
     let mut rng = rand::XorShiftRng::from_seed([0, 0, 0, seed]);
 
-    let invitation: String = if let Some(i) = matches.value_of("invite") {
+    let _invitation: String = if let Some(i) = matches.value_of("invite") {
```
nbaksalyar: Could we just remove this block entirely? And also remove the `invite` parameter in the command line arguments.

safe_app/examples/self_authentication.rs
```diff
@@ -99,21 +99,19 @@ fn main() {
         let _ = std::io::stdin().read_line(&mut secret_1);
         secret_1 = secret_1.trim().to_string();
 
-        let invitation: String = if let Some(i) = matches.value_of("invite") {
+        let _invitation: String = if let Some(i) = matches.value_of("invite") {
```
nbaksalyar: Could we just remove this block entirely? And also remove the `invite` parameter in the command line arguments.

safe_app/examples/client_stress_test.rs
```diff
@@ -197,10 +197,14 @@ fn main() {
         println!("\t================");
         println!("\nTrying to create an account ...");
 
+        // FIXME
```
nbaksalyar: Could you please expand what is required to be fixed? I.e. provide an expanded comment like `FIXME - remove this once we have proper accounts handling` or something like this.

safe_app/examples/self_authentication.rs
```diff
@@ -99,21 +99,19 @@ fn main() {
...
             i.to_string()
         } else {
             unwrap!(utils::generate_random_string(10))
         };
 
+        // FIXME
```
nbaksalyar: Could you please expand what is required to be fixed? I.e. provide an expanded comment like `FIXME - remove this once we have proper accounts handling` or something like this.

safe_authenticator/src/client.rs
```diff
@@ -154,52 +154,54 @@ where {
...
+
+        let sig = client_full_id.sign(&acc_ciphertext);
+        let client_pk = client_full_id.public_id().public_key();
+        let new_account_data = LoginPacket::new(acc_loc, *client_pk, acc_ciphertext, sig)?;
+        let balance_client_id = ClientFullId::with_bls_key(balance_sk);
+
+        {
+            let (mut routing, routing_rx) = setup_routing(
+                full_id.clone(),
+                PublicId::Client(balance_client_id.public_id().clone()),
```
nbaksalyar: Just a suggestion for the future: it's probably worth having something like
```rust
impl From<client::FullId> for PublicId {
  fn from(full_id: &client::FullId) -> Self {
    PublicId::Client(full_id.public_id().clone())
  }
}
```
in safe-nd to make it cleaner, so that this line can be replaced with `PublicId::from(balance_client_id)`.

safe_authenticator/src/client.rs
```diff
@@ -154,52 +154,54 @@ where {
...
+                None,
+            )?;
+
+            let rpc_response = routing.req_as_client(
+                &routing_rx,
+                Request::CreateLoginPacketFor {
+                    new_owner: pub_key,
+                    amount: unwrap!(Coins::from_str("1")),
```
nbaksalyar: Although it's temp code, I think we shouldn't use unwrap here:
```suggestion
                    amount: Coins::from_str("1")?
```
Even better, take the `amount` as an argument for this function.

safe_authenticator/src/client.rs
```diff
@@ -154,52 +154,54 @@ where {
...
+                    transaction_id: new_rand::random(),
+                    new_login_packet: new_account_data,
```
nbaksalyar: I'd suggest renaming `new_account_data` to `new_login_packet`, it's more clearly conveys the intent.
Then this line can be shortened to
```suggestion
                    new_login_packet,
```

safe_authenticator/src/client.rs
```diff
@@ -154,52 +154,54 @@ where {
...
+        {
+            let (mut routing, routing_rx) = setup_routing(
+                full_id.clone(),
+                PublicId::Client(balance_client_id.public_id().clone()),
+                None,
+            )?;
+
+            let rpc_response = routing.req_as_client(
+                &routing_rx,
+                Request::CreateLoginPacketFor {
```
nbaksalyar: Why `CreateLoginPacketFor` in this case? I would think it should rather be `CreateLoginPacket`, and we should have a separate function in the `Authenticator` (and probably even in the apps API - something to consider and discuss) to create a new account for someone else.

safe_authenticator/src/client.rs
```diff
@@ -296,41 +298,44 @@ where {
...
-        let user_cred = UserCred::new(password, pin);
 
-        let dst = Authority::NaeManager(acc_loc);
+        let mut seeder: Vec<u8> = Vec::with_capacity(acc_locator.len() + acc_password.len());
+        seeder.extend_from_slice(acc_locator);
+        seeder.extend_from_slice(acc_password);
+
+        let seed = sha3_256(&seeder);
+        let mut rng = StdRng::from_seed(seed);
+        let client_full_id = ClientFullId::new_bls(&mut rng);
```
nbaksalyar: I'm thinking that we should have an internal (module-level) function like `fn create_client_id(acc_locator: &[u8], acc_password: &[u8]) -> ClientFullId` with this code to have fewer duplicates.

safe_authenticator/src/client.rs
```diff
@@ -421,47 +426,45 @@ where {
...
-        ])
+
+        let sig = full_id.sign(&encrypted_account);
+        let client_pk = full_id.public_id().public_key();
+        LoginPacket::new(acc_loc, *client_pk, encrypted_account, sig).map_err(AuthError::from)
     }
 
     /// Updates user's account packet.
-    pub fn update_account_packet(&self) -> Box<AuthFuture<()>> {
+    pub fn update_account_packet(&self) -> Result<(), AuthError> {
```
nbaksalyar: This should remain `Box<AuthFuture<()>>`. It's OK to use `req_as_client` (a blocking function) in `register` and `login` because we can't do much without having a client with the full ID anyway, but I think it's better to return async result in this case. Otherwise this function would block the entire event loop.

safe_authenticator/src/client.rs
```diff
@@ -421,47 +426,45 @@ where {
...
+            Self::prepare_account_packet_update(*acc_loc, account, keys, &client_full_id)?;
 
-        let data_name = auth_inner.acc_loc;
+        let (mut routing, routing_rx) = setup_routing(
+            None,
+            PublicId::Client(client_full_id.public_id().clone()),
+            None,
+        )?;
 
-        self.mutate_mdata_entries(data_name, TYPE_TAG_SESSION_PACKET, update)
```
nbaksalyar: To retain the asynchronicity here, can't we just have a function similar to `mutate_mdata_entries`, but internal to the Auth client? Let's call it `update_login_packet`, and then it can use the same helper functions like `safe_core::send_mutation` etc.

safe_authenticator/src/client.rs
lionel1704: Great idea. I'll get this into a new PR which will make required changes to adapt to the latest changes in safe-nd.

nbaksalyar:
Closed in favour of #927 (rebased)

Issue #785, opened 2019-04-04
Add logic for releasing FileContextHandle without closing the file


m-cat:
This was reported by frontend:

> It's not possible to free/release a `FileContext` object without closing it so the client is forced to either store the file that was created or leak the `FileContext` object.
>This is to cover scenarios where the client has some logic to cancel/discard a file that was opened but for some reason/logic it's not desired to be stored anymore.

In other words: a `FileContextHandle` can only be free'd by calling `file_close`, but the files operations can be canceled, in which case `file_close` won't be called (as it would save the file to the network). This would result in leaking `FileContextHandle` as it is never freed.

sprucely:
I'm thinking of making a fn for this with a signature and body pretty similar to file_close...
```Rust
#[no_mangle]
pub unsafe extern "C" fn file_cancel(
    app: *const App,
    file_h: FileContextHandle,
    user_data: *mut c_void,
    o_cb: extern "C" fn(user_data: *mut c_void, result: *const FfiResult, file: *const File),
)  {
    catch_unwind_cb(user_data, o_cb, || {
        let user_data = OpaqueCtx(user_data);

        (*app).send(move |_client, context| {
            let file_ctx = try_cb!(context.object_cache().remove_file(file_h), user_data, o_cb);

            o_cb(
                user_data.0,
                FFI_RESULT_OK,
                &file_ctx.original_file.into_repr_c(),
            );
            None
        })
    })
}
```
I thought of removing the file parameter from the callback, but I think it could be helpful to keep it in case the client wants to do something further with the file after cancelling an operation. I haven't yet looked at the api for `file_ctx.writer`, so I'm not sure yet if anything needs to be done to clean it up without actually calling close() on it.

m-cat:
Hey @sprucely, thanks for taking this! Looks great, and I agree that keeping the function similar to `file_close` makes sense. So keeping the file parameter in the callback is something I agree with. 

> I haven't yet looked at the api for file_ctx.writer, so I'm not sure yet if anything needs to be done to clean it up without actually calling close() on it.

Nope, dropping the file context is enough to clean it and its contents up without causing a leak.

One note: in the documentation for this function, I would emphasize that this function should be limited as much as possible, as opening a file incurs a network GET, so cancelling the file operations would be wasteful. It should only be used in rare scenarios.

m-cat:
Closed in https://github.com/maidsafe/safe_client_libs/pull/925.

PR #925, opened 2019-07-24
enhancement/ffi: add fn file_cancel to free FileContextHandle without…

sprucely:
… closing #785

Not sure whether the test should be more thorough. I mean, it's pretty clear that a new file isn't saved to the directory. But should it also test whether changes to an existing file are not saved?

### DIFF
--- a/safe_app/src/ffi/nfs.rs
+++ b/safe_app/src/ffi/nfs.rs
@@ -370,3 +370,32 @@ pub unsafe extern "C" fn file_close(
         })
     })
 }
+
+/// Prevents saving to the network any changes performed on the given file
+/// context handle and frees the handle and its associated file contents.
+///
+/// This should be used sparingly, as it wastes the network GET used to
+/// obtain the file handle.
+#[no_mangle]
+pub unsafe extern "C" fn file_cancel(
+    app: *const App,
+    file_h: FileContextHandle,
+    user_data: *mut c_void,
+    o_cb: extern "C" fn(user_data: *mut c_void, result: *const FfiResult, file: *const File),
+) {
+    catch_unwind_cb(user_data, o_cb, || {
+        let user_data = OpaqueCtx(user_data);
+
+        (*app).send(move |_client, context| {
+            let file_ctx = try_cb!(context.object_cache().remove_file(file_h), user_data, o_cb);
+
+            // The reader and/or writer will be dropped automatically.
+            o_cb(
+                user_data.0,
+                FFI_RESULT_OK,
+                &file_ctx.original_file.into_repr_c(),
+            );
+            None
+        })
+    })
+}
--- a/safe_app/src/ffi/tests/nfs.rs
+++ b/safe_app/src/ffi/tests/nfs.rs
@@ -1172,3 +1172,89 @@ fn write_chunks(
         unwrap!(call_1(|ud, cb| file_close(app, write_h, ud, cb)))
     }
 }
+
+#[test]
+fn cancel_file() {
+    let (app, container_info) = setup();
+
+    // Create empty file and save to directory
+    let user_metadata = b"metadata".to_vec();
+    let file = NativeFile::new(user_metadata.clone());
+    let ffi_file = file.into_repr_c();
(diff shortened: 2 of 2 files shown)


safe_app/src/ffi/nfs.rs
```diff
@@ -370,3 +370,32 @@ pub unsafe extern "C" fn file_close(
         })
     })
 }
+
+/// Prevents saving to the network any changes performed on the given file
+/// context handle and frees the handle.
+///
+/// This should be used sparingly, as it wastes the network get used to
```
m-cat: Can we write "get" as `GET`, and fix the `..` at the end of this sentence?

safe_app/src/ffi/nfs.rs
```diff
@@ -370,3 +370,32 @@ pub unsafe extern "C" fn file_close(
         })
     })
 }
+
+/// Prevents saving to the network any changes performed on the given file
+/// context handle and frees the handle.
```
m-cat: Please add, "frees the handle and its associated file contents".

safe_app/src/ffi/tests/nfs.rs
```diff
@@ -1172,3 +1172,56 @@ fn write_chunks(
...
+            ud,
+            cb
+        )));
+        unwrap!(call_1(|ud, cb| file_cancel(&app, write_h, ud, cb)))
+    };
+
+    assert_eq!(cancelled_file, original_file);
+
+    // Fetching non-existing file should fail
+    let res: Result<(NativeFile, u64), i32> = unsafe {
```
m-cat: This will fail regardless of whether `file_cancel` works or not, because we never insert the file into the container. It doesn't matter that we called `file_cancel`, the returned file is still valid and can be inserted into the container. So this test shouldn't be about whether the file exists in the container, it should test whether the file handle and its contents have been freed.

Instead, we can test that trying to use the freed file handle again results in an error.

sprucely: Okay. I think I misunderstood the purpose of the dir functions. Let's see if I have this straight. The File object is strictly for accessing/modifying the datamap for a given file? And the dir functions serve to store, delete or replace named references to a file? Are these references like symbolic links, allowing multiple references to a single file?

m-cat: Yes, exactly. `File`s refer to some data on the network, and directory entries refer to `File`s. So when you want to read a file, you actually have to make two network requests: one for fetching the `File`, and one for fetching its data from its datamap (though it depends on the size of the file, so this picture is a bit simplified). [This page](https://github.com/maidsafe/safe_client_libs/wiki/GET-and-PUT-reference#nfs) might be helpful as well. Hope that is all clear!

sprucely: Sorry it took awhile to get back to this. I updated the doc comment and added to the test an attempt to use the freed file handle. Also, it's now inserting the empty file into a directory before writing to it and canceling, and then fetching it back from the directory to check if it has been modified.

Another thing I'm curious about. After inserting a zero-length file, if I attempt to open it with OPEN_MODE_APPEND, it returns a ERR_NO_SUCH_DATA. Is that because the XOR addressing requires at least some content in order to calculate the address of the first chunk? That would make more sense for attempting to read a nonexistent file. But for append I would expect it to open (perhaps with some internal indicator that there is no data) and, when closing, either create a new starting chunk for a handle to a zero-length file or link to a previous chunk for a file handle to existing content.

m-cat: `APPEND` always assumes there is existing data to append to and will try to retrieve it. Can you please write an issue about this? When one of us has time we'll investigate whether we can do something smarter in the case of empty files.

> link to a previous chunk for a file handle to existing content.

If I'm understanding you correctly, then what you're saying is impossible. Self-encryption requires knowing every chunk in order to correctly encrypt, which is why `APPEND` needs to get the existing chunks first. You can get more info about the algorithm we use [here](https://github.com/maidsafe/self_encryption#overview).

PR #926, opened 2019-07-24
fix/mock: refactor vault::process_request

### LINKED ISSUE #910: Errors should not be returned from vault::process_request
In the `vault::process_request` function there are a few `return` statements that return errors. Error should not be returned in this fashion. They should be returned as a response message corresponding to the incoming request message.

However, in certain cases such as applications sending requests for unsupported operations such as operations related to `AuthKeys`, `DeleteMData` and `DeleteIData`, vaults will identify these requests as malicious and drop them without a response. In such cases, we can use an early return to return a `UnsupportedOperation` error to `process_request`.
In `routing::send` this result should not be `unwrap!`ed, it should be handled greacefully using `map_err(...)?` using an early return to exit without sending a response.

m-cat:
+ Do not return early errors from `process_request`, encapsulate them in the
proper response variant instead
+ Do not unwrap result of `process_request` in `send`
+ Add missing checks for mismatch between requested data address and received
data kind
+ Do some refactoring and simplication of `process_request`

Closes #910

### DIFF
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -189,8 +189,9 @@ impl Routing {
                 None => self.public_id.clone(),
             };
             let mut vault = self.lock_vault(true);
-            unwrap!(vault.process_request(public_id, payload.to_vec()))
-        };
+            vault.process_request(public_id, payload.to_vec())
+        }?;
+
         // Send response back to a client
         let (message_id, response) = if let Message::Response {
             message_id,
@@ -205,6 +206,7 @@ impl Routing {
             res: Ok(unwrap!(serialise(&response))),
             msg_id: message_id,
         };
+
         // Use dummy authority for now
         let dummy_authority = Authority::ClientManager(new_rand::random());
         self.send_response(DEFAULT_DELAY_MS, dummy_authority, dummy_authority, response);
--- a/safe_core/src/client/mock/routing.rs
+++ b/safe_core/src/client/mock/routing.rs
@@ -189,7 +189,7 @@ impl Routing {
                 None => self.public_id.clone(),
             };
             let mut vault = self.lock_vault(true);
-            vault.process_request(public_id, payload.to_vec())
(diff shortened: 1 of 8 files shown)


nbaksalyar:
Should close #910.

@m-cat could you please rebase this PR?

m-cat:
Depends on https://github.com/maidsafe/safe-nd/pull/78, https://github.com/maidsafe/safe-nd/pull/79, https://github.com/maidsafe/safe-nd/pull/80, and https://github.com/maidsafe/safe-nd/pull/83.

safe_core/src/client/mock/vault.rs
```diff
@@ -451,6 +452,7 @@ impl Vault {
         requester: PublicId,
         payload: Vec<u8>,
```
nbaksalyar: While we're at it, could you please change the `process_request` signature to have the `&[u8]` type instead of `Vec<u8>`?

safe_core/src/client/mock/routing.rs
```diff
@@ -189,8 +189,9 @@ impl Routing {
                 None => self.public_id.clone(),
             };
             let mut vault = self.lock_vault(true);
-            unwrap!(vault.process_request(public_id, payload.to_vec()))
-        };
+            vault.process_request(public_id, payload.to_vec())
```
nbaksalyar: After the change in `process_request`, it will become unnecessary to convert the payload into a vector.

safe_core/src/client/mock/vault.rs
```diff
@@ -947,46 +925,54 @@ impl Vault {
...
                             self.insert_data(data_name, Data::NewMutable(data));
                             self.commit_mutation(requester.name());
+
                             Ok(())
                         });
                 Response::Mutation(result)
             }
             //
-            // ===== Immutable Data =====
+            // ===== AppendOnly Data =====
```
nbaksalyar: Nice catch :smile:

PR #939, opened 2019-07-29
Add test for create_login_packet_for flow

### LINKED ISSUE #928: Implement `CreateLoginPacketFor` function in Client
Implement a new function to create a new account for someone else, using your own balance. This function should be usable from both App and Authenticator contexts, so it should be added at the `Client` trait level in safe_core and have the following signature:
```rust
use safe_nd::{PublicKey, LoginPacket, Coins, Transaction};

pub fn create_login_packet_for(
  &self,
  new_balance_owner: PublicKey,
  login_packet: LoginPacket,
  amount: Coins
) -> Box<CoreFuture<Transaction>>;
```

The [reference for the flow is available here](https://hackmd.io/@lionel1704/account-creation).

lionel1704:
Closes #928

### DIFF
--- a/safe_authenticator/src/client.rs
+++ b/safe_authenticator/src/client.rs
@@ -903,4 +903,62 @@ mod tests {
                 })
         })
     }
+
+    // Create a login packet using some credentials and pass the login packet to a client
+    // who stores it on the network and creates a wallet for it.
+    // Now calling login using the same credentials should succeed and we must be able to fetch the balance.
+    #[test]
+    fn create_login_packet_for_test() {
+        let sec_0 = unwrap!(utils::generate_random_string(10));
+        let sec_1 = unwrap!(utils::generate_random_string(10));
+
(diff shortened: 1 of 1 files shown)


safe_authenticator/src/client.rs
```diff
@@ -903,4 +903,61 @@ mod tests {
...
+        let acc_ciphertext = unwrap!(acc.encrypt(&password, &pin));
+
+        let client_full_id = create_client_id(acc_locator, acc_password);
+
+        let sig = client_full_id.sign(&acc_ciphertext);
+        let client_pk = client_full_id.public_id().public_key().clone();
+        let new_login_packet = unwrap!(LoginPacket::new(acc_loc, client_pk, acc_ciphertext, sig));
+        let five_coins = unwrap!(Coins::from_str("5"));
+
+        random_client(move |client| {
```
nbaksalyar: Probably worth adding a comment saying that it's implied `random_client` creates a client with some pre-loaded balance?

PR #940, opened 2019-07-29
Add test to ensure that anyone can read published data

### LINKED ISSUE #919: Add tests to assert that any client/app can GET (Seq/Unseq)Published AppendOnly Data
Add tests to assert that Anyone(Client/App) in the SAFE Network can access and read a (Seq/Unseq)Published AppendOnly Data.

1. Upload(PUT) a Published AppendOnly Data into the network with an empty Permissions list.
2. Access(GET) the uploaded data from the network with a different Client(who is not present in the Permissions list)
3. Assert if both the uploaded data and the accessed data are the same.

lionel1704:
Resolves #919

### DIFF
--- a/safe_app/src/tests/mod.rs
+++ b/safe_app/src/tests/mod.rs
@@ -12,8 +12,8 @@ mod mutable_data;
 mod unpublished_mutable_data;
 
 use crate::ffi::test_utils::test_create_app_with_access;
+use crate::test_utils::{create_app, create_random_auth_req, gen_app_exchange_info};
 use crate::test_utils::{create_app_by_req, create_auth_req, create_auth_req_with_access};
-use crate::test_utils::{create_random_auth_req, gen_app_exchange_info};
 use crate::{run, App, AppError};
 use ffi_utils::test_utils::call_1;
 use futures::Future;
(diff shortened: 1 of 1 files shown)


safe_app/src/tests/mod.rs
```diff
@@ -340,6 +341,102 @@ fn unregistered_client() {
...
+                .put_pub_idata(pub_idata)
+                .and_then(move |_| client2.put_adata(pub_seq_adata.into()))
+                .and_then(move |_| client3.put_adata(pub_unseq_adata.into()))
+        });
+    }
+
+    // Unregistered apps should be able to read the data
+    {
+        let pub_idata = pub_idata.clone();
+        let pub_seq_adata = pub_seq_adata.clone();
```
m-cat: We don't need to clone the entire adata's here right? We only need the addresses :)

safe_app/src/tests/mod.rs
```diff
@@ -340,6 +341,100 @@ fn unregistered_client() {
...
+        client
+            .get_pub_idata(*pub_idata.name())
+            .map_err(AppError::from)
+            .and_then(move |data| {
+                assert_eq!(data, pub_idata);
+                client2
+                    .get_adata(ADataAddress::PubUnseq { name: addr, tag })
+                    .map_err(AppError::from)
+                    .map(move |data| {
+                        assert_eq!(data.address(), pub_unseq_adata.address());
```
m-cat: We're calling `pub_unseq_adata.address()` again here and below when it was already retrieved previously

safe_app/src/tests/mod.rs
```diff
@@ -340,6 +341,101 @@ fn unregistered_client() {
...
+        });
+    }
+
+    let pub_seq_adata_addr = ADataAddress::PubSeq { name, tag };
+    let pub_unseq_adata_addr = ADataAddress::UnpubSeq { name, tag };
+
+    // Unregistered apps should be able to read the data
+    {
+        let pub_idata = pub_idata.clone();
+        let pub_seq_adata_addr = pub_seq_adata_addr.clone();
```
m-cat: One more comment 😄 

This line and the one below it can be deleted :) Also, addresses implement `Copy` so no need to clone them.

PR #942, opened 2019-07-31
Redirect older Client API to the new data type API

### LINKED ISSUE #938: Redirect old safe_core::Client functions to new data types
Temporarily, we can use the same API facade we have in `safe_core::Client`, but under the hood we can redirect the calls to use the new data types (so that we have only the `send` function in mock Routing). The functions and calls to be replaced:

- `PutIData` (`mock_routing::put_idata`) -> put published IData
- `GetIData` (`mock_routing::get_idata`) -> get published IData
- `PutMData` (`mock_routing::put_mdata`) -> put sequenced unpublished Mutable Data (SeqMData)
- `GetMDataVersion` (`mock_routing::get_mdata_version`) -> get version of SeqMData
- `GetMData` (`mock_routing::get_mdata`) -> get SeqMData, convert the struct if needed.
- `GetMDataShell` (`mock_routing::get_mdata_shell`) -> get shell of SeqMData, convert the shell struct if needed (reusing fns from `GetMData`)
- `ListMDataEntries` (`mock_routing::list_mdata_entries`) -> list entries of SeqMData, convert the returned entries type if needed.
- `ListMDataKeys` (`mock_routing::list_mdata_keys`) -> list keys of SeqMData, convert the returned type if needed.
- `ListMDataValues` (`mock_routing::list_mdata_values`) -> list values of SeqMData, convert the returned type if needed.
- `GetMDataValue` (`mock_routing::get_mdata_value`) -> get a single value of SeqMData, convert the returned type if needed.
- `MutateMDataEntries` (`mock_routing::mutate_mdata_entries`) -> mutate SeqMData, convert the argument types if needed.
- `ListMDataPermissions` (`mock_routing::list_mdata_permissions`) -> list perms of SeqMData, convert the perm types if needed.
- `ListMDataUserPermissions` (`mock_routing::list_mdata_user_permissions`) -> list user perms of SeqMData, convert the perm types if needed.
- `SetMDataUserPermissions` (`mock_routing::set_mdata_user_permissions`) -> set user perms of SeqMData, convert the argument types if needed.
- `DelDataUserPermissions` (`mock_routing::del_mdata_user_permissions`) -> delete user perms of SeqMData, convert the argument types if needed.
- `ChangeMDataOwner` (`mock_routing::change_mdata_owner`) -> no op (we don't need it right now).

lionel1704:
Resolved #938


safe_core/src/client/mock/vault.rs
```diff
@@ -1429,6 +1429,12 @@ impl Vault {
             _ => return Err(SndError::AccessDenied),
         }
         if self.contains_data(&data_name) {
+            // Published Immutable Data is de-duplicated
+            if let DataId::Immutable(addr) = data_name {
+                if let IDataAddress::Pub(_) = addr {
```
m-cat: Can use `addr.is_pub()` from the latest safe-nd

safe_app/src/ffi/tests/nfs.rs
```diff
@@ -568,19 +568,17 @@ fn delete_then_open_file() {
         unwrap!(call_1(|ud, cb| file_close(&app, write_h, ud, cb)))
     };
 
-    // Update file in container.
-    let version: u64 = unsafe {
-        unwrap!(call_1(|ud, cb| dir_update_file(
+    // Add the file back to the container
+    unsafe {
+        unwrap!(call_0(|ud, cb| dir_insert_file(
```
m-cat: Should the documentation above this test be updated also?

safe_authenticator/src/tests/share_mdata.rs
```diff
@@ -282,13 +282,9 @@ fn share_some_mdatas_with_ownership_error() {
         ok!(PublicKey::from(unwrap!(client.public_bls_key())))
     }));
 
-    let someone_else = PublicKey::from(threshold_crypto::SecretKey::random().public_key());
-
     let ownerss = vec![
         btree_set![user /* , someone_else */], // currently can't handle having multiple owners
-        btree_set![someone_else],
```
nbaksalyar: I must've missed something, but I think this test should still fail with a randomly-generated `PublicKey`?
Also, it has 2 duplicate sets of `user`, so the test is needlessly redundant I believe.

lionel1704: I can't recall why this change was needed. Let me quickly check.

safe_core/src/client/mock/vault.rs
lionel1704: Can we address this in a separate PR with the other breaking changes from safe-nd?

m-cat: It's not a breaking change, just a small QoL improvement and it's already been merged. Would make the code here shorter, but it's up to you.

safe_app/src/tests/mutable_data.rs
```diff
@@ -98,18 +98,12 @@ fn md_created_by_app_2() {
...
         ));
         let name2 = name;
-        let cl2 = client.clone();
         client
             .put_mdata(mdata)
-            .then(move |res| {
-                unwrap!(res);
-                cl2.change_mdata_owner(name, DIR_TAG, sign_pk, 1)
-            })
             .then(move |res| -> Result<_, ()> {
```
m-cat: I don't see the point of these lines. Matching on a `Result` and doing nothing if it's `Ok`, and panicking on `Err`, is the same as `unwrap` isn't it?

safe_app/src/tests/mutable_data.rs
```diff
@@ -119,24 +113,18 @@ fn md_created_by_app_2() {
...
-                .then(move |res| {
-                    unwrap!(res);
-                    cl3.change_mdata_owner(name, DIR_TAG, sign_pk, 1)
-                })
                 .then(move |res| -> Result<(), ()> {
                     match res {
-                        Ok(()) => panic!("It should fail"),
-                        Err(CoreError::RoutingClientError(ClientError::AccessDenied)) => (),
-                        Err(x) => panic!("Expected ClientError::AccessDenied. Got {:?}", x),
+                        Ok(()) => (),
```
m-cat: Same comment as above.

safe_authenticator/src/access_container.rs
```diff
@@ -181,15 +183,34 @@ pub fn put_entry(
...
-        client,
-        access_container.name(),
-        access_container.type_tag(),
-        actions.into(),
-    )
-    .map_err(From::from)
-    .into_box()
+    client
+        .get_mdata_version(access_container.name(), access_container.type_tag())
+        .map_err(AuthError::from)
```
m-cat: Are multiple `map_err` lines necessary here?

lionel1704: I believe so. Removing any of those returns compilation errors.

PR #943, opened 2019-07-31
feat/vault: add `sign` field to `compose_message`

### LINKED ISSUE #916: Get requests for published data need not be signed
When requesting for published data, the request need not be signed. The signature field in the `Message::Request` should be `None`.

For this, add a `signed: bool` field to the `compose_message` function and conditionally set it to None.

In the mock vault, verify that all requests that need a signature are signed and only GETs for published data are allowed to have no signature

m-cat:
Fixes #916

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -156,7 +156,7 @@ pub trait Client: Clone + 'static {
 
     /// Create a `Message` from the given request.
     /// This function adds the requester signature and message ID.
-    fn compose_message(&self, req: Request) -> Message;
+    fn compose_message(&self, req: Request, sign: bool) -> Message;
 
     /// Return the public and secret signing keys.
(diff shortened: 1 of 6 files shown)


safe_core/src/client/mod.rs
```diff
@@ -1217,7 +1227,7 @@ pub trait Client: Clone + 'static {
                 address,
                 permissions_index,
                 user,
-            },
+            }, address.is_unpub()
```
lionel1704: Can be explicitly set to `true`

safe_core/src/client/mod.rs
```diff
@@ -1252,7 +1262,7 @@ pub trait Client: Clone + 'static {
                 address,
                 permissions_index,
                 public_key,
-            },
+            }, address.is_unpub()
```
lionel1704: Can be explicitly `false`

PR #947, opened 2019-08-02
Deprecate the use of the older Immutable Data API

### LINKED ISSUE #961: Deprecate the older Immutable Data API
The older immutable data API currently redirects to the new API. The usage of the API should be replaced to the new API and the tests should cover both published and unpublished variants as well.

lionel1704:
Depends on maidsafe/safe-nd#94
Closes #961

### DIFF
--- a/safe_app/examples/client_stress_test.rs
+++ b/safe_app/examples/client_stress_test.rs
@@ -69,7 +69,7 @@ use futures::Future;
 use rand::{Rng, SeedableRng};
 use safe_app::{Client, CoreError, CoreFuture, FutureExt, MutableData, PubImmutableData};
 use safe_authenticator::{AuthClient, Authenticator};
-use safe_nd::{IData, IDataAddress, PublicKey, XorName};
+use safe_nd::{IData, PublicKey, XorName};
 use std::sync::mpsc;
 
 fn random_mutable_data<R: Rng>(type_tag: u64, public_key: &PublicKey, rng: &mut R) -> MutableData {
@@ -256,12 +256,11 @@ fn main() {
 
                     fut.and_then(move |data| {
                         // Get all the chunks again.
-                        c3.get_idata(IDataAddress::Pub(*data.name()))
-                            .map(move |retrieved_data| {
-                                println!("Retrieved chunk #{}: {:?}", i, data.name());
-                                assert_eq!(data, retrieved_data);
-                                Ok(())
-                            })
+                        c3.get_idata(*data.address()).map(move |retrieved_data| {
+                            println!("Retrieved chunk #{}: {:?}", i, data.name());
+                            assert_eq!(data, retrieved_data);
+                            Ok(())
+                        })
                     })
                     .into_box()
                 }
@@ -307,11 +306,10 @@ fn put_idata(client: &AuthClient, data: IData, i: usize) -> Box<CoreFuture<IData
             println!("Put PubImmutableData chunk #{}: {:?}", i, data.name());
 
             // Get the data
-            c2.get_idata(IDataAddress::Pub(*data.name()))
-                .map(move |retrieved_data| {
-                    assert_eq!(data, retrieved_data);
-                    retrieved_data
-                })
+            c2.get_idata(*data.address()).map(move |retrieved_data| {
+                assert_eq!(data, retrieved_data);
+                retrieved_data
+            })
         })
         .and_then(move |retrieved_data| {
             println!(
--- a/safe_app/examples/client_stress_test.rs
+++ b/safe_app/examples/client_stress_test.rs
@@ -69,7 +69,7 @@ use futures::Future;
 use rand::{Rng, SeedableRng};
 use safe_app::{Client, CoreError, CoreFuture, FutureExt, MutableData, PubImmutableData};
 use safe_authenticator::{AuthClient, Authenticator};
-use safe_nd::{PublicKey, XorName};
+use safe_nd::{IData, IDataAddress, PublicKey, XorName};
 use std::sync::mpsc;
 
 fn random_mutable_data<R: Rng>(type_tag: u64, public_key: &PublicKey, rng: &mut R) -> MutableData {
@@ -87,7 +87,7 @@ fn random_mutable_data<R: Rng>(type_tag: u64, public_key: &PublicKey, rng: &mut 
 
 enum Data {
     Mutable(MutableData),
-    Immutable(PubImmutableData),
+    Immutable(IData),
 }
 
 fn main() {
@@ -207,7 +207,7 @@ fn main() {
         // Construct data
         let data = PubImmutableData::new(rng.gen_iter().take(1024).collect());
         println!("{:?}", data.name());
-        stored_data.push(Data::Immutable(data));
+        stored_data.push(Data::Immutable(data.into()));
     }
 
     let (tx, rx) = mpsc::channel();
@@ -256,11 +256,13 @@ fn main() {
 
                     fut.and_then(move |data| {
                         // Get all the chunks again.
-                        c3.get_idata(*data.name()).map(move |retrieved_data| {
-                            assert_eq!(data, retrieved_data);
-                            println!("Retrieved chunk #{}: {:?}", i, data.name());
-                            Ok(())
-                        })
+                        c3.get_idata(IDataAddress::Pub(*data.name())).map(
+                            move |retrieved_data| {
+                                println!("Retrieved chunk #{}: {:?}", i, data.name());
+                                assert_eq!(data, retrieved_data);
+                                Ok(())
+                            },
+                        )
                     })
                     .into_box()
                 }
@@ -297,11 +299,7 @@ fn main() {
     println!("Done");
 }
 
-fn put_idata(
-    client: &AuthClient,
-    data: PubImmutableData,
-    i: usize,
-) -> Box<CoreFuture<PubImmutableData>> {
+fn put_idata(client: &AuthClient, data: IData, i: usize) -> Box<CoreFuture<IData>> {
     let c2 = client.clone();
 
     client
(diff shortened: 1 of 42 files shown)


safe_app/src/tests/mod.rs
```diff
@@ -383,10 +383,10 @@ fn published_data_access() {
             let client3 = client.clone();
 
             client
-                .get_pub_idata(*pub_idata.name())
+                .get_idata(IDataAddress::Pub(*pub_idata.name()))
```
m-cat: You can just call `.address()` here (and below) :)

safe_app/src/tests/mod.rs
```diff
@@ -317,9 +317,9 @@ fn unregistered_client() {
         let client3 = client.clone();
 
         client
-            .get_pub_idata(*pub_idata.name())
+            .get_idata(IDataAddress::Pub(*pub_idata.name()))
```
m-cat: Same comment: you can call `.address()` here

safe_app/examples/client_stress_test.rs
```diff
@@ -256,11 +256,12 @@ fn main() {
 
                     fut.and_then(move |data| {
                         // Get all the chunks again.
-                        c3.get_idata(*data.name()).map(move |retrieved_data| {
-                            assert_eq!(data, retrieved_data);
-                            println!("Retrieved chunk #{}: {:?}", i, data.name());
-                            Ok(())
-                        })
+                        c3.get_idata(IDataAddress::Pub(*data.name()))
```
m-cat: Can call `data.address()` here.

safe_app/examples/client_stress_test.rs
```diff
@@ -310,10 +307,11 @@ fn put_idata(
             println!("Put PubImmutableData chunk #{}: {:?}", i, data.name());
 
             // Get the data
-            c2.get_idata(*data.name()).map(move |retrieved_data| {
-                assert_eq!(data, retrieved_data);
-                retrieved_data
-            })
+            c2.get_idata(IDataAddress::Pub(*data.name()))
```
m-cat: Same here.

safe_app/src/ffi/immutable_data.rs
```diff
@@ -177,7 +182,7 @@ pub unsafe extern "C" fn idata_fetch_self_encryptor(
             let context2 = context.clone();
             let context3 = context.clone();
 
-            immutable_data::get_value(client, &name, None)
+            immutable_data::get_value(client, &name, published, None)
```
m-cat: Having a boolean `published` flag for FFI makes sense, but I'm thinking the Rust API should take an `IDataAddress` instead of `name` and `published` separately.

safe_app/src/ffi/immutable_data.rs
```diff
@@ -216,8 +222,13 @@ pub unsafe extern "C" fn idata_serialised_size(
         let name = XorName(*name);
 
         (*app).send(move |client, _| {
+            let address = if published {
```
m-cat: Can use `IDataAddress::from_kind` here.

safe_core/src/immutable_data.rs
```diff
@@ -27,12 +27,13 @@ enum DataTypeEncoding {
 pub fn create(
     client: &impl Client,
     value: &[u8],
+    published: bool,
     encryption_key: Option<shared_secretbox::Key>,
-) -> Box<CoreFuture<PubImmutableData>> {
+) -> Box<CoreFuture<IData>> {
     trace!("Creating conformant PubImmutableData.");
```
m-cat: I think this can be updated, since we now support Unpub Idata as well here.

safe_core/src/immutable_data.rs
```diff
@@ -89,51 +91,61 @@ pub fn extract_value(
 pub fn get_value(
     client: &impl Client,
     name: &XorName,
+    published: bool,
     decryption_key: Option<shared_secretbox::Key>,
 ) -> Box<CoreFuture<Vec<u8>>> {
     let client2 = client.clone();
+    let address = if published {
```
m-cat: Same comment about `from_kind`.

safe_core/src/client/mod.rs
```diff
@@ -391,41 +375,26 @@ pub trait Client: Clone + 'static {
...
-                        Err(e) => Err(e).map_err(CoreError::from),
-                    },
-                    _ => Err(CoreError::ReceivedUnexpectedEvent),
-                }
-            })
-            .into_box()
+    /// Put immutable data to the network.
+    fn put_idata(&self, data: impl Into<IData>) -> Box<CoreFuture<()>> {
+        let idata: IData = data.into();
+        trace!("Put Published IData at {:?}", idata.name());
```
m-cat: Looks like there's an outdated `trace` here as well.

tests/src/lib.rs
```diff
@@ -75,15 +76,17 @@ fn unregistered_client() {
...
     }
 
     // Unregistered Client should be able to retrieve the data.
     let app = unwrap!(App::unregistered(|| (), None));
     unwrap!(run(&app, move |client, _context| {
-        let _ = client.get_pub_idata(*orig_data.name()).map(move |data| {
-            assert_eq!(data, orig_data);
-        });
+        let _ = client
+            .get_idata(IDataAddress::Pub(*orig_data.name()))
```
m-cat: You can just do `orig_data.address()` here, and in several other places in the PR.

safe_core/src/nfs/writer.rs
```diff
@@ -40,12 +41,18 @@ impl<C: Client> Writer<C> {
         storage: SelfEncryptionStorage<C>,
         file: File,
         mode: Mode,
+        published: bool,
```
m-cat: Don't need this parameter anymore now that `published` is in `File`. Same for several other functions in this PR.

safe_core/src/nfs/reader.rs
```diff
@@ -28,9 +28,10 @@ impl<C: Client> Reader<C> {
         client: C,
         storage: SelfEncryptionStorage<C>,
         file: &File,
+        published: bool,
```
m-cat: One lost spot where `published` parameter can be removed, making the API simpler :)

Overall the PR looks great, the code is very clean now!

safe_core/src/nfs/tests.rs
```diff
@@ -155,14 +158,121 @@ fn file_fetch_public_md() {
...
+            let c4 = client.clone();
+            let c5 = client.clone();
+            let c6 = client.clone();
+            let c7 = client.clone();
+
+            let root = unwrap!(MDataInfo::random_public(MDataKind::Unseq, DIR_TAG));
+            let root2 = root.clone();
+
+            create_dir(client, &root, btree_map![], btree_map![])
+                .then(move |res| {
```
m-cat: Just noticed this: you can simply do `.and_then` instead of the pattern of `.then` followed by `unwrap!` or `assert!`. It would save a lot of code throughout this test function. Just an FYI for the future.

safe_core/src/nfs/tests.rs
```diff
@@ -155,14 +158,121 @@ fn file_fetch_public_md() {
     });
 }
 
+// Test inserting files to, and fetching from, a public mdata.
+// Insert a file as Unpublished Immutable data and verify that it can be fetched.
+// Other clients should not be able to fetch the file.
+// After deletion the file should not be accessible anymore.
+#[allow(unsafe_code)]
+#[test]
+fn files_stored_in_unpublished_idata() {
```
m-cat: I had trouble understanding this function, the control flow is somewhat complex and not easily comprehended at a glance. I think it would be more idiomatic not to use channels and to just to split it up, if possible:

```rust
client1.do_stuff()...;

client2.do_stuff()...;

client1.do_more_stuff()...;
```

Otherwise, I think some comments are needed here to guide the reader.

safe_core/src/immutable_data.rs
```diff
@@ -88,44 +89,49 @@ pub fn extract_value(
...
 
 // TODO: consider rewriting these two function to not use recursion.
 
-fn pack(client: impl Client, value: Vec<u8>) -> Box<CoreFuture<PubImmutableData>> {
-    let data = PubImmutableData::new(value);
+fn pack(client: impl Client, value: Vec<u8>, published: bool) -> Box<CoreFuture<IData>> {
+    let data: IData = if published {
+        PubImmutableData::new(value).into()
+    } else {
+        UnpubImmutableData::new(value, unwrap!(client.public_bls_key()).into()).into()
```
nbaksalyar: I'd be wary of using `unwrap!` in the production code; this function call will panic if called on unregistered client. Why can't we return an error future instead?

In addition to that, I think it's better to add a new function in `Client` to return `PublicKey` rather than implicitly convert the BLS key specifically since we might want to change this in the future. So I would suggest just add this function to Clients:
```rust
pub fn public_key() -> Option<PublicKey> { self.public_bls_key() }
```
and not `unwrap!` it but just return an error:
```rust
fry!(client.public_key().ok_or_else(|| CoreError::from("Attempted to use unregistered client to send mutation request")))
```

safe_core/src/nfs/data_map.rs
```diff
@@ -15,17 +15,20 @@ use crate::nfs::NfsFuture;
...
 // If the `DataMap` is encrypted, an `encryption_key` must be passed in to decrypt it.
 pub fn get(
     client: &impl Client,
     name: &XorName,
+    published: bool,
     encryption_key: Option<shared_secretbox::Key>,
 ) -> Box<NfsFuture<DataMap>> {
-    immutable_data::get_value(client, name, encryption_key)
+    let kind = IDataKind::from_flag(published);
+    let address = IDataAddress::from_kind(kind, *name);
```
nbaksalyar: Maybe it'd be easier to just have this `address` as a function argument?
Just a suggestion though.

PR #948, opened 2019-08-02
 fix/vault: verify if the app is still authorised when reading data 

### LINKED ISSUE #915: Verify if the app is still authorised when reading data from the network
When reading data from the network, we should first check if an app still exists at the list of auth'ed apps at the source elders. When this app is revoked, there will be no re-encryption of data. The request will be blocked at the source elders itself. This check is missing in the mock vault implementation.

m-cat:
Fixes #915

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -14,12 +14,12 @@ use fs2::FileExt;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use routing::{Authority, ClientError, MutableData as OldMutableData};
 use safe_nd::{
-    verify_signature, AData, ADataAction, ADataAddress, ADataIndex, AppendOnlyData, Coins,
-    Error as SndError, IData, IDataAddress, LoginPacket, MData, MDataAction, MDataAddress,
-    MDataKind, Message, PublicId, PublicKey, Request, Response, SeqAppendOnly, Transaction,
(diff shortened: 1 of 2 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -1390,6 +1409,42 @@ impl Vault {
             Ok(())
         }
     }
+
+    fn list_auth_keys_and_version(
+        &mut self,
+        name: &XorName,
+    ) -> (BTreeMap<PublicKey, AppPermissions>, u64) {
+        if self.get_account(&name).is_none() {
+            self.insert_account(*name);
```
nbaksalyar: I think it shouldn't create a new account implicitly during the get request (which is also not paid for).
If an account doesn't exist, just return the default value, so the entire function can be rewritten as e.g.
```rust
self.get_account(&name)
  .map(|| (account.auth_keys().clone(), account.version()))
  .unwrap_or_else(Default::default)
```

m-cat:
> There are some FIXMEs in the [revocation tests](https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_authenticator/src/tests/revocation.rs) that need to be addressed along with this PR. Apps could still access data after revocation so those tests were failing. Now with this fixed the tests need to be corrected.
> Could you please address this as well?

Addressed this but it looks like the tests are failing, will debug this tomorrow.

PR #959, opened 2019-08-05
SAFE Authenticator Migration to the new data types

Yoga07:
- Fixes all safe_authenticator tests to the new impl
- Overhauls recovery functions to the new types
- Changes `share_mdata` functions and structs to the new type
- Removes re-encryption while revocation
- Revamps safe_core nfs which used old Mdata to the new type
- Safe_app's `object_cache` adapted temporarily

### DIFF
--- a/safe_authenticator/src/revocation.rs
+++ b/safe_authenticator/src/revocation.rs
@@ -14,12 +14,12 @@ use crate::client::AuthClient;
 use crate::config::{self, AppInfo, RevocationQueue};
 use futures::future::{self, Either, Loop};
 use futures::Future;
-use routing::User;
 use safe_core::recovery;
 use safe_core::{client::AuthActions, Client, CoreError, FutureExt, MDataInfo};
-use safe_nd::{Error as SndError, PublicKey};
+use safe_nd::{Error, MDataAddress, PublicKey};
 use std::collections::HashMap;
 
+//type MDataEntries = BTreeMap<Vec<u8>, MDataValue>;
 type Containers = HashMap<String, MDataInfo>;
 
 /// Revoke app access using a revocation queue.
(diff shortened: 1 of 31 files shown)


safe_core/src/nfs/tests.rs
```diff
@@ -599,7 +602,7 @@ fn file_delete_then_add() {
             })
             .then(move |res| {
                 let (version, file, dir) = unwrap!(res);
-                assert_eq!(version, 0);
+                assert_eq!(version, 1);
```
lionel1704: Version should be 0. Same reason as above.

safe_authenticator/src/revocation.rs
```diff
@@ -128,37 +128,33 @@ fn revoke_single_app(client: &AuthClient, app_id: &str) -> Box<AuthFuture<()>> {
...
+                            .map(|(name, (mdata_info, _))| (name, mdata_info))
+                            .collect();
+
+                        clear_from_access_container_entry(&c4, app, version, containers)
                     }
-                },
-            )
+                    // If the access container entry was not found, the entry must have been
+                    // deleted with the app having stayed on the revocation queue.
+                    Err(_e) => ok!(()),
```
lionel1704: It's possible that other errors can occur here as well. We should not evaluate all of them to `ok!(())`. To keep the changes simple, let's keep the return type of `access_container::fetch_entry` as `Result<u64, Option<..>>`.
I will be making this change when updating the PR.

safe_core/src/client/recovery.rs
```diff
@@ -261,39 +277,45 @@ fn fix_entry_actions(
...
-            (Some(false), _) | (_, Some(false)) => set.deny(action),
-            _ => set,
-        }
-    })
+    actions
+        .iter()
+        .fold(MDataPermissionSet::new(), |set, &action| {
+            if a.is_allowed(action) | b.is_allowed(action) {
+                set.allow(action)
+            } else if (a.is_allowed(action) == false) | (b.is_allowed(action) == false) {
```
lionel1704:
```suggestion
            } else if !a.is_allowed(action) | !b.is_allowed(action) {
```

lionel1704:
Closing in favour of #960

PR #973, opened 2019-08-09
SAFE App Migration to the new data types

### LINKED ISSUE #962: Deprecate the use of older MutableData API in safe_app
In #960 the usage of the older mutable data API has been removed in safe_auth. Tests for unpublished mutable data are also available in this [test module](https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_app/src/tests/unpublished_mutable_data.rs).

The older mutable data tests in safe_app can be removed along with any other usages of the older API. As a conclusion of the task, the older API itself should be removed from the Client trait.

Yoga07:
Resolves #962


safe_app/src/ffi/mutable_data/entries.rs
```diff
@@ -87,11 +86,9 @@ pub unsafe extern "C" fn seq_mdata_entries_len(
     })
 }
 
-/// Get the entry value at the given key.
```
lionel1704: Oops? :smile:

safe_app/src/ffi/mutable_data/entries.rs
```diff
@@ -489,7 +424,7 @@ mod tests {
                 &app,
                 perms_h,
                 app_pk_handle,
-                &permission_set_into_repr_c(perms_set),
+                &permission_set_into_repr_c_new(perms_set),
```
lionel1704: If the older function is not used, can it be removed and the `_new` suffix removed from here?

safe_app/src/ffi/mutable_data/helper.rs
```diff
@@ -8,20 +8,17 @@
...
 
 use crate::errors::AppError;
-use crate::ffi::mutable_data::permissions::USER_ANYONE;
 use crate::ffi::object_cache::{MDataPermissionsHandle, PubKeyHandle};
 use crate::object_cache::ObjectCache;
-use routing::{PermissionSet, User};
+use safe_nd::{MDataPermissionSet, PublicKey};
 use std::collections::BTreeMap;
 
 // Retrieve the sign key corresponding to the handle from the object cache and wrap it
```
lionel1704: You'll need to change the comment as well.

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -26,15 +26,16 @@ use ffi_utils::{
...
 };
 use futures::Future;
-use routing::MutableData;
+use maidsafe_utilities::serialisation::serialised_size;
 use safe_core::ffi::ipc::req::PermissionSet;
 use safe_core::ffi::ipc::resp::MDataKey;
 use safe_core::ffi::ipc::resp::MDataValue;
 use safe_core::ffi::MDataInfo;
-use safe_core::ipc::req::{permission_set_clone_from_repr_c, permission_set_into_repr_c};
+use safe_core::ipc::req::{permission_set_clone_from_repr_c_new, permission_set_into_repr_c_new};
```
lionel1704: Same comment about removing old functions and renaming the new.

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -93,22 +94,16 @@ pub unsafe extern "C" fn mdata_put(
...
+            let data = SeqMutableData::new_with_data(
+                info.name(),
+                info.type_tag(),
+                entries,
+                permissions,
+                owner_key,
             );
 
             client
-                .put_mdata(data)
```
lionel1704: Now that the old Client API is deprecated, can it be removed from the client trait as well?

safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -132,7 +127,7 @@ pub unsafe extern "C" fn mdata_get_version(
         let info = NativeMDataInfo::clone_from_repr_c(info)?;
 
         send(app, user_data, o_cb, move |client, _| {
-            client.get_mdata_version(info.name(), info.type_tag())
+            client.get_mdata_version_new(*info.address())
```
lionel1704: If the older API is deprecated, the `_new` suffix can be removed.

safe_app/src/ffi/mutable_data/tests.rs
```diff
@@ -21,13 +21,12 @@ use ffi_utils::test_utils::{
     call_0, call_1, call_vec, call_vec_u8, send_via_user_data, sender_as_user_data,
 };
 use ffi_utils::{vec_clone_from_raw_parts, FfiResult};
-use routing::{Action, PermissionSet as NativePermissionSet};
 use safe_core::ffi::ipc::req::PermissionSet as FfiPermissionSet;
-use safe_core::ffi::MDataKind;
-use safe_core::ipc::req::{permission_set_clone_from_repr_c, permission_set_into_repr_c};
+use safe_core::ffi::{MDataInfo as FFIMD, MDataKind};
```
lionel1704: We don't need to rename this here

safe_app/src/ffi/mutable_data/mod.rs
Yoga07: Hmm, a lot of the old tests still make use of these APIs. I think it'd be better to rip all these and adapt the tests in a separate PR.

safe_app/src/object_cache.rs
```diff
@@ -30,10 +29,9 @@ pub struct ObjectCache {
     cipher_opt: Store<CipherOpt>,
     encrypt_key: Store<box_::PublicKey>,
     secret_key: Store<shared_box::SecretKey>,
-    mdata_entries: Store<BTreeMap<Vec<u8>, Value>>,
-    mdata_entries_new: Store<BTreeMap<Vec<u8>, MDataSeqValue>>,
-    mdata_entry_actions: Store<BTreeMap<Vec<u8>, EntryAction>>,
-    mdata_permissions: Store<BTreeMap<User, PermissionSet>>,
+    mdata_entries: Store<BTreeMap<Vec<u8>, MDataSeqValue>>,
```
m-cat: We export the `MDataSeqEntries` type which can be used here instead.

Also, are we only supporting seq mdata for now? Maybe this should be `MDataEntries` which can be either sequenced or unsequenced.

safe_app/src/object_cache.rs
```diff
@@ -30,10 +29,9 @@ pub struct ObjectCache {
...
+    mdata_entry_actions: Store<BTreeMap<Vec<u8>, MDataSeqEntryAction>>,
```
m-cat: We export the `MDataSeqEntryActions` type which can be used here instead.

safe_app/src/object_cache.rs
```diff
@@ -134,33 +130,24 @@ impl_cache!(
 );
 impl_cache!(
     mdata_entries,
-    BTreeMap<Vec<u8>, Value>,
+    BTreeMap<Vec<u8>, MDataSeqValue>,
```
m-cat: This is equivalent to `MDataSeqEntries`.

safe_app/src/object_cache.rs
```diff
@@ -134,33 +130,24 @@ impl_cache!(
...
-    MDataEntriesHandle,
-    InvalidMDataEntriesHandle,
-    get_mdata_entries_new,
-    insert_mdata_entries_new,
-    remove_mdata_entries_new
-);
 impl_cache!(
     mdata_entry_actions,
-    BTreeMap<Vec<u8>, EntryAction>,
+    BTreeMap<Vec<u8>, MDataSeqEntryAction>,
```
m-cat: This can be `MDataSeqEntryActions`.

safe_app/src/tests/unpublished_mutable_data.rs
```diff
@@ -251,6 +254,46 @@ fn md_created_by_app_2() {
...
+// MaidManagers. Should pass when it lists the owner's sign_pk instead.
+#[test]
+#[allow(unsafe_code)]
+fn md_created_by_app_3() {
+    use crate::App;
+
+    let app_id = unwrap!(utils::generate_random_string(10));
+    let app_id = unwrap!(CString::new(app_id));
+    let app: *mut App =
+        unsafe { unwrap!(call_1(|ud, cb| test_create_app(app_id.as_ptr(), ud, cb))) };
```
m-cat: Why use `test_create_app` instead of `create_app` like the other tests here? If we get rid of this then we won't have any unsafe code, and we can remove `allow(unsafe_code)` above.

safe_app/src/tests/unpublished_mutable_data.rs
```diff
@@ -251,6 +254,46 @@ fn md_created_by_app_2() {
...
+                let mdata = SeqMutableData::new_with_data(
+                    name,
+                    DIR_TAG,
+                    BTreeMap::new(),
+                    BTreeMap::new(),
+                    owners,
+                );
+                cl2.put_seq_mutable_data(mdata)
+            })
+            .map_err(|e| panic!("{:?}", e))
```
m-cat: I'm pretty sure this line can be removed as we are `unwrap!`ing the result of `run`, but you might want to double check. If it's superfluous, then remove it from the other tests also.

safe_app/src/tests/unpublished_mutable_data.rs
```diff
@@ -251,6 +254,46 @@ fn md_created_by_app_2() {
...
+    let app_id = unwrap!(CString::new(app_id));
+    let app: *mut App =
+        unsafe { unwrap!(call_1(|ud, cb| test_create_app(app_id.as_ptr(), ud, cb))) };
+
+    unwrap!(run(unsafe { &*app }, |client: &AppClient, _app_context| {
+        let owners = PublicKey::from(client.public_bls_key());
+        let name: XorName = new_rand::random();
+        let mdata =
+            SeqMutableData::new_with_data(name, DIR_TAG, BTreeMap::new(), BTreeMap::new(), owners);
+        let cl2 = client.clone();
```
m-cat: We usually call Client clones `c2`, `c3` etc.

m-cat: Can you fix this in the rest of the file as well? Should be a quick search-replace.

safe_app/src/tests/unpublished_mutable_data.rs
Yoga07: Removing `map_err` did not work as the `run` function expects `CoreError` and the futures return an `AppError`. Therefore we'll anyhow have to call map_err to convert. So I guess it's okay to leave it as it is now and panic there? :)

safe_app/src/object_cache.rs
Yoga07: I am having a problem after changing this @m-cat . I am getting this error at places where these functions are being used:
```
error[E0599]: no method named `insert` found for type `std::cell::RefMut<'_, safe_nd::mutable_data::SeqEntryActions>` in the current scope
   --> safe_app\src\ffi\mutable_data\entry_actions.rs:132:29
    |
132 |             let _ = actions.insert(key, action);
```
It is still being a `RefMut` after fetching as well.

safe_app/src/object_cache.rs
Yoga07: >Also, are we only supporting seq mdata for now?

Yes, we are currently only re-wirting the older MData as Seq MData now. I am guessing we'll be dealing with Unseq MData and AppendOnly data later on in safe_app.

safe_app/src/tests/unpublished_mutable_data.rs
m-cat: Looks like `run` expects `AppError` actually. And I think I would still prefer converting errors since that is the pattern we tend to use. I did a grep and it seems like we don't use the `map_err` - `panic` pattern anywhere outside this file.

Yoga07: Hmm, this error confuses me. 🤔 
```
error[E0271]: type mismatch resolving `<futures::future::then::Then<std::boxed::Box<dyn futures::future::Future<Item = (), Error = safe_core::CoreError>>, std::boxed::Box<dyn futures::future::Future<Item = (), Error = safe_core::CoreError>>, [closure@safe_app\src\tests\unpublished_mutable_data.rs:271:19: 286:14 c2:_, name:_]> as futures::future::IntoFuture>::Error == errors::AppError`
   --> safe_app\src\tests\unpublished_mutable_data.rs:263:13
    |
263 |     unwrap!(run(&app, |client: &AppClient, _app_context| {
    |             ^^^ expected enum `safe_core::CoreError`, found enum `errors::AppError`
    |
    = note: expected type `safe_core::CoreError`
               found type `errors::AppError`
note: required by `run`
   --> safe_app\src\lib.rs:453:1
    |
453 | / pub fn run<F, I, T>(app: &App, f: F) -> Result<T, AppError>
454 | | where
455 | |     F: FnOnce(&AppClient, &AppContext) -> I + Send + 'static,
456 | |     I: IntoFuture<Item = T, Error = AppError> + 'static,
...   |
470 | |     rx.recv()?
471 | | }
    | |_^

error: aborting due to previous error
```

safe_app/src/object_cache.rs
m-cat: `MDataSeqEntryActions` doesn't have methods like `insert` or `get`, you need to call `actions()` on it first.

safe_app/src/tests/unpublished_mutable_data.rs
m-cat: Replacing it with

```rust
            .map_err(AppError::from)
```

seems to have done the trick :)

safe_app/src/object_cache.rs
Yoga07: Currently `.actions()` only give us a reference of the `actions` variable inside the `MDataSeqEntryActions` struct. But we borrow them mutably at the places where we call. We need to make a change in the safe_nd crate for this change (probably add `.into_actions()`?) Do we have any workarounds? 
But for the time being, I'd vote for keeping it as  `BTreeMap<Vec<u8>,MDataSeqEntryAction>` now. :) Your opinions @m-cat @lionel1704 @nbaksalyar ?

lionel1704: I remember the need for `into_actions()` somewhere else as well. (safe_auth ?)
So it's probably worth adding that in to safe-nd.

Issue #965, opened 2019-08-07
Add tests for Published Sequenced AppendOnly data in safe_app


lionel1704:
Similar to [this test module](https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_app/src/tests/unpublished_mutable_data.rs) there should be extensive tests to test the various scenarios of Unpublished Sequenced AppendOnly data.

lionel1704:
Resolved in #976

Issue #964, opened 2019-08-07
Add tests for Unpublished Unsequenced AppendOnly data


lionel1704:
Similar to [this test module](https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_app/src/tests/unpublished_mutable_data.rs) there should be extensive tests to test the various scenarios of Unpublished Unsequenced AppendOnly data.

lionel1704:
Resolved in #976

PR #976, opened 2019-08-12
Add tests for AData in safe_app

### LINKED ISSUE #963: Add tests for Unpublished Sequenced AppendOnly data in safe_app
Similar to [this test module](https://github.com/maidsafe/safe_client_libs/blob/experimental/safe_app/src/tests/unpublished_mutable_data.rs) there should be extensive tests to test the various scenarios of Unpublished Sequenced AppendOnly data.

lionel1704:
Resolves #963, #964, #965, #966


safe_app/src/tests/append_only_data/published_sequenced.rs
```diff
@@ -0,0 +1,387 @@
...
+use safe_nd::{
+    ADataAddress, ADataAppendOperation, ADataEntry, ADataIndex, ADataOwner, ADataPubPermissionSet,
+    ADataPubPermissions, ADataUser, AppendOnlyData, Error as SndError, PubSeqAppendOnlyData,
+    PublicKey, SeqAppendOnly, XorName,
+};
+use std::collections::BTreeMap;
+use std::sync::mpsc;
+use std::thread;
+
+// AD created by app. App lists it's own sign_pk in owners field. Put should fail - Rejected at the client handlers.
```
m-cat: Grammar: "it's" should be "its". Grep shows a few more cases of this.

safe_app/src/tests/append_only_data/published_sequenced.rs
```diff
@@ -0,0 +1,387 @@
...
+            .put_adata(invalid_data.into())
+            .then(move |res| {
+                match res {
+                    Err(CoreError::NewRoutingClientError(SndError::InvalidOwners)) => (),
+                    Ok(_) => panic!("Unexpected success"),
+                    Err(err) => panic!("Unexpected error {:?}", err),
+                }
+                client2.put_adata(valid_data.into())
+            })
+            .map_err(|e| panic!("{:?}", e))
```
m-cat: As I mentioned [here](https://github.com/maidsafe/safe_client_libs/pull/973#pullrequestreview-273212380), I would prefer if we avoid the `map_err` + `panic` pattern.

safe_app/src/tests/append_only_data/published_sequenced.rs
```diff
@@ -0,0 +1,387 @@
...
+    PublicKey, SeqAppendOnly, XorName,
+};
+use std::collections::BTreeMap;
+use std::sync::mpsc;
+use std::thread;
+
+// AD created by app. App lists it's own sign_pk in owners field. Put should fail - Rejected at the client handlers.
+// Should pass when it lists the owner's sign_pk instead.
+#[test]
+fn ad_created_by_app() {
```
m-cat: I think we can save some code duplication and make this a single test, for all adata kinds, in append_only_data/mod.rs. We would just need to add an `Adata::from_kind` function in safe-nd:

```rust
for kind in &[PubSeq, PubUnseq, UnpubSeq, UnpubUnseq] {
    let mut invalid_data = AData::from_kind(kind, name, tag);
    let mut valid_data = invalid_data.clone();
    ...
}
```

I can raise a PR making the necessary changes there.

lionel1704: That'd be great, thanks :+1:

m-cat:
Looking at the code I think we can actually merge all the separate test functions for each type. It would make maintenance of this code much easier, as well as making it easier to review.

We can take the same approach as described in my other comment of constructing an adata of each kind and iterating over them. We can then use `is_pub` and `is_seq` to conditionally call different functions. We just need to add some missing functions like `AData::append_pub_permissions` and `append_unpub_permissions`.

safe_app/src/tests/append_only_data.rs
```diff
@@ -0,0 +1,765 @@
...
+    let app = create_app();
+    let name: XorName = new_rand::random();
+    let tag = 15_002;
+    let data: &[AData; 4] = &[
+        PubSeqAppendOnlyData::new(name, tag).into(),
+        UnpubSeqAppendOnlyData::new(name, tag).into(),
+        PubUnseqAppendOnlyData::new(name, tag).into(),
+        UnpubUnseqAppendOnlyData::new(name, tag).into(),
+    ];
+    for mut invalid_data in data.iter().cloned() {
```
m-cat: Looks like you can avoid a call to `cloned()` by using a `Vec`:

```rust
    let data: Vec<AData> = vec![
        PubSeqAppendOnlyData::new(name, tag).into(),
        UnpubSeqAppendOnlyData::new(name, tag).into(),
        PubUnseqAppendOnlyData::new(name, tag).into(),
        UnpubUnseqAppendOnlyData::new(name, tag).into(),
    ];
    for mut invalid_data in data {
```

Should be slightly more performant and more readable 🙂

safe_app/src/tests/append_only_data.rs
```diff
@@ -0,0 +1,765 @@
...
+                let address = *adata.address();
+                if address.is_pub() {
+                    let mut permissions = BTreeMap::new();
+                    let _ = permissions.insert(
+                        ADataUser::Key(app_pk),
+                        ADataPubPermissionSet::new(true, None),
+                    );
+                    unwrap!(adata.append_pub_permissions(
+                        ADataPubPermissions {
+                            permissions: permissions,
```
m-cat: This line can just be `permissions,`. Found 3 instances of this in the file.

Issue #970, opened 2019-08-08
Make handling of zero-length files "do the right thing"


sprucely:
Currently, opening a file in `APPEND` mode assumes that there is existing data to append to and returns an `ERR_NO_SUCH_DATA` if the file happens to not contain any data. I think it would be more intuitive to have the open mode fall back to `OVERWRITE` in such a case rather than returning that error. Or perhaps it could maintain some other internal state that indicates file_close should result in creation of a new file rather than appending to existing data.

m-cat:
Thanks!

m-cat:
Closed via https://github.com/maidsafe/safe_client_libs/pull/1011

PR #979, opened 2019-08-15
fix/nfs: allow opening empty file in append mode #970

sprucely:
Wraps the Future returned by `data_map::get()` with an or_else Future that will return None when it matches the `NoSuchData` error. This effectively sets the writer to `OPEN_MODE_OVERWRITE` mode, though I believe it wastes a GET to do so. I updated the doc comment on `OPEN_MODE_APPEND` to indicate it is better to use overwrite mode when it is known ahead of time that the file is empty.

### DIFF
--- a/safe_core/src/nfs/writer.rs
+++ b/safe_core/src/nfs/writer.rs
@@ -8,12 +8,14 @@
 
 use crate::client::Client;
 use crate::crypto::shared_secretbox;
-use crate::nfs::{data_map, File, NfsFuture};
+use crate::errors::CoreError;
+use crate::nfs::{data_map, File, NfsError, NfsFuture};
 use crate::self_encryption_storage::SelfEncryptionStorage;
 use crate::utils::FutureExt;
 use chrono::Utc;
 use futures::Future;
-use self_encryption::SequentialEncryptor;
+use routing::ClientError;
+use self_encryption::{DataMap, SequentialEncryptor};
 
 /// Mode of the writer.
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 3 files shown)


safe_core/src/nfs/writer.rs
```diff
@@ -49,9 +51,15 @@ impl<C: Client> Writer<C> {
             Mode::Overwrite => ok!(None),
         };
         let client = client.clone();
-        fut.and_then(move |data_map| {
-            SequentialEncryptor::new(storage, data_map).map_err(From::from)
+        fut.or_else(|err| -> Box<NfsFuture<Option<DataMap>>> {
+            match err {
+                NfsError::CoreError(CoreError::RoutingClientError(ClientError::NoSuchData)) => {
```
m-cat: Worth adding a comment here explaining why we do this?

safe_core/src/nfs/writer.rs
```diff
@@ -49,9 +51,15 @@ impl<C: Client> Writer<C> {
             Mode::Overwrite => ok!(None),
         };
         let client = client.clone();
-        fut.and_then(move |data_map| {
-            SequentialEncryptor::new(storage, data_map).map_err(From::from)
+        fut.or_else(|err| -> Box<NfsFuture<Option<DataMap>>> {
```
m-cat: Is it possible to remove the `-> Box<...>` annotation here or is it required by the compiler? We don't usually include these as they are automatically inferred and are quite verbose if left in.

sprucely: Yeah, I tried, but it was having trouble inferring the type E in the call to the macro `ok!(None)`. I also tried inserting a cast to help the compiler out (`ok!(None as Option<DataMap>)`) but then it complained about it being a trivial cast.

sprucely:
So this was fun to work on. It's my first work with rust Futures. Similar enough to C# Tasks. But between that, coming to terms with usage of Box, nested enums, and the extremely long time it takes RLS to give any intellisense, it took a few hours to get it right.

PR #981, opened 2019-08-15
Refactor the test create balance API

m-cat:
refactor the test create balance API by sending the `CreateBalance` RPC
with the `From` and `To` fields as the same. This will be the special
case that vaults handle to crate a new test balance


safe_core/src/client/mock/vault.rs
```diff
@@ -625,22 +620,30 @@ impl Vault {
...
-                            }
-                            self.create_balance(destination, new_balance_owner)
-                        })
-                        .and_then(|_| {
-                            self.transfer_coins(source, destination, amount, transaction_id)
-                        })
-                };
-                Response::Transaction(result)
+                    let result =
+                        if let Err(e) = self.authorise_coin_operation(&source, requester_pk) {
```
m-cat: @lionel1704 I've already changed this locally, but for future reference: you can use `and_then` instead of `if let { ... } else { .. }` to save a few lines here.

safe_core/src/client/mod.rs
```diff
@@ -2391,20 +2450,23 @@ mod tests {
             let client3 = client.clone();
             let client4 = client.clone();
             let client5 = client.clone();
-            let bls_sk = threshold_crypto::SecretKey::random();
-            let bls_sk2 = bls_sk.clone();
+            // let bls_sk2 = bls_sk.clone();
```
nbaksalyar: This line can be removed

safe_core/src/client/mock/vault.rs
nbaksalyar: Can be just
```suggestion
self.authorise_coin_operation(&source, requester_pk)?;
```

m-cat: Actually I think we don't want to do that, returns from this function should be wrapped inside a `Response` message, even errors.

safe_core/src/client/mod.rs
```diff
@@ -1522,32 +1523,90 @@ pub trait Client: Clone + 'static {
...
+        })
+        .into_box()
+    }
+}
+
+/// Create a new mock balance at an arbitrary address.
+pub fn test_create_balance(
+    owner: &threshold_crypto::SecretKey,
+    amount: Coins,
+) -> Result<(), SndError> {
```
m-cat: Can just use `SndResult` here.

nbaksalyar:
Ignored the AppVeyor build failure because it passed on Jenkins.
Ignored Jenkins Linux failure because it passed on Travis CI.

PR #982, opened 2019-08-16
[Rebased on #983] Automatically read vault connection info

octol:
Reads the connection info generated by a running vault instance.

NOTE: rebased on #983


octol:
Raised as PRs against feature branch upstream instead https://github.com/nbaksalyar/safe_client_libs/pull/2

PR #985, opened 2019-08-19
fix/vault: CreateCoinBalanceFor should return a Transaction response

m-cat:
Fixes #985


safe_app/src/tests/unpublished_mutable_data.rs
```diff
@@ -31,20 +31,23 @@ fn md_created_by_app_1() {
...
         let name: XorName = unwrap!(name_rx.recv());
         let entry_actions = MDataSeqEntryActions::new().ins(vec![1, 2, 3, 4], vec![2, 3, 5], 0);
         let cl2 = client.clone();
         let c3 = client.clone();
         let name2 = name;
+
         client
             .mutate_seq_mdata_entries(name, DIR_TAG, entry_actions)
-            .then(move |res| {
-                unwrap!(res);
```
nbaksalyar: Probably it'd be better to leave it at `then(|res| unwrap!(res))` in this particular case: this pattern is used deliberately in tests because it allows to catch errors early. It prints a line number at the location where an error occurred. With `and_then(...)` it will be printed out only in the end of the combinator chain, not providing a context.

safe_core/src/client/recovery.rs
```diff
@@ -509,16 +509,9 @@ mod tests_with_mock_routing {
 
             client
                 .put_seq_mutable_data(data0)
-                .then(move |res| {
-                    unwrap!(res);
```
nbaksalyar: Same comment here.

nbaksalyar:
To be integrated as a part of #983

PR #991, opened 2019-08-21
Update to safe-nd 0.3.0

lionel1704:
- The `InvalidPermissions` error variant is now deprecated in favour of
`AccessDenied`
- Mapping of the error is no longer required in the mock vault's
check_perms_adata function

For the tests to pass against the local vault maidsafe/safe_vault#813 is required


safe_app/src/tests/append_only_data.rs
```diff
@@ -165,8 +165,7 @@ fn managing_permissions_for_an_app() {
                     match res {
                         // FIXME: we should expect only `AccessDenied` here, but due to
```
m-cat: This `FIXME` comment doesn't make sense anymore.

PR #997, opened 2019-08-21
refactor/client: make `Client::full_id` return `SafeKey`

### LINKED ISSUE #958: Replace `Data` in `mock::vault.rs` with `safe-nd::Data`
Once old mutable data has been phased out, we can remove the temporary `Data` struct in `mock::vault.rs`.

### LINKED ISSUE #994: Implement `compose_message` in the `Client` trait
The implementations of this function in `CoreClient`, `AuthClient`, and `AppClient` are essentially all the same. This can go into a shared implementation in the `Client` trait itself. The bls key should be obtained through the trait method `secret_bls_key`.

### LINKED ISSUE #995: Consider updating `Client::full_id`
https://github.com/maidsafe/safe_client_libs/blob/97baefd38965b7a0bc5cdb62f76611b9df8dd90b/safe_core/src/client/mod.rs#L101-L102
We have this `full_id()` function that returns an old, Routing definition of `FullId`. It'd be nice to see if we really use it anywhere, and if not, change this function to return a `SafeKey` instead.

m-cat:
Also implement `compose_message` as a trait method on `Client`

Fixes #995
Fixes #994
Fixes #958

### DIFF
--- a/safe_app/src/client.rs
+++ b/safe_app/src/client.rs
@@ -10,14 +10,13 @@
 use crate::errors::AppError;
 use crate::{AppContext, AppMsgTx};
 use lru_cache::LruCache;
-use routing::FullId;
 use rust_sodium::crypto::{box_, sign};
-use safe_core::client::{ClientInner, SafeKey, IMMUT_DATA_CACHE_SIZE}; //, REQUEST_TIMEOUT_SECS};
+use safe_core::client::{ClientInner, SafeKey, IMMUT_DATA_CACHE_SIZE};
 use safe_core::config_handler::Config;
 use safe_core::crypto::{shared_box, shared_secretbox, shared_sign};
 use safe_core::ipc::BootstrapConfig;
 use safe_core::{Client, ClientKeys, ConnectionManager, NetworkTx};
-use safe_nd::{AppFullId, Message, MessageId, PublicId, PublicKey, Request, Signature};
+use safe_nd::{AppFullId, PublicId, PublicKey};
 use std::cell::RefCell;
 use std::fmt;
 use std::rc::Rc;
@@ -47,9 +46,7 @@ impl AppClient {
 
         let mut connection_manager =
             ConnectionManager::new(Config::new().quic_p2p, &net_tx.clone())?;
-        block_on_all(connection_manager.bootstrap(SafeKey::app(
-            client_keys.clone().into_app_full_id(client_pk),
-        )))?;
+        block_on_all(connection_manager.bootstrap(client_keys.app_safe_key(client_pk)))?;
 
         // let (routing, routing_rx) = setup_routing(
         //     None,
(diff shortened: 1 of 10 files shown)


safe_core/src/client/account.rs
```diff
@@ -190,15 +189,6 @@ impl Default for ClientKeys {
     }
 }
 
-impl Into<FullId> for ClientKeys {
```
nbaksalyar: I would suggest replacing it with `Into<SafeKey>` instead of removing this conversion.

safe_app/src/client.rs
```diff
@@ -162,11 +161,6 @@ impl AppClient {
 impl Client for AppClient {
     type MsgType = AppContext;
 
-    fn full_id(&self) -> Option<FullId> {
-        let app_inner = self.app_inner.borrow();
-        Some(app_inner.keys.clone()).map(Into::into)
```
nbaksalyar: Then, this code can remain as it is now.
So if we would want to change how identities work in the future, we wouldn't need to change this function.

m-cat: I came up with a bit of a different solution, let me know what you think.

safe_core/src/client/account.rs
m-cat: Good point, I implemented `From` since that is preferred over `Into`, where possible.

safe_app/src/client.rs
```diff
@@ -184,6 +178,11 @@ impl Client for AppClient {
         self.inner.clone()
     }
 
+    fn keys(&self) -> ClientKeys {
```
nbaksalyar: Hmm. I'm not sure we would need to expose the entire set of keys here, especially in the case of `safe_app` which needs to have an owner's PK as a part of its Identity. Otherwise it'd be a bit inconvenient at the point of use: we'd need to do something like this:
```rust
SafeKey::app(AppFullId::with_keys(client.keys().into(), client.owner_key()))
```
instead of just
```rust
client.full_id()
```
I'd suggest leaving it at `full_id(&self) -> SafeKey` and explicitly export it as `SafeKey::app(..)` here.

safe_core/src/client/id.rs
```diff
@@ -21,27 +22,33 @@ pub enum SafeKey {
...
-            SafeKey::Client(client_full_id) => PublicId::Client(client_full_id.public_id().clone()),
+            Self::App(app_full_id) => PublicId::App(app_full_id.public_id().clone()),
+            Self::Client(client_full_id) => PublicId::Client(client_full_id.public_id().clone()),
         }
     }
 }
+
+impl From<ClientKeys> for SafeKey {
+    fn from(keys: ClientKeys) -> Self {
+        Self::client(keys.into())
```
nbaksalyar: This wouldn't be correct for the case of apps which should use the `SafeKey::app` identity.
So probably we should have two explicit functions here instead of relying on `From` / `Into` traits:
```rust
impl ClientKeys {
  fn into_client_safe_key(self) -> SafeKey {
    SafeKey::client(keys.into())
  }

  fn into_app_safe_key(self, owner_key: PublicKey) -> SafeKey {
    SafeKey::app(keys.into(), owner_key)
  }
}
```
Or something like this - pls see if we can do without these functions in this context.

PR #998, opened 2019-08-21
Use tempfile storage in tests

### LINKED ISSUE #996: Investigate why safe_app tests can run slow with on-disk mock vault storage
Sometimes the safe_app tests can run so slow they time out during the entire test suite run.

There are several possible causes for that, but it's worth starting with one of these:
1. The Vault locking logic is inefficient.
2. The storage itself gets overflowed and causes the slowdowns.

octol:
(Maybe) resolves #996

### DIFF
--- a/safe_core/Cargo.toml
+++ b/safe_core/Cargo.toml
@@ -38,10 +38,10 @@ tiny-keccak = "~1.5.0"
 threshold_crypto = "~0.3.2"
 tokio = "~0.1.22"
 unwrap = "~1.2.0"
-tempfile = "3.1.0"
 
 [dev-dependencies]
 serde_json = "~1.0.9"
+tempfile = "3.1.0"
 
 [features]
 mock-network = []
--- a/safe_core/Cargo.toml
+++ b/safe_core/Cargo.toml
@@ -38,6 +38,7 @@ tiny-keccak = "~1.5.0"
 threshold_crypto = "~0.3.2"
 tokio = "~0.1.22"
 unwrap = "~1.2.0"
+tempfile = "3.1.0"
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -62,9 +63,12 @@ fn init_vault_path(devconfig: Option<&DevConfig>) -> PathBuf {
 }
 
 // Initializes vault storage. The type of storage is chosen with the following precedence:
-// 1. "SAFE_MOCK_IN_MEMORY_STORAGE" env var => in-memory storage
-// 2. DevConfig `mock_in_memory_storage` option => in-memory storage
-// 3. Else => file storage, use path from `init_vault_path`
+// 1.  "SAFE_MOCK_IN_MEMORY_STORAGE" env var => in-memory storage
+// 2.  DevConfig `mock_in_memory_storage` option => in-memory storage
+// Then either (non-tests)
```
m-cat: This "(non-tests)" comment is confusing, as the 3b case is for tests.

Maybe instead write "Else (non-test) => ..." and "Else (test) => ..."

octol: Updated!

safe_core/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tiny-keccak = "~1.5.0"
 threshold_crypto = "~0.3.2"
 tokio = "~0.1.22"
 unwrap = "~1.2.0"
+tempfile = "3.1.0"
```
nbaksalyar: I wonder if we can leave this in `[dev-depdendencies]` since we use it only for tests?

octol: Ah yes good point, I'll move it

PR #999, opened 2019-08-22
Refactor Jenkins build to use mock vault file on OSX

lionel1704:
With AppVeyor now deprecated, we need to consider other options to run tests
against the mock vault file. For this purpose, on OS X the test suite can be
run withouth enabling SAFE_MOCK_IN_MEMORY_STORAGE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,8 +220,15 @@ ifeq ($(UNAME_S),Linux)
 else
 	./scripts/test-mock
 endif
-	mkdir artifacts
-	find target/release -maxdepth 1 -type f -exec cp '{}' artifacts \;
+	make copy-artifacts
+
+tests-mock-file: clean
+	rm -rf artifacts
+	cargo test --verbose --release --features=mock-network --manifest-path=safe_core/Cargo.toml
+	cargo test --verbose --release --features=mock-network --manifest-path=safe_authenticator/Cargo.toml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -220,8 +220,15 @@ ifeq ($(UNAME_S),Linux)
 else
 	./scripts/test-mock
 endif
-	mkdir artifacts
-	find target/release -maxdepth 1 -type f -exec cp '{}' artifacts \;
+	make copy-artifacts
+
+tests-mock-file: clean
+	rm -rf artifacts
+	cargo test --verbose --release --features=mock-network --manifest-path=safe_core/Cargo.toml
```
m-cat: Can these `cargo test` instructions be moved to their own file in `scripts/`? It can be called something like `test-mock-file`. See `scripts/test-mock`.

Makefile
```diff
@@ -220,8 +220,15 @@ ifeq ($(UNAME_S),Linux)
 else
 	./scripts/test-mock
 endif
-	mkdir artifacts
-	find target/release -maxdepth 1 -type f -exec cp '{}' artifacts \;
+	make copy-artifacts
+
+tests-mock-file: clean
```
m-cat: Maybe worth adding a comment that we only run this target on OSX, which is why we don't need Linux-specific instructions like we have in `tests`?

PR #1000, opened 2019-08-22
Refactor Jenkins to use mock file on OSX and lock file only for mutations

### LINKED ISSUE #996: Investigate why safe_app tests can run slow with on-disk mock vault storage

lionel1704:
Hey guys, 
I wanted to close and re-open #999 to re-trigger the build since GH is having issues, but I'm unable to re-open it now. So here's a new one.

closes #996

### DIFF
--- a/safe_core/src/client/mock/connection_manager.rs
+++ b/safe_core/src/client/mock/connection_manager.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::vault::{self, Vault};
+use super::vault::{self, Vault, RequestType, request_is_get};
 use crate::config_handler::{get_config, Config};
 use crate::{
     client::SafeKey,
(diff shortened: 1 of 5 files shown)


safe_core/src/client/mock/connection_manager.rs
```diff
@@ -63,7 +63,7 @@ impl ConnectionManager {
     /// Send `message` via the `ConnectionGroup` specified by our given `pub_id`.
     pub fn send(&mut self, pub_id: &PublicId, msg: &Message) -> Box<CoreFuture<Response>> {
         let msg: Message = {
-            let mut vault = vault::lock(&self.vault, true);
+            let mut vault = vault::lock(&self.vault, is_read(msg));
```
m-cat: Correct me if I'm wrong but shouldn't this be `!is_read` (as this is the `writing` parameter)?

safe_core/src/client/mock/connection_manager.rs
```diff
@@ -140,6 +140,61 @@ impl ConnectionManager {
     }
 }
 
+fn is_read(msg: &Message) -> bool {
```
m-cat: Possible to reuse `request_is_get()` instead of introducing a new, similar function?

lionel1704:
You're right @m-cat !
I'll refactor this :)

m-cat:
Superseded by https://github.com/maidsafe/safe_client_libs/pull/1006.

PR #1001, opened 2019-08-22
feat/hooks: fixes hooks functionality

### LINKED ISSUE #972: Update revocation tests
We already updated the [revocation tests](https://github.com/maidsafe/safe_client_libs/blob/master/safe_authenticator/src/tests/revocation.rs) after https://github.com/maidsafe/safe_client_libs/pull/948, but there are still some vestiges of our old approach (e.g. checking for re-encryption by fetching files after revocation, and comments to that effect (search "re-encryption" in the file)). We also have a lot of commented-out code, e.g. in `app_revocation_recovery`. This can be cleaned up.

Yoga07:
Also partially resolves #972

### DIFF
--- a/safe_app/src/tests/mod.rs
+++ b/safe_app/src/tests/mod.rs
@@ -29,8 +29,8 @@ use safe_core::ConnectionManager;
 use safe_core::{Client, CoreError};
 use safe_nd::{
     ADataAddress, ADataOwner, AppPermissions, AppendOnlyData, Coins, Error as SndError,
-    PubImmutableData, PubSeqAppendOnlyData, PubUnseqAppendOnlyData, UnpubUnseqAppendOnlyData,
-    XorName,
+    PubImmutableData, PubSeqAppendOnlyData, PubUnseqAppendOnlyData, Request, Response,
+    UnpubUnseqAppendOnlyData, XorName,
 };
 use std::collections::HashMap;
 use std::rc::Rc;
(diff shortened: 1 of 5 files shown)


safe_app/src/tests/mod.rs
```diff
@@ -108,52 +108,31 @@ fn get_access_info() {
...
-                }),
-                // Request::InsAuthKey { msg_id, .. } => Some(Response::InsAuthKey {
-                //     res: Err(ClientError::LowBalance),
-                //     msg_id,
-                // }),
-                // Request::DelAuthKey { msg_id, .. } => Some(Response::DelAuthKey {
-                //     res: Err(ClientError::LowBalance),
-                //     msg_id,
-                // }),
+                // Immutable Data
```
m-cat: I think you can replace this code with the `request_is_get` function (rename it to `request_get_type`?) and test if the returned `RequestType` is a `Mutation`.

safe_core/src/client/mock/connection_manager.rs
```diff
@@ -142,35 +151,16 @@ impl ConnectionManager {
...
-            } else {
-                None
-            };
-
-            if let Some(response) = response {
-                self.send_response(delay_ms, src, dst, response);
-                return true;
+    fn intercept_request(&mut self, message: Message) -> Option<Response> {
+        if let Message::Request { request, .. } = message {
+            if let Some(hook) = Arc::get_mut(self.request_hook.as_mut()?) {
```
m-cat: Just wondering, do we need to call `Arc::get_mut`? We don't do it in the previous, commented-out code here, which was a big simpler. And I don't think we need a *mutable* reference to `hook`.

safe_core/src/client/mock/connection_manager.rs
```diff
@@ -62,6 +62,15 @@ impl ConnectionManager {
 
     /// Send `message` via the `ConnectionGroup` specified by our given `pub_id`.
     pub fn send(&mut self, pub_id: &PublicId, msg: &Message) -> Box<CoreFuture<Response>> {
+        #[cfg(any(feature = "testing", test))]
+        {
+            if self.request_hook.is_some() {
```
m-cat: You can omit this first check, since `intercept_request` will already return `None` if the request hook is not set.

safe_authenticator/src/tests/revocation.rs
```diff
@@ -335,9 +312,8 @@ mod mock_routing {
 
         // Attempt to re-authenticate the app fails, because revocation is pending.
         match register_app(&auth, &auth_req) {
-            // Hooks are disabled
-            Ok(_) => (), // TODO: assert expected error variant
-            x => panic!("Unexpected {:?}", x),
+            Err(_) => (),
```
m-cat: Can we match on the exact error variant that is expected here?

safe_authenticator/src/tests/revocation.rs
```diff
@@ -225,46 +227,21 @@ mod mock_routing {
...
+        let auth = unwrap!(Authenticator::login(
+            locator.clone(),
+            password.clone(),
+            || (),
+        ));
 
         // Revoke the app.
         match try_revoke(&auth, &app_id) {
-            // This will succeed since there are no hooks
             Ok(()) => (),
```
m-cat: Isn't this supposed to fail now that we reintroduced hooks?

On the other hand, if this is correct, then this `match` statement can be replaced with `unwrap!`.

safe_authenticator/src/tests/revocation.rs
```diff
@@ -225,46 +227,21 @@ mod mock_routing {
...
-        //            });
-        //            routing
-        //        };
-        //        let auth = unwrap!(Authenticator::login_with_hook(
-        //            locator.clone(),
-        //            password.clone(),
-        //            || (),
-        //            routing_hook,
-        //        ));
+        let auth = unwrap!(Authenticator::login(
```
m-cat: Why not re-instate the hooks here. Isn't this test supposed to simulate a network failure during re-encryption? Maybe update the comments above this test?

safe_authenticator/src/tests/mod.rs
```diff
@@ -153,43 +143,18 @@ mod mock_routing {
...
-                //         })
-                //     }
-                //     Request::ChangeMDataOwner { msg_id, .. } => Some(Response::ChangeMDataOwner {
-                //         res: Err(SndError::InsufficientBalance),
-                //         msg_id,
-                //     }),
-                //     // Pass-through
-                //     _ => None,
-                // }
+            cm.set_request_hook(move |req| {
```
m-cat: If I remember correctly, the purpose of this test is to ensure that login can happen despite there being insufficient funds for any mutation. So I think to ensure that all mutations are caught, it's important to use the `request_is_get` function here (which I think should be renamed).

safe_app/src/tests/mod.rs
nbaksalyar: There's also [`Request::error_response`](https://docs.rs/safe-nd/0.3.1/safe_nd/enum.Request.html#method.error_response) which might be useful in this case :+1:

safe_authenticator/src/tests/revocation.rs
nbaksalyar: We don't have re-encryption anymore, so this part should be fine :+1:

m-cat: Ah yeah. So the comments should be updated.

m-cat:
Superseded by https://github.com/maidsafe/safe_client_libs/pull/1007.

PR #1004, opened 2019-08-23
Use a unique quic-p2p for each ConnectionGroup

### LINKED ISSUE #993: Use a unique quic-p2p for each ConnectionGroup
Currently, we assume that we can use the same `ConnectionManager` for multiple identities, which is not the case due to the way Phase 1 Vaults work:

1. When we start a new connection to a vault, we send a challenge response which authenticates this particular connection, mapping it to a user/identity A.
2. It's good for the simple cases, but now imagine you need to do something as a user/identity B. It would just work if it was managed by a vault group corresponding to identity B, but since we don't have sections until after Phase 3, we need to connect to the same Vault for both identities.
3. Since we use the same quic-p2p instance for `ConnectionManager` because of efficiency reasons (as a single quic-p2p instance encapsulates a lot of things like the event loop within its own thread, a udp socket, and some more), we need to first disconnect the identity A, because otherwise the Vault wouldn't care if you use identity B: it knows you as a user A since you've already sent your challenge response.
4. Because of this, the API and the code we write becomes more convoluted: a developer needs to be aware of the fact and needs to reconnect at the right times. However, it _shouldn't_ be the case because it is a Phase 1 implementation detail.

So far we needed to implemented a 'hack' for functions such as [`send_as`](https://github.com/maidsafe/safe_client_libs/blob/f851b9ae1be9acaf7e35aabe8e4410bbf4c1fcec/safe_core/src/client/mod.rs#L1344), which requires reconnecting to the Vault multiple times, e.g.:

```rust
fn send_as(.., secret_key: Option<&BlsSecretKey>) -> Box<CoreFuture<Response>> {
  let opt_full_id = <..>; // if `secret_key` is Some(..), create a SafeKey::client(ClientFullId::with_bls_key(secret_key))
let cm = connection_manager.clone();

// .. construct a message - this code doesn't need to be changed

if connection_manager.has_connection(&opt_full_id) {
  Box::new(connection_manager.send(&opt_full_id, &message))
} else {
  connection_manager.disconnect(&client.full_id())
    .and_then(move |_| cm.bootstrap(&opt_full_id))
    .and_then(move |_| cm.send(&opt_full_id, &message))
    .and_then(move |_| cm.disconnect(&opt_full_id)
    .and_then(move |_| cm.bootstrap(&client.full_id()))
}
```

But we can solve this problem in a different way if we used a new connection to a Vault for identity B, because in this case the Vault would have sent a new challenge request. Hence, we should use a unique `quic_p2p` object per each [`ConnectionGroup`](https://github.com/maidsafe/safe_client_libs/blob/f851b9ae1be9acaf7e35aabe8e4410bbf4c1fcec/safe_core/src/connection_manager/connection_group.rs#L62), encapsulating the challenge handling and the rest of the things inside of a group (representing a connection to a group of vaults handling an identity X).

nbaksalyar:
Closes #993.

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -45,8 +45,8 @@ use rust_sodium::crypto::{box_, sign};
 use safe_nd::{
     AData, ADataAddress, ADataAppendOperation, ADataEntries, ADataEntry, ADataIndex, ADataIndices,
     ADataOwner, ADataPermissions, ADataPubPermissionSet, ADataPubPermissions,
-    ADataUnpubPermissionSet, ADataUnpubPermissions, ADataUser, AppPermissions, ClientFullId, Coins,
-    IData, IDataAddress, LoginPacket, MData, MDataAddress, MDataEntries, MDataEntryActions,
+    ADataUnpubPermissionSet, ADataUnpubPermissions, ADataUser, AppPermissions, Coins, IData,
+    IDataAddress, LoginPacket, MData, MDataAddress, MDataEntries, MDataEntryActions,
     MDataPermissionSet, MDataSeqEntries, MDataSeqEntryActions, MDataSeqValue,
     MDataUnseqEntryActions, MDataValue, MDataValues, Message, MessageId, PublicId, PublicKey,
     Request, Response, SeqMutableData, Signature, Transaction, UnseqMutableData, XorName,
@@ -1091,7 +1091,7 @@ pub trait Client: Clone + 'static {
     }
 
     /// Set the coin balance to a specific value for testing
-    #[cfg(any(test, all(feature = "testing", feature = "mock-network")))]
+    #[cfg(any(test, feature = "testing"))]
     fn test_set_balance(
         &self,
         secret_key: Option<&BlsSecretKey>,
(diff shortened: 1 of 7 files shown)


safe_core/src/client/mod.rs
```diff
@@ -1362,18 +1362,25 @@ fn send_as(
     request: Request,
     secret_key: Option<&BlsSecretKey>,
 ) -> Box<CoreFuture<Response>> {
-    let (message, requester) = match secret_key {
+    let (message, identity) = match secret_key {
         Some(key) => (
             sign_request_with_key(request, key),
-            PublicId::Client(ClientFullId::with_bls_key(key.clone()).public_id().clone()),
+            SafeKey::client(ClientFullId::with_bls_key(key.clone())),
```
m-cat: You can use the new client_from_bls_key method here.

octol: Outside the scope of the PR, but is it potential source of confusion to use the name `SafeKey`? Isn't that the new name for `CoinBalance`? https://github.com/maidsafe/safe-cli/pull/209

nbaksalyar: >Isn't that the new name for CoinBalance?

Hmm, technically this _is_ the `CoinBalance` though, only with the full set of keys (public + private). Or do you think another name would be a better suit here?

octol: Ah I see, ok then it's probably ok :)

PR #1007, opened 2019-08-23
feat/hooks: fixes hooks functionality 

### LINKED ISSUE #972: Update revocation tests

m-cat:
Also partially resolves #972

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1869,7 +1869,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "safe-nd"
 version = "0.3.1"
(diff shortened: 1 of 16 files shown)


m-cat:
Will re-raise this once https://github.com/maidsafe/safe_client_libs/pull/1009 is merged.

PR #1009, opened 2019-08-27
chore/deps: update to safe-nd 0.3.2

m-cat:

### DIFF
--- a/safe_core/src/client/mock/connection_manager.rs
+++ b/safe_core/src/client/mock/connection_manager.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::vault::{self, Vault};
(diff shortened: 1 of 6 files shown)


safe_core/src/client/mock/connection_manager.rs
```diff
@@ -63,7 +63,18 @@ impl ConnectionManager {
     /// Send `message` via the `ConnectionGroup` specified by our given `pub_id`.
     pub fn send(&mut self, pub_id: &PublicId, msg: &Message) -> Box<CoreFuture<Response>> {
         let msg: Message = {
-            let mut vault = vault::lock(&self.vault, true);
+            let writing = {
+                if let Message::Request { request, .. } = msg {
```
nbaksalyar: Maybe try it with `match`?
```rust
let writing = match msg {
  Message::Request { ref request, .. } if request.get_type() == RequestType::Mutation => true,
  _ => false,
};
```

PR #1010, opened 2019-08-27
 refactor/ffi: rework MDataKey and MDataValue to use FFI conventions 

### LINKED ISSUE #770: FFI structs may leak memory
There may be memory leaks in our client_libs ffi:
https://github.com/maidsafe/safe_client_libs/blob/master/safe_core/src/ffi/ipc/resp.rs#L148

~The strings in this struct are never freed (see Drop implementations for the other structs in this file).~ [Fixed in #758]

MDataKey.val and MDataValue.content (in the same file) also need to be handled as the pointed-to data is never freed.

Fix these issues and check that all FFI structs in safe_client_libs have a correct Drop defined.

Notes:

* ~We need to consider the caveats about implementing Drop and whether they apply: https://doc.rust-lang.org/1.16.0/nomicon/dropck.html~ Checked, does not apply
* ~We need to make sure that we actually run drop. For example, AuthGranted::clone_from_repr_c takes ownership of a pointer to ffi::AuthGranted and makes a clone, but never explicitly drops the pointed-to struct. Is this a memory leak? Write a short test to check this.~ Checked, see comment below

m-cat:
Also:

+ make miscellaneous doc fixes
+ clean up FFI documentation

Closes #770


safe_core/src/client/mock/routing.rs
```diff
@@ -6,8 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-/// Helper macro to receive a routing event and assert it's a response
-/// success.
+/// Helper macro to receive a routing event and assert it's a response success.
 #[macro_export]
```
nbaksalyar: Hmm, I wonder where this file comes from.
It should've been removed, but seemingly it was just forgotten. Could you please remove this module completely? We no longer need it.

safe_core/src/client/mock/tests.rs
```diff
@@ -32,6 +32,26 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use threshold_crypto::SecretKey;
 
+// Helper macro to receive a routing event and assert it's a response success.
+macro_rules! expect_success {
```
nbaksalyar: Hmm, do we still use this macro?
I'd expect it to be removed completely at this point (as it relates to Routing), so probably we should do just that and leave this module as it is.

Same applies to `expect_failure`, but it can be changed later.

safe_core/src/ffi/ipc/resp.rs
```diff
@@ -98,7 +98,7 @@ pub struct AuthGranted {
     pub access_container_entry: AccessContainerEntry,
 
     /// Crust's bootstrap config
```
nbaksalyar: While we're at it, could you please change this comment to 
```suggestion
     /// QuicP2P bootstrap config
```

safe_core/src/client/mock/tests.rs
m-cat: Yes, we use these two macros throughout `mock/tests.rs`.

nbaksalyar: `mock/tests` need to be reworked though because it depends on the old mock routing system. So technically they're used, but since that module is broken anyway, I think we can just ignore it at this moment.

PR #1011, opened 2019-08-27
 fix/nfs: allow opening empty file in append mode #970 

m-cat:
Also includes commits that were authored by @sprucely and previously merged, but seem to have been overwritten.


m-cat:
My bad, those commits were not overwritten, `master` branch was renamed to `stable`. Removing the commits from this PR.

PR #1013, opened 2019-08-28
Some refactors and fixes in the mock vault with relevant tests

lionel1704:

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -382,16 +382,18 @@ impl Vault {
     }
 
     #[allow(clippy::cognitive_complexity)]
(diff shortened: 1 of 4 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -949,7 +956,6 @@ impl Vault {
                         AData::PubSeq(_) | AData::PubUnseq(_) => Err(SndError::InvalidOperation),
                         AData::UnpubSeq(_) | AData::UnpubUnseq(_) => {
                             self.delete_data(id);
-                            self.commit_mutation(requester.name());
```
nbaksalyar: Just a side note (not related to this PR): probably in future it'd be better to rename `commit_mutation` to `pay_for_mutation` or even just `pay` (as in Vaults). Otherwise it's not very clear what it's doing (I assumed it's like 'commit a transaction', which I believe it did before).

PR #1015, opened 2019-08-28
Use the bootstrap config from the auth response and add a cost for create balance

lionel1704:
apps should connect using the bootstrap config from the auth response to
bootstrap to the network

### DIFF
--- a/safe_app/src/client.rs
+++ b/safe_app/src/client.rs
@@ -127,7 +127,11 @@ impl AppClient {
         trace!("Attempting to log into an acc using client keys.");
 
         let mut qp2p_config = Config::new().quic_p2p;
-        qp2p_config.hard_coded_contacts = qp2p_config.hard_coded_contacts.union(&config).cloned().collect();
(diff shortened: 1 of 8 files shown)


safe_app/src/client.rs
```diff
@@ -44,16 +44,15 @@ impl AppClient {
         let client_keys = ClientKeys::new(None);
         let client_pk = PublicKey::from(client_keys.bls_pk);
 
-        let mut connection_manager =
-            ConnectionManager::new(Config::new().quic_p2p, &net_tx.clone())?;
-        block_on_all(connection_manager.bootstrap(client_keys.app_safe_key(client_pk)))?;
+        let mut qp2p_config = Config::new().quic_p2p;
+        if let Some(additional_contacts) = config.clone() {
+            additional_contacts.iter().cloned().for_each(|contact| {
```
nbaksalyar: I think this can be replaced with
```rust
qp2p_config.hard_coded_contacts = qp2p_config.hard_coded_contacts.union(&additional_contacts).collect();
```

PR #1016, opened 2019-08-29
Make some required modules public

lionel1704:


safe_authenticator/src/lib.rs
```diff
@@ -85,13 +85,13 @@ pub use ffi::ipc::*;
 pub use ffi::logging::*;
 pub use ffi::*;
 
-mod access_container;
-mod app_auth;
-mod app_container;
+pub mod access_container;
```
m-cat: Can these pub modules be grouped with the other pub modules above?

Issue #1008, opened 2019-08-26
Fix safe_app_jni and safe_authenticator_jni builds


nbaksalyar:
1. Replace the `routing` dependency with `safe-nd`.
2. See if we need to update `syn` in safe-bindgen to support the latest syntax.
3. Remove unnecessary structs if needed (e.g., `safe_core::ffi::AccountInfo` is not being used anywhere).

PR #1023, opened 2019-08-29
fix/jni: fix safe_app and safe_auth jni builds

m-cat:
Fixes https://github.com/maidsafe/safe_client_libs/issues/1008


nbaksalyar:
Thanks!

PR #1027, opened 2019-09-04
refactor/ffi: replace OPEN_MODE_* constants with FFI enum

### LINKED ISSUE #967: Remove `OPEN_MODE_*` constants in favor of an enum
We have manually-enumerated `OPEN_MODE_*` constants in `ffi/nfs.rs` that we perform bitwise operations on, for example:

```rust
            let writer = if open_mode & (OPEN_MODE_OVERWRITE | OPEN_MODE_APPEND) != 0 {
                let writer_mode = if open_mode & OPEN_MODE_APPEND != 0 {
```

This is not idiomatic Rust: we should be using `match` and `if let` instead.

We should change these constants into an FFI-compatible enum:

```rust
#[repr(C)]
pub enum OpenMode {
    Overwrite = 1,
    Append = 2,
    Read = 4,
}
```

While we're at it, we should explicitly define integer values for the `Repr(C)` `Permission` enum in `safe_core::ipc::req::mod.rs`, enabling C users to perform bitwise operations on its variants:

```rust
/// Permission enum - use for internal storage only.
#[repr(C)]
#[derive(Copy, Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
pub enum Permission {
    /// Read.
    Read = 1,
    /// Insert.
    Insert = 2,
    /// Update.
    Update = 4,
    /// Delete.
    Delete = 8,
    /// Modify permissions.
    ManagePermissions = 16,
}
```

m-cat:
Fixes #967 

Note that this doesn't remove the bitwise operations as mentioned in the issue -- I think this isn't possible as the FFI function must take a `u64`, to support the user passing in multiple modes via the `|` operator. However, the enum still ended up being cleaner than separate constants.

When we refactor out the API to have Rust-native equivalents for all current FFI functions (and the FFI functions become merely wrappers around those), we can make a Rust-native equivalent to the enum which looks like:

```rust
pub struct OpenMode {
    pub overwrite: bool,
    pub append: bool,
    pub read: bool,
}
```

which implements the `ReprC` trait to provide conversions to and from the FFI enum.

We should take our time thinking about this since we will need a strategy for FFI enums going forward, i.e. in safe-nd. Alternatives to the strategy in this PR include using crates such as [bitflags](https://docs.rs/bitflags/1.1.0/bitflags/) or [enum-primitive](https://andersk.github.io/enum_primitive-rs/enum_primitive/) but they are probably not worth the extra dependency as there is not much boilerplate here. The downsides to the current strategy include many `as u64` conversions and the continued use of bitwise operations, but these will go away once we can convert to the Rust-native equivalent and use more idiomatic operations.

### DIFF
--- a/safe_app/src/ffi/nfs.rs
+++ b/safe_app/src/ffi/nfs.rs
@@ -22,7 +22,7 @@ use safe_core::ffi::nfs::File;
 use safe_core::ffi::MDataInfo;
 use safe_core::nfs::file_helper::{self, Version};
 use safe_core::nfs::File as NativeFile;
-use safe_core::nfs::{Mode, Reader, Writer};
+use safe_core::nfs::{Reader, Writer, WriterMode};
 use safe_core::{FutureExt, MDataInfo as NativeMDataInfo};
 use std::os::raw::{c_char, c_void};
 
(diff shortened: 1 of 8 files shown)


ravinderjangra:
The PR is no longer relevant as the repo doesn't contain any FFI APIs.

Issue #990, opened 2019-08-20
Add `get_balance` to AppPermissions


nbaksalyar:
Currently, the `get_balance` operation is gated on the `transfer_coins` app permission. We should make it more granular by allowing to look up an owner's balance without necessarily allowing to transfer coins.

lionel1704:
This issue has been resolved in the PRs linked above.

PR #1034, opened 2019-09-12
refactor/permissions: refactor permissions handling in Mock Vault to the new granularized AppPermissions

### LINKED ISSUE #989: Add `allow_mutations` to AppsPermissions
Currently, apps are disallowed to perform any mutation operations if they don't have `transfer_coins` enabled. We should separate it into its own permission type, `allow_mutations`.

Yoga07:
Resolves #989 and #990

### DIFF
--- a/safe_core/src/client/mock/vault.rs
+++ b/safe_core/src/client/mock/vault.rs
@@ -8,7 +8,7 @@
 
 use super::DataId;
 use super::{Account, CoinBalance};
-use crate::client::mock::connection_manager::unlimited_muts;
+use crate::client::mock::connection_manager::unlimited_coins;
 use crate::client::COST_OF_PUT;
 use crate::config_handler::{Config, DevConfig};
 use fs2::FileExt;
@@ -28,6 +28,7 @@ use std::fs::OpenOptions;
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::ops::{Deref, DerefMut};
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::{Mutex, MutexGuard};
 use std::time::Duration;
 use std::time::SystemTime;
@@ -149,6 +150,12 @@ fn check_perms_mdata(data: &MData, request: &Request, requester: PublicKey) -> S
     }
 }
 
+enum Operation {
+    TransferCoins,
+    Mutation,
+    GetBalance,
+}
+
 impl Vault {
     pub fn new(config: Config) -> Self {
         let store = init_vault_store(&config);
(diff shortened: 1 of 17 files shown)


safe_core/src/client/mock/vault.rs
```diff
@@ -374,10 +375,38 @@ impl Vault {
         })
     }
 
-    fn get_balance(&self, coins_balance_id: &XorName) -> SndResult<Coins> {
-        match self.get_coin_balance(coins_balance_id) {
-            Some(balance) => Ok(balance.balance()),
-            None => Err(SndError::NoSuchBalance),
+    fn get_balance(&self, coins_balance_id: &XorName, requester_pk: PublicKey) -> SndResult<Coins> {
```
lionel1704: This function shares some logic with the `authorise_mutation`.
It might be worth adding something like an `Operations` enum with `GetBalance, TransferCoins and Mutation` and a common function to authorise the operation depending on the request

safe_core/src/client/mock/vault.rs
```diff
@@ -228,89 +238,124 @@ impl Vault {
...
+            || {
+                debug!("Coin balance {:?} not found", coins_balance_id);
+                Err(SndError::NoSuchBalance)
+            },
+            |bal| Ok(bal.balance()),
+        )
+    }
+
+    // Authorises coin transfers, mutations and get balance operations.
+    pub fn authorise_operation(
```
lionel1704: This can be improved a bit.
Will discuss offline.

safe_core/src/client/mock/vault.rs
```diff
@@ -228,88 +234,87 @@ impl Vault {
...
+            || {
+                debug!("Account not found for {:?}", owner);
+                Err(SndError::AccessDenied)
+            },
+            Ok,
+        )?;
+        // Fetches permissions granted to the application
+        let perms = account.auth_keys().get(&requester_pk).map_or_else(
+            || {
+                debug!("Account not found for {:?}", owner);
```
lionel1704: This `debug!` trace should say something like "App not authorised"

safe_core/src/client/mock/vault.rs
```diff
@@ -228,88 +234,87 @@ impl Vault {
...
+        // Fetches the account of the owner
+        let account: &Account = self.get_account(&owner).map_or_else(
+            || {
+                debug!("Account not found for {:?}", owner);
+                Err(SndError::AccessDenied)
+            },
+            Ok,
+        )?;
+        // Fetches permissions granted to the application
+        let perms = account.auth_keys().get(&requester_pk).map_or_else(
```
lionel1704: Do you think using `ok_or()` is a better option here?

safe_core/src/client/mock/vault.rs
```diff
@@ -228,88 +234,87 @@ impl Vault {
...
-                            return Err(SndError::AccessDenied);
-                        }
-                    } else {
-                        trace!("App not authorised");
+                Operation::Mutation => {
+                    if !perms.perform_mutations {
+                        debug!("Performing mutations not authorised");
                         return Err(SndError::AccessDenied);
                     }
+                    let unlimited_mut = unlimited_muts(&self.config);
```
lionel1704: I've an idea.
Can we add an `unlimited_coins: bool` to the dev config? It would be useful for app devs.
This way we can also add a `has_sufficient_balance(op: Operation, balance: Coins) -> bool` function and use it whenever this check is required.

safe_core/src/client/mock/vault.rs
```diff
@@ -550,32 +548,38 @@ impl Vault {
...
-                                })
-                                .and_then(|()| {
-                                    self.commit_mutation(&source);
-                                    self.transfer_coins(source, destination, amount, transaction_id)
-                                })
-                        })
+                    self.authorise_operation(
+                        &[
+                            Operation::Mutation,
+                            Operation::GetBalance,
```
lionel1704: A couple of changes required here:
- An app does not need `GetBalance` permissions to create a balance.
- TransferCoins permissions is required only if amount > 0

safe_core/src/client/mock/vault.rs
```diff
@@ -1308,12 +1321,16 @@ impl Vault {
         requester: PublicId,
     ) -> SndResult<()> {
         match requester.clone() {
-            PublicId::Client(client_public_id) => {
-                self.authorise_mutation(client_public_id.name(), client_public_id.public_key())?
-            }
-            PublicId::App(app_public_id) => {
-                self.authorise_mutation(app_public_id.owner_name(), app_public_id.public_key())?
-            }
+            PublicId::Client(client_public_id) => self.authorise_operation(
```
m-cat: My previous comment here still applies, I think you can refactor out `authorise_operation` so it only gets called once. Just return the name and key from these two branches.

safe_core/src/client/mock/vault.rs
m-cat: Looks like you fixed one case of this but missed this one 🙂

safe_core/src/client/mock/vault.rs
```diff
@@ -588,17 +592,22 @@ impl Vault {
...
+                let result = if self
                     .get_login_packet(new_login_packet.destination())
-                    .is_some()
-                {
+                    .is_some() {
                     Err(SndError::LoginPacketExists)
-                } else {
+                } else if let Err(e) =
+                // Check if the requester is authorized to perform coin transactions, mutate, and read balance.
+                    if amount == unwrap!(Coins::from_str("0")) {
```
lionel1704: Same comment as above

safe_core/src/client/mock/vault.rs
```diff
@@ -640,10 +649,18 @@ impl Vault {
             Request::CreateLoginPacket(account_data) => {
                 let source = owner_pk.into();
 
-                if let Err(e) = self.authorise_coin_operation(&source, requester_pk) {
-                    Response::Mutation(Err(e))
-                } else if self.get_login_packet(account_data.destination()).is_some() {
+                if self.get_login_packet(account_data.destination()).is_some() {
```
lionel1704: This needs to be coherent with safe_vault.
In the real network, the operation will be authorised first (at my client handlers) and then passed on to a different section which is responsible for the login packet storage.
So depending on which check is performed first a different error would be returned in tests. So we should test this against a vault too.

safe_core/src/client/mock/vault.rs
```diff
@@ -360,8 +358,13 @@ impl Vault {
         amount: Coins,
         transaction_id: u64,
     ) -> SndResult<Transaction> {
+        let ul = unlimited_coins(&self.config);
```
m-cat: Why store this in a variable if it's only used once? Other than that, looks good!

Yoga07: `self` is being borrowed mutablly in the next line(L362), and it doesn't allow us to use `self.config`while still being borrowed. Hence storing the result before hand.

safe_core/src/ipc/req/auth.rs
```diff
@@ -44,6 +44,8 @@ impl AuthReq {
             app: app.into_repr_c()?,
             app_container,
             app_permission_transfer_coins: app_permissions.transfer_coins,
+            app_permission_perform_mutations: app_permissions.perform_mutations,
```
nbaksalyar: I'm thinking that it would probably be better to have a separate FFI struct for `AppPermissions` instead of relying on prefixes? I.e. we can replace
```rust
app_permission_transfer_coins: app_permissions.transfer_coins,
app_permission_perform_mutations: app_permissions.perform_mutations,
...
```
with
```rust
app_permissions: app_permissions.into_repr_c(),
```

And because `AppPermissions` itself is just a primitive struct with some booleans, perhaps it can be just made `#[repr(C)]` right in `safe-nd`?

Yoga07: Sure @nbaksalyar ! Sounds like a good idea. We shall do this in the upcoming PRs.

safe_core/src/client/mock/vault.rs
```diff
@@ -640,21 +634,26 @@ impl Vault {
             Request::CreateLoginPacket(account_data) => {
                 let source = owner_pk.into();
 
-                if let Err(e) = self.authorise_coin_operation(&source, requester_pk) {
+                if let Err(e) = self.authorise_operations(
+                    &[
+                        Operation::Mutation,
+                        Operation::GetBalance,
+                        Operation::TransferCoins,
```
nbaksalyar: Adding as a comment here too: as was discussed on Slack, this is not what is expected for the `CreateLoginPacket` operation, it should require only the `Operation::Mutation` permissions.

safe_core/src/client/mock/vault.rs
```diff
@@ -640,21 +634,20 @@ impl Vault {
...
+                {
                     Response::Mutation(Err(e))
                 } else if self.get_login_packet(account_data.destination()).is_some() {
                     Response::Mutation(Err(SndError::LoginPacketExists))
                 } else {
                     let result = self
                         .get_balance(&source)
                         .and_then(|source_balance| {
-                            if source_balance.checked_sub(*COST_OF_PUT).is_none() {
+                            if !self.has_sufficient_balance(source_balance, *COST_OF_PUT) {
```
nbaksalyar: We don't need this check here: it's already performed by `self.authorise_operations`.
It's OK to address this in later PRs though. :)

safe_core/src/client/mock/vault.rs
```diff
@@ -588,36 +586,32 @@ impl Vault {
...
-                                    // an additional PUT operation.
-                                    if source_balance.checked_sub(amt).is_none() {
-                                        return Err(SndError::InsufficientBalance);
-                                    }
-                                }
-                                None => return Err(SndError::ExcessiveValue),
+                            let debit_amt = amount
+                                .checked_add(*COST_OF_PUT)
+                                .ok_or(SndError::ExcessiveValue)?;
+                            if !self.has_sufficient_balance(source_balance, debit_amt) {
```
nbaksalyar: Same here: this check is already performed by `authorise_operations`.

PR #1051, opened 2019-10-21
Add function to set path from which the config file is read

lionel1704:


safe_core/src/config_handler.rs
```diff
@@ -84,21 +125,41 @@ pub fn get_config() -> Config {
...
-    ProjectDirs::from(
-        CONFIG_DIR_QUALIFIER,
-        CONFIG_DIR_ORGANISATION,
-        VAULT_CONFIG_DIR_APPLICATION,
-    )
-    .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "Home directory not found").into())
+    let project_dirs = if let Some(custom_path) = unwrap!(CONFIG_DIR_PATH.lock()).clone() {
+        ProjectDirs::from_path(custom_path)
+    } else {
+        ProjectDirs::from(
```
m-cat: We can use `lazy_static` here instead of computing the project dirs every time; see safe_vault:

https://github.com/maidsafe/safe_vault/blob/master/src/config_handler.rs#L24

safe_core/src/config_handler.rs
```diff
@@ -84,21 +125,41 @@ pub fn get_config() -> Config {
...
-    ProjectDirs::from(
-        CONFIG_DIR_QUALIFIER,
-        CONFIG_DIR_ORGANISATION,
-        CONFIG_DIR_APPLICATION,
-    )
-    .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "Home directory not found").into())
+    let project_dirs = if let Some(custom_path) = unwrap!(CONFIG_DIR_PATH.lock()).clone() {
+        ProjectDirs::from_path(custom_path)
+    } else {
+        ProjectDirs::from(
```
m-cat: Same comment as below.

safe_core/src/config_handler.rs
```diff
@@ -84,21 +125,41 @@ pub fn get_config() -> Config {
...
+            CONFIG_DIR_QUALIFIER,
+            CONFIG_DIR_ORGANISATION,
+            VAULT_CONFIG_DIR_APPLICATION,
+        )
+    };
+    project_dirs.ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::NotFound,
+            "Cannot determine project directory paths",
```
m-cat: I would make this error different from the one in `dirs`, otherwise someone who comes across this error and tries searching for it won't know exactly where it comes from.

safe_core/src/config_handler.rs
```diff
@@ -25,6 +28,15 @@ const CONFIG_FILE: &str = "safe_core.config";
 const VAULT_CONFIG_DIR_APPLICATION: &str = "safe_vault";
 const VAULT_CONNECTION_INFO_FILE: &str = "vault_connection_info.config";
 
+lazy_static! {
+    static ref CONFIG_DIR_PATH: Mutex<Option<PathBuf>> = Mutex::new(None);
+}
+
+/// Set a custom path for the config files.
+pub fn set_config_dir_path<P: AsRef<OsStr> + ?Sized>(path: &P) {
```
m-cat: I'm probably missing something but I'm not sure why this has to take an `OsStr`. A regular `String` can be converted to `PathBuf` just as well. Maybe leave a comment explaining this?

lionel1704: TBH, I'm not too sure about this @m-cat . I simply followed (copy-pasted :P) the same pattern used in `config_file_handler::set_additonal_search_path`

m-cat: I see, no worries. Can you just add a comment to that effect?

safe_core/src/config_handler.rs
```diff
@@ -25,6 +28,36 @@ const CONFIG_FILE: &str = "safe_core.config";
 const VAULT_CONFIG_DIR_APPLICATION: &str = "safe_vault";
 const VAULT_CONNECTION_INFO_FILE: &str = "vault_connection_info.config";
 
+lazy_static! {
+    static ref CONFIG_DIR_PATH: Mutex<Option<PathBuf>> = Mutex::new(None);
+    static ref SAFE_CORE_PROJECT_DIRS: Option<ProjectDirs> =
+        if let Some(custom_path) = unwrap!(CONFIG_DIR_PATH.lock()).clone() {
```
m-cat: This doesn't seem right to me -- we only calculate `SAFE_CORE_PROJECT_DIRS` once, after which the value of `CONFIG_DIR_PATH` can change. I think the part that depends on `CONFIG_DIR_PATH` should go back to the function, the only thing I had in mind for putting in `lazy_static` was

```rust
ProjectDirs::from(
    CONFIG_DIR_QUALIFIER,
    CONFIG_DIR_ORGANISATION,
    CONFIG_DIR_APPLICATION,
)
```

as in safe_vault.

lionel1704: Ah I see what you mean. My bad.
But we'll still need two. One for safe_core and one for safe_vault. They will be the fallback values if `CONFIG_DIR_PATH` is not set, yes?

m-cat: Yep, exactly.

safe_authenticator/src/ffi/mod.rs
```diff
@@ -152,6 +152,21 @@ pub unsafe extern "C" fn auth_exe_file_stem(
...
+    o_cb: extern "C" fn(user_data: *mut c_void, result: *const FfiResult),
+) {
+    catch_unwind_cb(user_data, o_cb, || -> Result<_, AuthError> {
+        let new_path = CStr::from_ptr(new_path).to_str()?;
+        config_handler::set_config_dir_path(OsStr::new(new_path));
+        o_cb(user_data, FFI_RESULT_OK);
+        Ok(())
+    });
+}
+
```
loziniak: Seems like violation of DRY principle.

lionel1704: Hi @loziniak
Thanks for your review. The repetition is because this is a step towards deprecating `auth_set_additional_search_path`.

As of this PR the reading of the config file is migrated but `maidsafe_utilities::init_logging` still uses the path from `set_additional_search_path`. That will be deprecated in the near future.

m-cat: @loziniak Do you mean the duplication of this function between `safe_auth` and `safe_app`?

loziniak: @m-cat yes, they look exactly the same. But now, when I think about it, there is mostly bilerplate code, so duplication is probably not so important here...

m-cat: Yeah, we have some ideas on reducing the FFI boilerplate but we may not get around to it for a while. As for the two separate functions, I think the repetition here is unfortunately necessary as `safe_auth` and `safe_app` are two separate libraries. Thanks for pointing this out!

PR #1052, opened 2019-10-24
tests/client: test to verify that coins are refunded on a failed transaction

Yoga07:


safe_authenticator/src/client.rs
```diff
@@ -902,22 +902,49 @@ mod tests {
...
+                .and_then(move |_| {
+                    c1.insert_login_packet_for(
+                        None,
+                        maid_keys.bls_pk.into(),
+                        unwrap!(Coins::from_str("3")),
+                        None,
+                        new_login_packet.clone(),
+                    )
+                })
+                // Re-insert to check for refunds for a failed Login packet operation
```
m-cat: I think this comment should be above the call to `insert_login_packet_for` 🙂

safe_core/src/client/mod.rs
```diff
@@ -1936,18 +1970,26 @@ mod tests {
                         Err(CoreError::DataError(SndError::InsufficientBalance)) => (),
                         res => panic!("Unexpected result: {:?}", res),
                     }
-                    c5.transfer_coins(None, wallet1, unwrap!(Coins::from_str("0")), None)
+                    Ok(())
+                })
+                // Check if coins are refunded
+                .and_then(move |_| c5.get_balance(None))
+                .and_then(move |balance| {
+                    let mut expected = unwrap!(Coins::from_str("5"));
```
m-cat: `expected` doesn't need to be `mut`, you can use variable shadowing:

```rust
let expected = initialize();
let expected = expected.do_something();
```

Not a big deal here, but this is more idiomatic so it's a good thing to keep in mind. 🙂

safe_core/src/client/mod.rs
```diff
@@ -1879,6 +1903,15 @@ mod tests {
                             Ok(())
                         })
                 })
+                // Check if coins are refunded after a failed CreateBalance Operation.
+                .and_then(move |_| {
+                    client6.get_balance(None).and_then(|balance| {
+                        let mut expected = unwrap!(Coins::from_str("405.0"));
```
m-cat: Same comment here about unnecessary `mut`

safe_authenticator/src/client.rs
```diff
@@ -902,22 +902,49 @@ mod tests {
...
+                .then(|result| match result {
+                    Err(CoreError::DataError(SndError::LoginPacketExists)) => Ok::<_, CoreError>(()),
+                    res => panic!("Unexpected {:?}", res),
+                })
+                // Check if coins are refunded
+                .and_then(move |_| c2.get_balance(None))
+                .and_then(move |balance| {
+                    assert_eq!(
+                        balance,
+                        unwrap!(start_bal.checked_sub(unwrap!(Coins::from_str("5.000000002"))))
```
lionel1704: Instead of hard coding this can this be something like.
`Coins::from_nano(COST_OF_PUT.0 * 2) + 5 coins`
That way we won't need to change anything if COST_OF_PUT changes.

safe_core/src/client/mod.rs
```diff
@@ -1431,9 +1434,17 @@ mod tests {
                         Err(e) => panic!("Unexpected: {:?}", e),
                     })
                 })
+                .and_then(move |_| client4.get_balance(None))
+                .and_then(move |balance| {
+                    assert_eq!(
+                        balance,
+                        unwrap!(start_bal.checked_sub(unwrap!(Coins::from_nano(2))))
```
lionel1704: Same comment about hard-coding "2".

safe_core/src/client/mod.rs
```diff
@@ -1486,23 +1501,31 @@ mod tests {
                         res => panic!("Unexpected: {:?}", res),
                     }
                 })
+                .and_then(move |_| client4.get_balance(None))
+                .and_then(move |balance| {
+                    assert_eq!(
+                        balance,
+                        unwrap!(start_bal.checked_sub(unwrap!(Coins::from_nano(2))))
```
lionel1704: Here too.

safe_core/src/client/mod.rs
```diff
@@ -2043,11 +2088,20 @@ mod tests {
                             Err(e) => panic!("Unexpected: {:?}", e),
                         })
                 })
+                // Check if coins are refunded
+                .and_then(move |_| client2.get_balance(None))
+                .and_then(move |balance| {
+                    assert_eq!(
+                        balance,
+                        unwrap!(start_bal.checked_sub(unwrap!(Coins::from_nano(2))))
```
lionel1704: Here as well.

safe_authenticator/src/client.rs
```diff
@@ -902,22 +902,51 @@ mod tests {
...
                     Ok(_transaction) => Ok::<_, CoreError>(()),
                     res => panic!("Unexpected {:?}", res),
                 })
+                .and_then(move |_| {
+                    c1.insert_login_packet_for(
+                        None,
+                        maid_keys.bls_pk.into(),
+                        unwrap!(Coins::from_str("3")),
+                        None,
+                        new_login_packet.clone(),
```
lionel1704: Does this need to be cloned here?

safe_core/src/client/mod.rs
```diff
@@ -1854,8 +1874,8 @@ mod tests {
                 })
                 .and_then(move |_| {
                     client4.get_balance(None).and_then(|balance| {
-                        let mut expected = unwrap!(Coins::from_str("405.0"));
-                        expected = unwrap!(expected.checked_sub(*COST_OF_PUT));
+                        let expected = unwrap!(Coins::from_str("405.0"));
```
lionel1704: Can use the new helper function here.

safe_core/src/client/mod.rs
```diff
@@ -1879,6 +1899,15 @@ mod tests {
                             Ok(())
                         })
                 })
+                // Check if coins are refunded after a failed CreateBalance Operation.
+                .and_then(move |_| {
+                    client6.get_balance(None).and_then(|balance| {
+                        let expected = unwrap!(Coins::from_str("405.0"));
```
lionel1704: Same comment as above.

safe_core/src/utils/test_utils/mod.rs
```diff
@@ -161,3 +162,16 @@ where
 
     unwrap!(result_rx.recv())
 }
+
+/// Helper function to calculate the total cost of expenditure by adding number of mutations and
+/// amount of transferred coins if any.
+pub fn calc_total_cost(mutation_count: Option<u64>, transferred_coins: Option<Coins>) -> Coins {
```
lionel1704: Do you think we can change this function to `calculate_new_balance(balance: ..., mutation_count: ..., transferred_coins: ...)`
That way we need not call `Coins::from_nano(0)` - it seems unnecessary.
And the redundant `unwrap!(balance.checked_sub(cost))` can also be removed.

PR #1058, opened 2019-11-01
refactor/keys: refactor ClientKeys and AppKeys to use full IDs

m-cat:


m-cat:
Yes, this PR depends on the safe-nd changes. Among other things, we create Keypairs in tests to get public keys -- technically not needed, but it is convenient.

PR #1062, opened 2019-11-05
WIP: remove maidsafe_utilites dependency

### LINKED ISSUE #1059: Deprecate the use of maidsafe_utilities
It is desired that we move away from the use of `maidsafe_utilities`. Identify the usages of modules from maidsafe utilities and replace them with their alternatives.

- [ ] Serialisation
          Use bincode which is already being used in other places.
- [ ] Logging
          Use log4rs which is the underlying dependency for maidsafe_utilities. Make sure that all the use cases work correctly.
          - Logging to `Client.log` file.
          - Logging to a file with custom name.
          - The log levels in the `log.toml` file are picked up correctly.
          - Specifying a custom path for the log files are picked up correctly.
          - TODO: Anything more?
- [ ] Threads
          - Replace the use of named threads that are used only for testing.
- [ ] RNG
          - Used only for testing. Replace with alternative.

Yoga07:
Partially resolves #1059 
- This is a WIP PR - Logging is yet to be removed to completely deprecate `maidsafe_utilities` from dependency

### DIFF
--- a/safe_app/src/ffi/mutable_data/mod.rs
+++ b/safe_app/src/ffi/mutable_data/mod.rs
@@ -22,11 +22,11 @@ use crate::ffi::object_cache::{
     NULL_OBJECT_HANDLE,
 };
 use crate::App;
+use bincode::serialized_size;
 use ffi_utils::{
     catch_unwind_cb, vec_clone_from_raw_parts, FfiResult, OpaqueCtx, ReprC, SafePtr, FFI_RESULT_OK,
 };
 use futures::Future;
-use maidsafe_utilities::serialisation::serialised_size;
 use safe_core::ffi::ipc::req::PermissionSet;
 use safe_core::ffi::ipc::resp::MDataKey;
 use safe_core::ffi::ipc::resp::MDataValue;
@@ -143,7 +143,7 @@ pub unsafe extern "C" fn mdata_serialised_size(
             client
                 .get_seq_mdata(info.name(), info.type_tag())
                 .map_err(AppError::from)
-                .and_then(move |mdata| Ok(serialised_size(&mdata)))
+                .and_then(move |mdata| serialized_size(&mdata).map_err(AppError::from))
         })
     })
 }
--- a/safe_authenticator/src/tests/revocation.rs
+++ b/safe_authenticator/src/tests/revocation.rs
@@ -158,7 +158,7 @@ mod mock_routing {
     };
     use config;
     use ffi_utils::test_utils::call_0;
-    use maidsafe_utilities::SeededRng;
+    use rand::XorShiftRng;
     use safe_core::client::AuthActions;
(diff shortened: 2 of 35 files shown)


safe_app/src/ffi/mutable_data/mod.rs
```diff
@@ -143,7 +143,7 @@ pub unsafe extern "C" fn mdata_serialised_size(
             client
                 .get_seq_mdata(info.name(), info.type_tag())
                 .map_err(AppError::from)
-                .and_then(move |mdata| Ok(serialised_size(&mdata)))
+                .and_then(move |mdata| serialized_size(&mdata).map_err(AppError::from))
```
m-cat: This `.map_err` shouldn't be necessary. Try moving this `and_then` up one line, before the previous `map_err`.

safe_authenticator/src/tests/revocation.rs
```diff
@@ -400,7 +400,7 @@ mod mock_routing {
     // Test one app being revoked by multiple authenticator concurrently.
     #[test]
     fn concurrent_revocation_of_single_app() {
-        let rng = SeededRng::new();
+        let rng = XorShiftRng::new_unseeded();
```
m-cat: Looking at the implementation of `SeededRng::new()`, it looks like it creates a static rng for the thread (using `thread_local!`) -- the equivalent would be to replace it with `rand::thread_rng()`.

safe_core/src/utils/test_utils/sync.rs
```diff
@@ -27,7 +28,7 @@ impl Synchronizer {
     /// Create new instance of `Synchronizer` using the given random number
     /// generator. The generator can be initialized with a seed to guarantee
     /// repeatable, deterministic runs.
-    pub fn new(rng: SeededRng) -> Self {
+    pub fn new(rng: XorShiftRng) -> Self {
```
m-cat: I would take a generic `Rng` here (trait defined in `rand`). `XorShiftRng` in particular is not defined in the latest version of `rand`. Speaking of which, can we update to `rand = "~0.7"` in this PR?

m-cat: Actually, just noticed that `rand` needs to be at `"~0.6"` to avoid conflicts with `ed25519_dalek`.

m-cat: I just did the work in my WIP PR to upgrade to "0.6", let me know if you want to see that.

lionel1704:
Closing in favour of #1069

safe_core/src/utils/test_utils/sync.rs
lionel1704: @m-cat since you have already implemented this in your PR do you think we can merge these changes and let your PR take care of this?

Issue #1055, opened 2019-10-31
Move FFI error codes to FFI modules


m-cat:
Currently, FFI-specific error codes are stored outside of the safe_auth and safe_app FFI modules ([example](https://github.com/maidsafe/safe_client_libs/blob/2468c6012cb0ca5ca4aa2348e76099436a050701/safe_authenticator/src/errors.rs#L29)).

1. These error codes should be moved to the corresponding FFI modules for better separation of concerns (i.e. FFI-specific functionality should not appear outside of FFI modules). Follow [this PR](https://github.com/maidsafe/safe-api/pull/296) as a guide.
2. Separate FFI error types can be implemented, as in that PR.
3. Also, most of the error codes are shared between safe_auth and _app and can actually go into safe_core FFI.

peter-wilkins:
Hi, as you can see I've had a crack at this.

Not sure I've fulfilled point 2: Separate FFI error types can be implemented.

I've only run the mock tests - will look into now to run a network locally.

m-cat:
Thanks for the PR! I'll take a look tomorrow.

> I've only run the mock tests - will look into how to run a network locally.

I don't think that's strictly necessary, this is just a refactor, right? Theoretically the functionality shouldn't change -- just make sure it compiles 🙂

PR #1063, opened 2019-11-05
WIP refactor/ffi: move safe_app and safe_auth errors into ffi modules

peter-wilkins:
This PR moves FFI-specific error codes into ffi module.
Common error codes have been moved to safe_core.

Resolves https://github.com/maidsafe/safe_client_libs/issues/1055

### DIFF
--- a/safe_app/src/lib.rs
+++ b/safe_app/src/lib.rs
@@ -108,7 +108,6 @@ pub use crate::ffi::*;
 
 pub mod cipher_opt;
 mod client;
-mod errors;
 pub mod object_cache;
 pub mod permissions;
 
@@ -119,7 +118,7 @@ mod tests;
 #[cfg(any(test, feature = "testing"))]
 pub mod test_utils;
 
-pub use self::errors::*;
+pub use self::ffi::errors::*;
 pub use client::AppClient;
 
 use self::object_cache::ObjectCache;
--- a/safe_authenticator/src/ffi/errors.rs
+++ b/safe_authenticator/src/ffi/errors.rs
@@ -1,4 +1,4 @@
-// Copyright 2018 MaidSafe.net limited.
+// Copyright 2019 MaidSafe.net limited.
 //
 // This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
 // Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
@@ -6,13 +6,12 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-//! Errors thrown by Authenticator routines.
-
 pub use self::codes::*;
 use config_file_handler::Error as ConfigFileHandlerError;
 use ffi_utils::{ErrorCode, StringError};
 use futures::sync::mpsc::SendError;
 use maidsafe_utilities::serialisation::SerialisationError;
+pub use safe_core::ffi::errors::shared_codes::*;
 use safe_core::ipc::IpcError;
 use safe_core::nfs::NfsError;
 use safe_core::CoreError;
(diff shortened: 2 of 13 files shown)


safe_core/src/ffi/errors.rs
```diff
@@ -0,0 +1,75 @@
...
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+#[allow(missing_docs)]
+#[allow(dead_code)]
```
m-cat: I don't think we need to allow `dead_code` here since the error codes are being publicly exported.

safe_authenticator/src/ffi/errors.rs
```diff
@@ -1,18 +1,17 @@
...
 // permissions and limitations relating to use of the SAFE Network Software.
 
-//! Errors thrown by Authenticator routines.
-
 pub use self::codes::*;
 use config_file_handler::Error as ConfigFileHandlerError;
 use ffi_utils::{ErrorCode, StringError};
 use futures::sync::mpsc::SendError;
 use maidsafe_utilities::serialisation::SerialisationError;
+pub use safe_core::ffi::errors::shared_codes::*;
```
m-cat: It's a good idea to re-export the shared error codes here so that they're accessible to the FFI caller 👍 

One thing though, can you move this up to line 10 and add a newline after this? We like to keep imports (`use`) and exports (`pub use`) visually separated.

Also, can you add the same line in `safe_app/src/ffi/errors.rs`?

safe_app/src/lib.rs
```diff
@@ -108,7 +108,6 @@ pub use crate::ffi::*;
 
 pub mod cipher_opt;
 mod client;
-mod errors;
```
m-cat: Instead of moving the `errors.rs` files to `ffi` entirely, we should be keeping it with only the non-FFI stuff. For example, it looks like you moved `AppError` to `ffi/errors.`, but this error type is used widely outside of the FFI module -- so we again have a poor separation of concerns. What I really had in mind was to create a *new* FFI error type, specific to the FFI module. You can look at the linked `safe-cli` PR in the original issue for an example of creating a separate FFI error type while leaving the original, non-FFI one.

safe_app/src/lib.rs
```diff
@@ -119,7 +118,7 @@ mod tests;
 #[cfg(any(test, feature = "testing"))]
 pub mod test_utils;
 
-pub use self::errors::*;
+pub use self::ffi::errors::*;
```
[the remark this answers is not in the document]
m-cat: Also, we need to export the crate's error type here:

```rust
pub use self::errors::AppError;
```

This can stay on line 121 since it is a non-FFI export.

Looks like we already do this in `safe_auth` so we're good there 👍

safe_app/src/lib.rs
peter-wilkins: Hi, thanks for all the feedback. 

https://github.com/maidsafe/safe_client_libs/issues/1066
I saw your new issue here about result types and something clicked.

I've started a separate PR on my fork to see how FFI Error and Result types look and check I'm on the right track before going on.
https://github.com/peter-wilkins/safe_client_libs/pull/1/files

It's a bit more involved than the safe_cli refactor for this project I think?
I've used the new Result type throughout the ffi folder but I haven't been able to move the
`impl ErrorCode for AppError` to the ffi Error type as there are interesting calls to ffi_util macros and callbacks I'm still trying to understand.  
eg http://github.com/peter-wilkins/safe_client_libs/blob/5848530a03c792f4c6d256bc07cfcd6946c60494/safe_app/src/ffi/crypto.rs#L472-L472

Is it ok that AppError implements the ErrorCode trait?

m-cat: > Is it ok that AppError implements the ErrorCode trait?

Hmm, this is what we are trying to avoid. Only the FFI version of AppError should implement it. The code you linked is from the FFI module, so we shouldn't be using the non-FFI AppError there.

Yes, unfortunately our FFI code is hard to understand. To find out more about it, I invite you to read [our wiki](https://github.com/maidsafe/safe_client_libs/wiki/FFI-overview) -- you may have to do some clicking around to find what you're looking for, but hopefully you find the info interesting.

Please let me know if all this is too much work, we definitely don't want to take advantage of the community. I had in mind that this would be a small task, and I feel a bit guilty.

Let me know if you have any questions. Just FYI, I'm going to Switzerland for a couple of days and won't be back at work until Wednesday.

peter-wilkins:
closing this PR as superceded by https://github.com/maidsafe/safe_client_libs/pull/1068

safe_app/src/lib.rs
peter-wilkins: Ok, I managed to do it. :crossed_fingers:  https://github.com/maidsafe/safe_client_libs/pull/1068
Don't feel guilty, this is the first time I'm tackling a big rust codebase and I'm loving it. The compiler is doing all the work!
Have a good trip. Be warned when you get back I may have done all the 'need help' issues. :smiley:

PR #1064, opened 2019-11-06
Fix broken path in config_handler

lionel1704:


safe_core/src/config_handler.rs
```diff
@@ -201,31 +198,35 @@ pub fn write_config_file(config: &Config) -> Result<PathBuf, CoreError> {
...
-
-        let mut file = unwrap!(File::create(&path));
-        unwrap!(serde_json::to_writer_pretty(&mut file, &config));
+        let config: Config = Default::default();
+        unwrap!(write_config_file(&config));
 
-        let read_cfg: QuicP2pConfig = unwrap!(Config::read_qp2p_from_file());
-        assert_eq!(config, read_cfg)
+        let read_cfg = Config::new();
+        assert_eq!(config, read_cfg);
```
m-cat: To make sure we really are reading the `config` that was written to disk, we should change one of `config`'s fields to something unique before writing it, rather than writing a default config.

lionel1704: Makes sense. It might not be obvious, but we might already have that.

`Default::default()` will set the `our_type` field to `Node` by default.
Whereas, if the file was missing `our_type` will be set to `Client`.

Do you think this is sufficient?

m-cat: Yeah, that wasn't obvious to me when reviewing. I would suggest adding some comments here as this test relies a lot on the behavior of `new` and `default`, which readers may not be aware of.

PR #1068, opened 2019-11-08
refactor/ffi: FFI-specific Result types for Safe App and Safe Auth

peter-wilkins:
This PR creates FFI-specific Result types for better separation.

Common FFI error codes have been moved to safe_core.

Resolves https://github.com/maidsafe/safe_client_libs/issues/1055

PR #1074, opened 2019-11-19
refactor/clippy: fix all clippy warnings

peter-wilkins:
requested here: https://github.com/maidsafe/safe_client_libs/pull/1068#issuecomment-554485439

mostly redundant clones and a few unit_cmp warnings:

https://rust-lang.github.io/rust-clippy/master/index.html#unit_cmp

### DIFF
--- a/safe_core/src/client/mod.rs
+++ b/safe_core/src/client/mod.rs
@@ -129,7 +129,7 @@ fn send_as_helper(
     let mut cm2 = cm.clone();
 
     Box::new(
-        cm.bootstrap(identity.clone())
+        cm.bootstrap(identity)
             .and_then(move |_| cm2.send(&pub_id, &message)),
     )
 }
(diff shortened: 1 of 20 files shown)


safe_core/src/client/mod.rs
```diff
@@ -2200,11 +2180,8 @@ mod tests {
                         .ins(b"key3".to_vec(), b"value".to_vec(), 0);
 
                     client3
-                        .mutate_seq_mdata_entries(name, tag, entry_actions.clone())
-                        .then(move |res| {
-                            assert_eq!(unwrap!(res), ());
-                            Ok(())
-                        })
+                        .mutate_seq_mdata_entries(name, tag, entry_actions)
+                        .then(move |_| Ok(()))
```
lionel1704: When using `.then(...)` if the result is ignored then we wouldn't know if there is a failure. `.and_then()` on the other hand would propagates the error to the outer function. However, asserting that the result is a success would help easily pinpoint which function failed. If you agree, could you please make this change here and in some other places too?
```suggestion
                        .then(move |res| {
                             assert!(res.is_ok());
                             Ok(())
                        })
```

m-cat:
Thanks for the PR. Looks like all these clippy fixes are already in the `master` branch 🙂

PR #1088, opened 2019-11-21
chore/ci: make misc fixes in GHA workflows and readme

m-cat:
1. Update Readme to display result of last Master build (and fix the Issues badge)
2. Fix deploy steps not being version-change-gated
3. Remove `SAFE_MOCK_IN_MEMORY_STORAGE` from the `tests` job (would cause the `config_mock_vault_path` test to fail once that is re-enabled)
4. Use the `test-mock` script in the `test` job

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -205,6 +205,14 @@ jobs:
           #name: safe_client_libs-x86_64-apple-darwin-dev
           #path: artifacts/dev/x86_64-apple-darwin/release
 
+      # Upload all the release archives to S3.
+      - uses: actions/aws/cli@master
+        with:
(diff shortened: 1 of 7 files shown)


lionel1704:
As far as I understand, there are S3 deploys are done for every master build as well. Do we need that condition for the deploys?

m-cat:
@lionel1704 if I understand correctly, are you suggesting that I remove the condition for the S3 uploads? I am not too familiar with our deploy process yet 🙂

lionel1704:
Yes please @m-cat. Let's remove the condition from the S3 releases alone. The condition is right for the `upload-release-artifacts` steps.

PR #1090, opened 2019-11-22
Update ffi-utils to 0.13.0

lionel1704:


m-cat:
Just one thing @lionel1704 , previously the `from_c_str` function would return an empty String for a null pointer input, the new function will return an error instead. Have you checked if we rely on the previous behavior anywhere? It's probably fine as the tests are passing, but I'll do a quick audit if you haven't already.

lionel1704:
Thanks for the heads up @m-cat. I went thorough the code and we don't rely on the previous behavior. So we're all good.

PR #1093, opened 2019-11-25
Migrate signing away from rust-sodium

### LINKED ISSUE #1079: Migrate signing away from rust-sodium
Remove the use of `rust_sodium::crypto::sign::*`. It is used as a part of the `AppKeys` and `ClientKeys` struct. It's use is redundant as of #1058. Signing will be done by `AppFullID` and `ClientFullId`. The relevant functions in `safe_app::ffi::crypto` should also be updated to use BLS keys for signing.

lionel1704:
closes #1079


safe_core/src/utils/seed.rs
```diff
@@ -6,22 +6,11 @@
...
-/// Calculate sign key from seed.
-pub fn sign_pk_from_seed(seed: &str) -> Result<sign::PublicKey, CoreError> {
-    let arr = divide_seed(seed)?;
-    let id_seed = Seed(sha3_256(arr[SEED_SUBPARTS - 2]));
-    let maid_keys = ClientKeys::new(Some(&id_seed));
-    Ok(maid_keys.sign_pk)
-}
-
 /// Divide `seed` into the number of subparts given by `SEED_SUBPARTS`.
 pub fn divide_seed(seed: &str) -> Result<[&[u8]; SEED_SUBPARTS], CoreError> {
```
m-cat: I think it should be possible to deprecate this entire file.

safe_app/src/tests/append_only_data.rs
```diff
@@ -105,9 +105,9 @@ fn managing_permissions_for_an_app() {
             let client5 = client.clone();
             let client6 = client.clone();
 
-            let sign_pk = client.public_key();
+            let app_pk = client.public_key();
```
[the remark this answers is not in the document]
lionel1704: Yes I have. I just did another re-check to verify too. Everything seems right.
We can consider renaming those variables. But that can be in a separate PR.

safe_authenticator/src/client.rs
```diff
@@ -80,7 +79,8 @@ impl AuthClient {
 where {
```
m-cat: Can we remove this `where`? I'm surprised the compiler doesn't complain about it 😄

lionel1704: I'll sneak this change in a different PR :)

PR #1094, opened 2019-11-27
Fixes for mobile builds

lionel1704:


safe_core/src/ffi/ipc/resp.rs
```diff
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::ffi::arrays;
+use crate::ffi::arrays::*;
```
m-cat: Maybe you can add a comment above this line saying that it's currently required for safe_bindgen to parse correctly?

safe_core/src/ffi/ipc/resp.rs
```diff
@@ -18,33 +18,34 @@ use std::ptr;
 #[repr(C)]
 #[derive(Clone, Copy)]
 pub struct AppKeys {
```
m-cat: Just noticed something, I think you need to implement `Drop` for this struct since it has ownership of the full id bytes.

Also, can we do the same thing for `ClientKeys` or is that for a later PR?

lionel1704: We don't have an FFI representation for ClientKeys :)

safe_core/src/ipc/resp.rs
```diff
@@ -210,6 +210,28 @@ impl ReprC for AppKeys {
     }
 }
 
+/// Convert FFI representation of `AppKeys` to native rust representation by cloning.
```
m-cat: Why add this function when we have the very similar `AppKeys::clone_from_repr_c`, apart from the input type? Maybe try `type C = *const ffi::AppKeys;`?

PR #1099, opened 2019-12-02
Migrate Asymmetric encryptions from rust_sodium to threshold_crypto

### LINKED ISSUE #1081: Migrate asymmetric encryption away from rust-sodium
Remove the use of `rust_sodium::sealedbox::*` and replace it with functions from the threshold_crypto crate for asymmetric encryption.

Yoga07:
Rebased on top of #1096 
Resolves #1081

### DIFF
--- a/safe_app/build.rs
+++ b/safe_app/build.rs
@@ -17,12 +17,12 @@ fn main() {
 #[cfg(feature = "bindings")]
 mod bindings {
     use jni::signature::{JavaType, Primitive};
-    use rust_sodium::crypto::{box_, sign};
     use safe_bindgen::{Bindgen, FilterMode, LangC, LangCSharp, LangJava};
     use safe_nd::XOR_NAME_LEN;
     use std::collections::HashMap;
     use std::env;
     use std::path::Path;
+    use threshold_crypto::PK_SIZE;
     use unwrap::unwrap;
 
     const BSD_MIT_LICENSE: &str =
@@ -201,11 +201,7 @@ mod bindings {
             "BindingUtils",
         );
 
-        lang.add_const("ulong", "ASYM_PUBLIC_KEY_LEN", box_::PUBLICKEYBYTES);
-        lang.add_const("ulong", "ASYM_SECRET_KEY_LEN", box_::SECRETKEYBYTES);
-        lang.add_const("ulong", "ASYM_NONCE_LEN", box_::NONCEBYTES);
-        lang.add_const("ulong", "SIGN_PUBLIC_KEY_LEN", sign::PUBLICKEYBYTES);
-        lang.add_const("ulong", "SIGN_SECRET_KEY_LEN", sign::SECRETKEYBYTES);
+        lang.add_const("ulong", "ASYM_PUBLIC_KEY_LEN", PK_SIZE);
         lang.add_const("ulong", "XOR_NAME_LEN", XOR_NAME_LEN);
         lang.add_opaque_type("App");
 
--- a/safe_app/src/ffi/crypto.rs
+++ b/safe_app/src/ffi/crypto.rs
@@ -17,13 +17,12 @@ use crate::App;
 use bincode::{deserialize, serialize};
 use ffi_utils::{catch_unwind_cb, vec_clone_from_raw_parts, FfiResult, OpaqueCtx, FFI_RESULT_OK};
 use rand::thread_rng;
-use rust_sodium::crypto::{box_, sealedbox};
 use safe_core::crypto::shared_box;
-use safe_core::ffi::arrays::{AsymNonce, AsymPublicKey, AsymSecretKey};
-use safe_core::Client;
+use safe_core::{AsymPublicKey, Client};
 use safe_nd::{ClientFullId, PublicKey, Signature};
 use std::os::raw::c_void;
 use std::slice;
+use threshold_crypto::{Ciphertext, PublicKey as AsymEncryptKey};
 use tiny_keccak::sha3_256;
 
 /// Special value that represents that a message should be signed by the app.
@@ -275,7 +274,8 @@ pub unsafe extern "C" fn enc_pub_key_new(
     ),
 ) {
     catch_unwind_cb(user_data, o_cb, || {
-        let key = box_::PublicKey(*data);
+        let raw = data.as_ref().ok_or_else(|| AppError::EncodeDecodeError)?;
+        let key = AsymEncryptKey::from_bytes(raw).map_err(|_| AppError::EncodeDecodeError)?;
         send_sync(app, user_data, o_cb, move |_, context| {
             Ok(context.object_cache().insert_encrypt_key(key))
         })
@@ -296,8 +296,9 @@ pub unsafe extern "C" fn enc_pub_key_get(
 ) {
     catch_unwind_cb(user_data, o_cb, || {
         send_sync(app, user_data, o_cb, move |_, context| {
-            let key = context.object_cache().get_encrypt_key(handle)?;
-            Ok(&key.0)
+            let key: AsymEncryptKey = *context.object_cache().get_encrypt_key(handle)?;
+            let buf = &key.to_bytes();
+            Ok(buf)
         })
     })
 }
@@ -322,7 +323,8 @@ pub unsafe extern "C" fn enc_pub_key_free(
 #[no_mangle]
 pub unsafe extern "C" fn enc_secret_key_new(
     app: *const App,
-    data: *const AsymSecretKey,
+    sk: *const u8,
+    sk_len: usize,
     user_data: *mut c_void,
     o_cb: extern "C" fn(
         user_data: *mut c_void,
@@ -331,7 +333,9 @@ pub unsafe extern "C" fn enc_secret_key_new(
     ),
 ) {
     catch_unwind_cb(user_data, o_cb, || {
-        let key = shared_box::SecretKey::from_raw(&*data);
+        let raw = vec_clone_from_raw_parts(sk, sk_len);
+        let key =
+            shared_box::SecretKey::from_raw(&*raw).map_err(|_| AppError::EncodeDecodeError)?;
         send_sync(app, user_data, o_cb, move |_, context| {
             Ok(context.object_cache().insert_secret_key(key))
         })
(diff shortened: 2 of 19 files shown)


safe_app/src/ffi/crypto.rs
```diff
@@ -266,7 +265,8 @@ pub unsafe extern "C" fn enc_generate_key_pair(
 #[no_mangle]
 pub unsafe extern "C" fn enc_pub_key_new(
     app: *const App,
-    data: *const AsymPublicKey,
+    data: *const u8,
```
lionel1704: `threshold_crypto::PublicKey` has constant length. This can be an array defined in `ffi::arrays`

safe_app/src/ffi/crypto.rs
```diff
@@ -291,13 +293,16 @@ pub unsafe extern "C" fn enc_pub_key_get(
     o_cb: extern "C" fn(
         user_data: *mut c_void,
         result: *const FfiResult,
-        pub_enc_key: *const AsymPublicKey,
+        pub_enc_key: *const u8,
```
lionel1704: Same comment about the length being fixed.

safe_app/src/ffi/crypto.rs
```diff
@@ -466,7 +476,6 @@ pub unsafe extern "C" fn encrypt(
     data: *const u8,
     data_len: usize,
     public_key_h: EncryptPubKeyHandle,
-    secret_key_h: EncryptSecKeyHandle,
```
lionel1704: This function is for public key encryption. It should not be changed.

safe_app/src/ffi/crypto.rs
```diff
@@ -515,7 +517,6 @@ pub unsafe extern "C" fn decrypt(
     app: *const App,
```
lionel1704: Same comment as above.

safe_app/src/ffi/crypto.rs
```diff
@@ -615,7 +609,7 @@ pub unsafe extern "C" fn decrypt_sealed_box(
...
     data: *const u8,
     data_len: usize,
-    public_key_h: EncryptPubKeyHandle,
+    _public_key_h: EncryptPubKeyHandle,
```
lionel1704: This field can be removed. 
The doc needs to be updated as well.

safe_authenticator/build.rs
```diff
@@ -197,9 +197,7 @@ mod bindings {
             "BindingUtils",
         );
 
-        lang.add_const("ulong", "ASYM_PUBLIC_KEY_LEN", box_::PUBLICKEYBYTES);
-        lang.add_const("ulong", "ASYM_SECRET_KEY_LEN", box_::SECRETKEYBYTES);
-        lang.add_const("ulong", "ASYM_NONCE_LEN", box_::NONCEBYTES);
+        lang.add_const("ulong", "ASYM_KEY_LEN", threshold_crypto::PK_SIZE);
         lang.add_const("ulong", "SIGN_PUBLIC_KEY_LEN", sign::PUBLICKEYBYTES);
```
lionel1704: Not related to this PR but could you please remove these two lines please?
We don't need `sign::*` anymore.

safe_app/src/ffi/crypto.rs
Yoga07: Ah Right! Muddled between `encrypt/decrypt_sealed_box` and `encrypt/decrypt`. Reverting now. Thanks! :)

safe_core/src/crypto.rs
```diff
@@ -71,55 +71,45 @@ pub mod shared_secretbox {
...
     use std::fmt::{self, Debug};
     use std::ops::Deref;
     use std::sync::Arc;
+    use threshold_crypto::{serde_impl::SerdeSecret, SecretKey as BlsSecretKey};
 
     /// Shared secret encryption key.
-    #[derive(Clone, Eq, PartialEq)]
-    pub struct SecretKey(Arc<box_::SecretKey>);
+    #[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+    pub struct SecretKey(Arc<SerdeSecret<BlsSecretKey>>);
```
m-cat: Since `SerdeSecret` already implements everything required including `Send`, `Sync`, `Clone`, `Serialize` and `Deserialize`, I don't think `Arc` is needed here anymore. Is there any reason to keep it?

safe_core/src/crypto.rs
```diff
@@ -71,55 +71,45 @@ pub mod shared_secretbox {
...
     impl Deref for SecretKey {
-        type Target = box_::SecretKey;
+        type Target = threshold_crypto::SecretKey;
 
         fn deref(&self) -> &Self::Target {
-            &*self.0
-        }
-    }
-
-    impl<'de> Deserialize<'de> for SecretKey {
```
m-cat: Always nice to get rid of unneeded code 👍

safe_core/src/ffi/ipc/resp.rs
```diff
@@ -29,7 +29,9 @@ pub struct AppKeys {
     /// Asymmetric enc public key.
     pub enc_pk: AsymPublicKey,
     /// Asymmetric enc private key.
-    pub enc_sk: AsymSecretKey,
+    pub enc_sk: *const u8,
```
m-cat: Similar to Lionel's comment above, can we use an array here? Or is this always serialized? Maybe add a comment clarifying in what form this is being stored here.

safe_core/src/ipc/resp.rs
```diff
@@ -189,12 +190,15 @@ impl ReprC for AppKeys {
...
+        let raw_sk = vec_clone_from_raw_parts((*repr_c).enc_sk, (*repr_c).enc_sk_len);
+        let enc_sk = deserialize(&raw_sk)?;
 
         Ok(Self {
             app_full_id,
             enc_key: shared_secretbox::Key::from_raw(&(*repr_c).enc_key),
-            enc_pk: box_::PublicKey((*repr_c).enc_pk),
-            enc_sk: shared_box::SecretKey::from_raw(&(*repr_c).enc_sk),
+            enc_pk: threshold_crypto::PublicKey::from_bytes((*repr_c).enc_pk)
+                .map_err(|e| format!("threshold_crypto PublicKey conversion error{:?}", e))?,
```
m-cat: I think Lionel introduced a `FromBytes` error in his PR that we can use here!

safe_app/build.rs
```diff
@@ -17,7 +17,7 @@ fn main() {
 #[cfg(feature = "bindings")]
 mod bindings {
     use jni::signature::{JavaType, Primitive};
-    use rust_sodium::crypto::{box_, sign};
+    use rust_sodium::crypto::sign;
```
lionel1704: I think this import can be removed now.

safe_app/src/ffi/crypto.rs
```diff
@@ -275,7 +274,9 @@ pub unsafe extern "C" fn enc_pub_key_new(
     ),
 ) {
     catch_unwind_cb(user_data, o_cb, || {
-        let key = box_::PublicKey(*data);
+        let raw = unwrap!(data.as_ref());
+        let key = threshold_crypto::PublicKey::from_bytes(raw)
```
lionel1704: `threshold_crypto::PublicKey` can be imported to avoid verbosity.

safe_authenticator/build.rs
```diff
@@ -16,7 +16,7 @@ fn main() {
 #[cfg(feature = "bindings")]
 mod bindings {
     use jni::signature::{JavaType, Primitive};
-    use rust_sodium::crypto::{box_, sign};
+    use rust_sodium::crypto::sign;
```
lionel1704: This can be removed too.

safe_core/src/ipc/resp.rs
```diff
@@ -189,12 +190,15 @@ impl ReprC for AppKeys {
...
+        let raw_sk = vec_clone_from_raw_parts((*repr_c).enc_sk, (*repr_c).enc_sk_len);
+        let enc_sk = deserialize(&raw_sk)?;
 
         Ok(Self {
             app_full_id,
             enc_key: shared_secretbox::Key::from_raw(&(*repr_c).enc_key),
-            enc_pk: box_::PublicKey((*repr_c).enc_pk),
-            enc_sk: shared_box::SecretKey::from_raw(&(*repr_c).enc_sk),
+            enc_pk: threshold_crypto::PublicKey::from_bytes((*repr_c).enc_pk)
+                .map_err(IpcError::from)?,
```
lionel1704: I think we don't need this `map_err`
`?` would automatically convert the error into the expected type. Which is defined above as `IpcError`

safe_core/src/ffi/ipc/resp.rs
lionel1704: Looks like the requested comment has not been added.
For clarity we can add a non-doc comment saying,
"Pointer to serialized asymmetric encryption private key"

safe_core/src/ipc/resp.rs
```diff
@@ -635,12 +638,8 @@ mod tests {
...
         );
         assert_eq!(
-            ffi_ak.enc_pk.iter().collect::<Vec<_>>(),
-            enc_pk.0.iter().collect::<Vec<_>>()
-        );
-        assert_eq!(
-            ffi_ak.enc_sk.iter().collect::<Vec<_>>(),
-            enc_sk.0.iter().collect::<Vec<_>>()
+            ffi_ak.enc_pk.iter().cloned().collect::<Vec<_>>(),
+            enc_pk.to_bytes().to_vec()
```
m-cat: Can we test the sk conversion here as well?

safe_app/src/ffi/crypto.rs
```diff
@@ -275,7 +274,8 @@ pub unsafe extern "C" fn enc_pub_key_new(
     ),
 ) {
     catch_unwind_cb(user_data, o_cb, || {
-        let key = box_::PublicKey(*data);
+        let raw = unwrap!(data.as_ref());
```
lionel1704: Looks like an unwrap sneaked in here

safe_core/src/crypto.rs
```diff
@@ -71,55 +71,44 @@ pub mod shared_secretbox {
...
-    #[derive(Clone, Eq, PartialEq)]
-    pub struct SecretKey(Arc<box_::SecretKey>);
+    #[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+    pub struct SecretKey(SerdeSecret<BlsSecretKey>);
 
     impl SecretKey {
         /// Create new safe-to-share key from the given regular key.
-        pub fn new(inner: &box_::SecretKey) -> Self {
-            SecretKey(Arc::new(box_::SecretKey(inner.0)))
+        pub fn new(inner: &BlsSecretKey) -> Self {
```
m-cat: I forgot to write this earlier, but we should just take `BlsSecretKey` by value here, since we just clone it anyway. This will make the caller have to clone it, but it will make it more explicit to them that there is a cost to this function.

safe_app/src/ffi/crypto.rs
```diff
@@ -347,13 +351,25 @@ pub unsafe extern "C" fn enc_secret_key_get(
...
-        sec_enc_key: *const AsymSecretKey,
+        sec_enc_key: *const u8,
+        sec_enc_key_len: usize,
     ),
 ) {
     catch_unwind_cb(user_data, o_cb, || {
-        send_sync(app, user_data, o_cb, move |_, context| {
-            let key = context.object_cache().get_secret_key(handle)?;
-            Ok(&key.0)
+        let user_data = OpaqueCtx(user_data);
```
m-cat: I'm pretty sure we can continue to use `send_sync` here. It *should* support returning two values instead of one, as long as they're in a tuple.

lionel1704: @m-cat there is something funny happening when using `send_sync`.
I had the issue when working with other key types as well. When I use `send_sync` the key serialises and deserialises fine but it fails `assert_eq!`
When a similar failure happened on CI, I asked Yogesh to try using `app.send(..`.
Let's see if this passes now and if it does let's look into the reason why it fails when using `send_sync`. What do you think?

m-cat: Hmm, weird. Could be that the values get dropped after `send_sync`'s call to `f` finishes.

PR #1100, opened 2019-12-02
fix/build: fix cargo build issues and add to GHA

m-cat:
Depends on https://github.com/maidsafe/safe_bindgen/pull/56, which fixes issue with https://github.com/maidsafe/safe_client_libs/pull/1068


lionel1704:
A new version of `safe_bindgen` has been published. Happy to merge this PR once the dependency is updated.

PR #1103, opened 2019-12-05
chore/ci: add macOS to GHA and remove Jenkinsfile

m-cat:
Closes https://github.com/maidsafe/QA/issues/111


.github/workflows/master.yml
```diff
@@ -67,66 +68,84 @@ jobs:
...
 
-  # Put the universal iOS build here when macOS is ready.
-  # It will need to depend on the build-ios job.
+  build-ios-universal:
+    name: Build iOS Universal
+    runs-on: macos-latest
+    needs: build-ios
+    strategy:
+      matrix:
+        target: [x86_64-apple-ios, aarch64-apple-ios]
```
lionel1704: Wouldn't this trigger two separate flows?
All we need here is one flow that downloads the artifacts built in the previous step and combines them

.github/workflows/master.yml
```diff
@@ -148,7 +167,7 @@ jobs:
 
   deploy:
     runs-on: ubuntu-latest
-    needs: [build, build-android]
+    needs: [build, build-android, build-io, build-ios-universal]
```
lionel1704: build-ios*

.github/workflows/pr.yml
```diff
@@ -43,66 +43,72 @@ jobs:
...
 
-  # Put the universal iOS build here when macOS is ready.
-  # It will need to depend on the build-ios job.
+  build-ios-universal:
+    name: Build iOS Universal
+    runs-on: macos-latest
+    needs: build-ios
+    strategy:
+      matrix:
+        target: [x86_64-apple-ios, aarch64-apple-ios]
```
lionel1704: Same comment as above.

PR #1106, opened 2019-12-06
Enable tests in safe_core::client::mock

lionel1704:
@m-cat this PR is a work in progress to re-enable the tests in the `safe_core::client::mock`.

I have refactored the helper functions used for the tests and re-enabled a single test. Let me know what you think about this approach. If it's okay, then I can proceed with refactoring the other tests as well using the same pattern. Thanks!

EDIT: PR is now ready for review

### DIFF
--- a/safe_core/src/client/mock/tests.rs
+++ b/safe_core/src/client/mock/tests.rs
@@ -13,6 +13,7 @@
 use crate::client::mock::vault::Vault;
 use crate::client::{SafeKey, COST_OF_PUT};
 use crate::config_handler::{Config, DevConfig};
+use crate::utils::test_utils::{gen_app_id, gen_client_id};
 use crate::{utils, NetworkEvent};
 
 use super::connection_manager::ConnectionManager;
@@ -77,11 +78,7 @@ fn process_request(
 // Test the basics idata operations.
 #[test]
 fn immutable_data_basics() {
-    let (mut connection_manager, _, client_full_id) = setup();
-
-    // Create account
-    let coins = unwrap!(Coins::from_str("10"));
-    let client_safe_key = register_client(&mut connection_manager, coins, client_full_id);
+    let (mut connection_manager, _, client_safe_key, _) = setup();
 
     // Construct PubImmutableData
     let orig_data: IData =
@@ -113,7 +110,9 @@ fn immutable_data_basics() {
         orig_data
     );
 
-    let balance = unwrap!(coins.checked_sub(*COST_OF_PUT));
+    // Initial balance is 10 coins
+    let balance = unwrap!(Coins::from_str("10"));
+    let balance = unwrap!(balance.checked_sub(*COST_OF_PUT));
     send_req_expect_ok!(
         &mut connection_manager,
         &client_safe_key,
@@ -145,13 +144,7 @@ fn immutable_data_basics() {
 // Test the basic mdata operations.
 #[test]
 fn mutable_data_basics() {
-    let (mut connection_manager, _, client_full_id) = setup();
-
-    // Create account
-    let coins = unwrap!(Coins::from_str("10"));
-    let client_safe_key = register_client(&mut connection_manager, coins, client_full_id);
-
-    let owner_key = client_safe_key.public_key();
+    let (mut connection_manager, _, client_safe_key, owner_key) = setup();
 
     // Construct MutableData
     let name = rand::random();
@@ -407,13 +400,7 @@ fn mutable_data_basics() {
 // Test reclamation of deleted mdata.
 #[test]
 fn mutable_data_reclaim() {
-    let (mut connection_manager, _, client_full_id) = setup();
-
-    // Create account
-    let coins = unwrap!(Coins::from_str("10"));
-    let client_safe_key = register_client(&mut connection_manager, coins, client_full_id);
-
-    let owner_key = client_safe_key.public_key();
+    let (mut connection_manager, _, client_safe_key, owner_key) = setup();
 
     // Construct MutableData
     let name = rand::random();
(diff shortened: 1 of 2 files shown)


safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
+        },
+        ()
+    );
 
     // GetMDataVersion should respond with 0 as the mdata itself hasn't changed.
-    let msg_id = MessageId::new();
-    unwrap!(routing.get_mdata_version(nae_mgr, name, tag, msg_id));
-    let version = expect_success!(routing_rx, msg_id, Response::GetMDataVersion);
-    assert_eq!(version, 0);
+    // GetMDataVersion should respond with 0
```
lionel1704: I will remove this unnecessary comment.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
+            // Deleted entry
+            let entry = entries.get(&key1[..]);
+            assert!(entry.is_none());
+
+            // Inserted entry
+            let entry = unwrap!(entries.get(&key2[..]));
+            assert_eq!(entry.data, value2_v0);
+            assert_eq!(entry.version, 0);
+        }
+        _ => panic!("Unexpected response type"),
```
lionel1704: Some places have "MDataType mismatch". This needs to be consistent.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
+    let response = process_request(
+        &mut connection_manager,
+        &client_safe_key,
+        Request::MutateMDataEntries {
+            address,
+            actions: actions.into(),
+        },
     );
+    match response {
+        Response::Mutation(Err(Error::InvalidEntryActions(_))) => (),
```
lionel1704: Can use macro here.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
```
lionel1704: Use macro here and exact error.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
```
lionel1704: Here too.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,667 @@
...
```
lionel1704: Here too.

safe_core/src/client/mock/tests.rs
lionel1704: Actually we cannot use the macro here. Not without the exact entry action error that are expected to be returned. That will be unnecessarily verbose.

safe_core/src/client/mock/tests.rs
```diff
@@ -1778,60 +1715,86 @@ fn request_hooks() {
...
+// Register a new app for an account with the given permissions.
+// Return the app's safe key and it's connection manager along with the reciever
+// for network events.
+fn register_new_app(
+    conn_manager: &mut ConnectionManager,
+    client_safe_key: &SafeKey,
+    permissions: AppPermissions,
+) -> (SafeKey, ConnectionManager, UnboundedReceiver<NetworkEvent>) {
+    let client_id = unwrap!(client_safe_key.public_id().client_public_id()).clone();
+    let app_full_id = AppFullId::new_bls(&mut thread_rng(), client_id);
```
m-cat: Can use `gen_app_id` here.

safe_core/src/client/mock/tests.rs
```diff
@@ -1778,60 +1715,86 @@ fn request_hooks() {
...
 
-    routing.public_id = PublicId::Client(safe_nd::ClientPublicId::new(owner_key.into(), owner_key));
+// Register a new app for an account with the given permissions.
+// Return the app's safe key and it's connection manager along with the reciever
+// for network events.
+fn register_new_app(
+    conn_manager: &mut ConnectionManager,
+    client_safe_key: &SafeKey,
+    permissions: AppPermissions,
+) -> (SafeKey, ConnectionManager, UnboundedReceiver<NetworkEvent>) {
```
m-cat: I think we also don't need the third return parameter here

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,666 @@
...
-                    Ok(_) => panic!("Unexpected success"),
-                    Err($err) => (),
-                    Err(err) => panic!("Unexpected error {:?}", err),
-                }
-            }
-            event => panic!("Unexpected event {:?}", event),
-        }
+// Helper macro to fetch the response for a request and
+// assert that the expected error is returned.
+macro_rules! send_req_expect_failure {
```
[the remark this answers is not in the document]
lionel1704: The reason why I chose macros instead of functions is because in a function the assertion fails on the same line every time (i.e. inside the helper function). Using a macro would five me which macro call has failed which makes debugging easier.

safe_core/src/client/mock/tests.rs
```diff
@@ -8,518 +8,666 @@
...
+    unwrap!(connection_manager
+        .send(&sender.public_id(), &message)
+        .wait())
 }
 
 // Test the basics idata operations.
 #[test]
 fn immutable_data_basics() {
-    let (mut routing, routing_rx, full_id, _) = setup();
+    let (mut connection_manager, _, client_full_id) = setup();
```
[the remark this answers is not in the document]
lionel1704: > Why have the second return argument if it always gets dropped?
The second argument can be used for the network disconnection listener. If we have a tests that use simulate network disconnect then that receiver can be put to use.

safe_core/src/client/mock/tests.rs
lionel1704: Can be of use in tests that simulate disconnection

PR #1113, opened 2019-12-11
feat/immutable_data: expose a new API which generates a file's data map without committing the chunks to the network

bochaco:


safe_core/src/lib.rs
```diff
@@ -66,6 +66,8 @@ pub mod ipc;
 pub mod nfs;
 /// Implements the Self Encryption storage trait.
 pub mod self_encryption_storage;
+/// Implements the Self Encryption storage trait but for dry run.
+pub mod self_encryption_storage_dry_run;
```
m-cat: I don't think we need a separate module for this, to me it is logically coupled to `self_encryption_storage` and imo belongs in that module

safe_core/src/self_encryption_storage_dry_run.rs
```diff
@@ -0,0 +1,59 @@
...
+        let err = CoreError::Unexpected("Cannot put to storage since it's a dry run.".to_owned());
+        let err = SelfEncryptionStorageError::from(err);
+        Box::new(futures::failed(err))
+    }
+
+    fn generate_address(&self, data: &[u8]) -> Vec<u8> {
+        let idata: IData = if self.published {
+            PubImmutableData::new(data.to_vec()).into()
+        } else {
+            let pk = PublicKey::Bls(SecretKey::random().public_key());
```
m-cat: The address depends on the owner, so if the owner is random then the address is random as well. The data map generated during a dry run would not be the same data map generated in a "real" run.

safe_core/src/self_encryption_storage_dry_run.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    fn put(
+        &mut self,
+        _: Vec<u8>,
+        _data: Vec<u8>,
+    ) -> Box<dyn Future<Item = (), Error = Self::Error>> {
+        trace!("Self encrypt invoked PutIData dry run.");
+        let err = CoreError::Unexpected("Cannot put to storage since it's a dry run.".to_owned());
+        let err = SelfEncryptionStorageError::from(err);
+        Box::new(futures::failed(err))
```
m-cat: I haven't seen this pattern, there is an `err!` macro which we use in SCL instead. Can you see if it works here?

safe_core/src/self_encryption_storage_dry_run.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct SelfEncryptionStorageDryRun {
+    published: bool,
+}
+
+impl SelfEncryptionStorageDryRun {
+    /// Create a new SelfEncryptionStorageDryRun instance.
+    pub fn new(published: bool) -> Self {
+        Self { published }
+    }
+}
```
m-cat: I think this function isn't necessary, and `SelfEncryptionStorageDryRun { published: true }` is more clear than `SelfEncryptionStorage::new(true)`.

safe_core/src/self_encryption_storage_dry_run.rs
bochaco: nice catch!

lionel1704:
The code looks good @bochaco. However I'm afraid that this will not work as expected.
When the self encryptor calls the `write` function, it starts self-encrypting the data and generating the data map while _putting data to the network_ simultaneously.

As chunks are ready to be stored on the network, it calls the `put()` function that is defined for the object that implements the `Storage` trait. In the SEDryRun object the `put()` implementation returns an error so the `gen_data_map` function will always error out.

Could you please write a test to confirm that it always errors out. If I'm right, then the fix is quite simple. We can just write a log saying that put was called for SE dry run and return `ok!(())` instead of an error. It's probably worth keeping a unit test that tests the functionality of the `gen_data_map` function as well.

bochaco:
As agreed @lionel1704 I added a test for `gen_data_map`, it passes fine so it seems it's all good. Please take a look at what it does

safe_core/src/immutable_data.rs
```diff
@@ -180,6 +210,60 @@ mod tests {
...
+            gen_data_map(client, &value.clone(), true, None)
+                .then(move |res| {
+                    let data = unwrap!(res);
+                    let address_before = *data.address();
+                    // attempt to retrieve it with generated address (it should error)
+                    get_value(&client2, address_before, None)
+                        .then(move |res| {
+                            match res {
+                                Ok(_) => panic!("ImmutableData unexpectedly retrieved using address generated by gen_data_map"),
+                                Err(CoreError::DataError(SndError::NoSuchData)) => assert!(true),
```
lionel1704:
```suggestion
                                Err(CoreError::DataError(SndError::NoSuchData)) => (),
```

bochaco:
You were right @lionel1704 , so I made the change to the `put` for SE dry run to simply return ok, and then also made sure we test bigger size files for the gen_data_map test.

bochaco:
I just added tests for gen_data_map on unpublished ImmD and they fail, which is good that I found that issue, I'll have to find out why and fix it. For published ImmD tests are going well now.

lionel1704:
`utils::symmetric_encrypt` takes an optional nonce field. So one way to go about it would be to add the optional `Nonce` field to `gen_data_map` and `immutable_data::create`.

bochaco:
Yes, but that doesn't solve it, end users won't be providing a nonce in many cases, they'll be just requesting CLI to encrypt the files and upload them, so even we could expose that as an option we will still have the same issue if the user doesn't provide the nonce.

lionel1704:
I guess we could have a nonce generated from a user generated string.
But I guess it's not working adding an additional param for the dry-run purpose.

If you want to remove the optional encryption key from `gen_data_map` and support it only for purely self-encrypted data, I'm good with that.

bochaco:
> and support it only for purely self-encrypted data

Did you mean for unencrypted? 

I guess we can leave the option, it's still a dry-run after all, just that the address you get is not deterministic, so to me it sounds like a valid case just that cannot be used by users to check the address of files if they are being encrypted. Perhaps SCL is not the place to limit this since it does what it's expected due to random nonce, but then on CLI we can simply mention it (eventually whenever we support encrypted files) that a dry-run will generate diff addresses for encrypted data

PR #1114, opened 2019-12-12
chore/ci: fix automatic publishing

m-cat:

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -71,6 +71,7 @@ jobs:
           path: artifacts
 
   build-ios:
+    name: Build iOS
     runs-on: macOS-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


m-cat:
I'm going to remove the `publish` job, I only added it temporarily to test publishing. This job is not expected to succeed if safe_core *and* safe_auth or _app are version updated, since auth and app will try to get the new version of safe_core from crates.io instead of the `path` key, which won't exist because it wasn't published for real. This is also why the `test-publish` job only tests safe_core.

> It's a step that runs even for non-version change PRs.

Yeah, sometimes issues come up when running `cargo publish` that didn't come up before, because that command has slightly different dependency resolution than the typical cargo commands. It would be nice to catch these kinds of issues as early as possible. I don't see a reason not to include this job.

m-cat:
The `publish` job should work now, based on the tests I've been running on my fork:

https://github.com/m-cat/safe_client_libs/commit/6665cccd564cfe757c3bb161619a3e13a0c3821a/checks?check_suite_id=356951499

(It still fails, because auth couldn't find the new core version on crates.io)

PR #1115, opened 2019-12-13
Version change: safe_core to 0.37.2

m-cat:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1947,7 +1947,7 @@ dependencies = [
 
 [[package]]
 name = "safe_core"
-version = "0.37.1"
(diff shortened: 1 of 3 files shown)


lionel1704:
@m-cat won't this publish a new version of safe_core alone?

m-cat:
No, it should publish auth 0.12.0 and app 0.12.0 since the latest versions on crates.io are 0.11.0 🙂 

Note that since the previous version publishes failed, core 0.37.0 and 0.37.1 will not be available on crates.io. This shouldn't be an issue since those don't differ from 0.37.2 in any significant way.

PR #1124, opened 2019-12-30
chore/CI: use GHA for Android libs build

ravinderjangra:


.github/workflows/pr.yml
```diff
@@ -184,14 +184,92 @@ jobs:
...
+        shell: bash
+        run:  rustup target add ${{matrix.target}}
+
+    # Build prod native lib.
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --manifest-path=${{ matrix.component }}/Cargo.toml --target=${{ matrix.target }}
+
+      # Upload prod native lib.
```
m-cat: Is there any reason to upload the artifacts in the PR workflow? It doesn't hurt to do so, but I also don't think it's necessary.

ravinderjangra: I am uploading those so that I can try the changes directly from the PRs instead of waiting for the PRs to be merged.

Makefile
```diff
@@ -8,35 +8,6 @@ USER_ID := $(shell id -u)
...
-	@echo "A build target must be specified."
-	@echo "Please set SAFE_CLIENT_LIBS_CONTAINER_TARGET to a valid Rust 'target triple', e.g. 'x86_64-unknown-linux-gnu'."
-	@exit 1
-endif
-	docker push \
-		maidsafe/safe-client-libs-build:${SAFE_CLIENT_LIBS_CONTAINER_TARGET}-${SAFE_CLIENT_LIBS_CONTAINER_TYPE}
-
 build:
 	rm -rf artifacts
 ifeq ($(UNAME_S),Linux)
```
m-cat: Can the `build-with-container` stuff be removed also?

ravinderjangra: yep, removing now

m-cat: There are still lines in the Makefile that call these container scripts, I think they can be removed as well 🙂

ravinderjangra:
@m-cat, I think there are lots of files in scripts that can be removed/deleted. I think once both the refactoring PRs are merged, we can raise another PR to remove all unnecessary files/ci related code.

.github/workflows/master.yml
```diff
@@ -200,18 +200,92 @@ jobs:
...
+      - name: Cache cargo build
+        uses: actions/cache@v1
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+
+      # Check if NDK bundle is installed and list files.
+      - name: Check Android NDK bundle
+        shell: bash
+        run: ls /usr/local/lib/android/sdk/ndk-bundle
```
lionel1704: What happens if this is not available at this location?
Will this step fail?

.github/workflows/master.yml
```diff
@@ -200,18 +200,92 @@ jobs:
...
+        run: |
+          mkdir toolchains
+          /usr/local/lib/android/sdk/ndk-bundle/build/tools/make_standalone_toolchain.py --arch x86_64 --api 21 --install-dir toolchains --force
+          echo "::set-env name=CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER::x86_64-linux-android-gcc"
+          echo "::add-path::$PWD/toolchains/bin"
+
+      # Verify installed NDK toolchain for the target.
+      - name: Check Android NDK toolchain Files
+        shell: bash
+        run: ls toolchains
```
lionel1704: Is this the best way to verify the toolchain?
Since we have run `mkdir toolchains` this step will never fail

.github/workflows/master.yml
```diff
@@ -200,18 +200,92 @@ jobs:
...
+          command: build
+          args: >
+            --release --manifest-path=${{ matrix.component }}/Cargo.toml --target=${{ matrix.target }}
+            --features=mock-network
+
+      # Upload dev native lib.
+      - uses: actions/upload-artifact@master
+        with:
+          name: ${{ matrix.component }}-${{ matrix.target }}-dev
+          path: target/${{ matrix.target }}/release/lib${{matrix.component}}.so
```
lionel1704: These seem a little repititive. Can we add prod and dev to the build matrix?

.github/workflows/pr.yml
```diff
@@ -184,14 +184,92 @@ jobs:
...
+      - name: Cache cargo build
+        uses: actions/cache@v1
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+
+      # Check if NDK bundle is installed and list files.
+      - name: Check Android NDK bundle
+        shell: bash
+        run: ls /usr/local/lib/android/sdk/ndk-bundle
```
lionel1704: Comment in master flow.

.github/workflows/pr.yml
```diff
@@ -184,14 +184,92 @@ jobs:
...
+        run: |
+          mkdir toolchains
+          /usr/local/lib/android/sdk/ndk-bundle/build/tools/make_standalone_toolchain.py --arch x86_64 --api 21 --install-dir toolchains --force
+          echo "::set-env name=CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER::x86_64-linux-android-gcc"
+          echo "::add-path::$PWD/toolchains/bin"
+
+      # Verify installed NDK toolchain for the target.
+      - name: Check Android NDK toolchain Files
+        shell: bash
+        run: ls toolchains
```
lionel1704: Comment in master flow.

.github/workflows/pr.yml
```diff
@@ -184,14 +184,92 @@ jobs:
...
+          command: build
+          args: >
+            --release --manifest-path=${{ matrix.component }}/Cargo.toml --target=${{ matrix.target }}
+            --features=mock-network
+
+      # Upload dev native lib.
+      - uses: actions/upload-artifact@master
+        with:
+          name: ${{ matrix.component }}-${{ matrix.target }}-dev
+          path: target/${{ matrix.target }}/release/lib${{matrix.component}}.so
```
lionel1704: Same comment about repetition.

.github/workflows/master.yml
ravinderjangra: I haven't found any alternative to this yet. maybe we can use something like find.
```
find ./ndk-default-loc -name ndk-bundle 
```
we can combine that with some kind of count

.github/workflows/master.yml
ravinderjangra: Ya comment is wrong here, actually, I was using this to list the files form the toolchains dir. We can remove this step completely

.github/workflows/master.yml
ravinderjangra: The no of jobs will increase in the case of a matrix and then it may go in the queue.  
Currently there are 4 ([safe_app, safe_authenticator] x 2 arch) parallel tasks which will be doubled.

.github/workflows/master.yml
ravinderjangra: I can add a check to see if ndk is installed or not in https://github.com/ravinderjangra/android-ndk-toolchain-setup and use that everywhere or just add the check directly for now.

lionel1704: That might be unnecessary. I remember you mentioning that the ndk is installed by default for all GH actions machines. If that's the case, then we can skip this step

ravinderjangra: yep, that's correct.

.github/workflows/master.yml
```diff
@@ -159,18 +159,87 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
         with:
-          name: safe_client_libs-${{ matrix.target }}-${{ matrix.type }}
-          path: artifacts
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+      
+      # Set environment variable for the feature flag.
```
lionel1704: This comment needs to be updated.

.github/workflows/master.yml
```diff
@@ -159,18 +159,87 @@ jobs:
...
+          /usr/local/lib/android/sdk/ndk-bundle/build/tools/make_standalone_toolchain.py --arch x86_64 --api 21 --install-dir toolchains --force
+          echo "::set-env name=CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER::x86_64-linux-android-gcc"
+          echo "::add-path::$PWD/toolchains/bin"
+
+    # Build safe_app native lib.
+      - name: build safe_app
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --manifest-path=safe_app/Cargo.toml --target=${{ matrix.target }} ${{ steps.set_dev_flag.features.docker_tag }}
```
lionel1704: Why does this say `docker_tag` ?

.github/workflows/pr.yml
```diff
@@ -147,14 +147,87 @@ jobs:
...
-      - shell: bash
-        run: make build-android
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+      
+      # Set environment variable for the feature flag.
```
lionel1704: This comment too.

.github/workflows/pr.yml
```diff
@@ -147,14 +147,87 @@ jobs:
...
+          /usr/local/lib/android/sdk/ndk-bundle/build/tools/make_standalone_toolchain.py --arch x86_64 --api 21 --install-dir toolchains --force
+          echo "::set-env name=CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER::x86_64-linux-android-gcc"
+          echo "::add-path::$PWD/toolchains/bin"
+
+    # Build safe_app native lib.
+      - name: build safe_app
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --manifest-path=safe_app/Cargo.toml --target=${{ matrix.target }} ${{ steps.set_dev_flag.outputs.features }}
```
lionel1704: Seems to be correct here.

PR #1125, opened 2019-12-30
chore/ci: use cargo lipo to generate universal iOS fat libs

ravinderjangra:

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -95,7 +95,7 @@ jobs:
     runs-on: macOS-latest
     strategy:
       matrix:
-        target: [aarch64-apple-ios, x86_64-apple-ios]
+        component: [safe_app, safe_authenticator]
     steps:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -216,7 +175,7 @@ jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    needs: [build, build-android, build-ios, build-ios-universal]
+    needs: [build, build-android, build-ios, build-ios]
```
m-cat: Looks like we have `build-ios` twice here

.github/workflows/master.yml
```diff
@@ -120,79 +127,31 @@ jobs:
...
-
-      # Build dev libraries.
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: >
-            --release --manifest-path=safe_authenticator/Cargo.toml
-            --features=mock-network --target=${{ matrix.target }}
+      
+      # Install cargo-lipe for universal iOS libs.
```
m-cat:
```suggestion
      # Install cargo-lipo for universal iOS libs.
```

.github/workflows/pr.yml
```diff
@@ -108,75 +115,31 @@ jobs:
         with:
           path: target
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
-
-      # Build prod libraries.
+      
+      # Install cargo-lipe for universal iOS libs.
```
m-cat:
```suggestion
      # Install cargo-lipo for universal iOS libs.
```

.github/workflows/master.yml
ravinderjangra: will update this

.github/workflows/master.yml
ravinderjangra: will update this

PR #1126, opened 2020-01-02
refactor/ffi: make RegisteredApp struct include AppPermissions

### LINKED ISSUE #1123: Refactor RegisteredApp struct to hold AppPermissions
It is desired to return the an application's permissions with [`auth_registered_apps()` API](https://github.com/maidsafe/safe_client_libs/blob/1ab3fb3fba3f1080e221f6757a166233538c636f/safe_authenticator/src/ffi/apps.rs#L118) in `safe_authenticator`. This can be done by making [`RegisteredApp`](https://github.com/maidsafe/safe_client_libs/blob/1ab3fb3fba3f1080e221f6757a166233538c636f/safe_authenticator/src/ffi/apps.rs#L29) to hold the `AppPermissions` struct. So that when returning RegisteredApp via the API, it can hold the Apps details along with it's respective permissions.

Yoga07:
Resolves #1123

### DIFF
--- a/safe_authenticator/src/tests/mod.rs
+++ b/safe_authenticator/src/tests/mod.rs
@@ -34,6 +34,7 @@ use futures::{future, Future};
 use safe_core::config_handler::Config;
 use safe_core::ffi::error_codes::ERR_NO_SUCH_CONTAINER;
 use safe_core::{app_container_name, mdata_info, AuthActions, Client};
+use safe_nd::AppPermissions;
 use std::collections::HashMap;
 use std::ffi::CString;
 use std::sync::mpsc;
@@ -889,15 +890,20 @@ fn containers_access_request() {
     );
 }
 
-struct RegisteredAppId(String);
+struct RegisteredAppId {
+    id: String,
+    perms: FfiAppPermissions,
+}
+
 impl ReprC for RegisteredAppId {
     type C = *const RegisteredApp;
     type Error = StringError;
 
     unsafe fn clone_from_repr_c(repr_c: Self::C) -> Result<Self, Self::Error> {
-        Ok(RegisteredAppId(String::clone_from_repr_c(
-            (*repr_c).app_info.id,
-        )?))
+        Ok(RegisteredAppId {
+            id: String::clone_from_repr_c((*repr_c).app_info.id)?,
+            perms: (*repr_c).app_perms,
+        })
     }
 }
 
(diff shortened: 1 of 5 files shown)


safe_authenticator/src/tests/mod.rs
```diff
@@ -1009,6 +1011,62 @@ fn lists_of_registered_and_revoked_apps() {
...
+fn test_registered_apps() {
+    let authenticator = test_utils::create_account_and_login();
+
+    let app1_perms = AppPermissions {
+        transfer_coins: true,
+        perform_mutations: true,
+        get_balance: true,
+    };
+
+    // Register two apps.
```
m-cat: Should this comment be above `let app1_perms ...`?

safe_authenticator/src/tests/mod.rs
```diff
@@ -887,15 +888,16 @@ fn containers_access_request() {
     );
 }
 
-struct RegisteredAppId(String);
+struct RegisteredAppId(String, FfiAppPermissions);
```
m-cat: I think this should now be a struct with named fields instead of a tuple struct. It is a bit confusing when referring to the `.1` field.

safe_authenticator/src/tests/mod.rs
```diff
@@ -1009,6 +1011,62 @@ fn lists_of_registered_and_revoked_apps() {
...
+        app: test_utils::rand_app(),
+        app_container: false,
+        app_permissions: app2_perms,
+        containers: Default::default(),
+    };
+
+    let _ = unwrap!(test_utils::register_app(&authenticator, &auth_req1));
+    let _ = unwrap!(test_utils::register_app(&authenticator, &auth_req2));
+
+    // There are now two registered apps.
```
m-cat: Can we add an assertion to verify that this comment is true?

safe_authenticator/src/tests/mod.rs
```diff
@@ -1009,6 +1011,62 @@ fn lists_of_registered_and_revoked_apps() {
...
+    let mut registered: Vec<RegisteredAppId> = unsafe {
+        unwrap!(call_vec(|ud, cb| auth_registered_apps(
+            &authenticator,
+            ud,
+            cb
+        ),))
+    };
+
+    // Assert App1's Permissions
+    let app1: RegisteredAppId = unwrap!(registered.pop());
```
m-cat: I'm a bit confused, shouldn't `app1` be at `registered[0]` instead of at the end of `registered`? Maybe use indices instead of `pop`, we can also get rid of an `unwrap` that way.

Yoga07: The order of fetching the list of the apps seems to be completely random @m-cat . I found this while addressing this comment. Sometimes `app1` randomly seems to be at the end of the list(ie.`registered[1]`) and sometimes at the begining. Not sure why this behavior tho.

m-cat: I suspected that might be the case. The test needs to handle this randomness accordingly, i.e. detect which app is in which position based on the ID.

safe_authenticator/src/tests/mod.rs
```diff
@@ -889,15 +890,20 @@ fn containers_access_request() {
     );
 }
 
-struct RegisteredAppId(String);
+struct RegisteredAppId {
+    _id: String,
```
lionel1704: Why is this unused?

safe_authenticator/src/tests/mod.rs
```diff
@@ -1011,6 +1017,70 @@ fn lists_of_registered_and_revoked_apps() {
...
+        unwrap!(call_vec(|ud, cb| auth_registered_apps(
+            &authenticator,
+            ud,
+            cb
+        ),))
+    };
+    assert_eq!(registered.len(), 2);
+
+    for app in registered {
+        if app1_id == app._id {
```
lionel1704: Doesn't look like it's unused :) 
I'm guessing clippy is failing because it isn't used outside of tests?

Yoga07: Ah yes! The `id` field was unused before the comment the was addressed(It was a tuple struct previously, and hence didn't throw any errors). Now that `app.id` is being used, we can get rid of the underscore.

PR #1131, opened 2020-01-06
Refactoring

joshuef:
Some changes to code location and naming tweaks to clarify things (as per my understanding)


safe_authenticator/src/access_container.rs
```diff
@@ -13,11 +13,13 @@
...
 use crate::client::AuthClient;
 use bincode::{deserialize, serialize};
+use futures::future;
 use futures::Future;
+use safe_core::ipc::req::{container_perms_into_permission_set, ContainerPermissions};
 use safe_core::ipc::resp::{access_container_enc_key, AccessContainerEntry};
 use safe_core::ipc::AppKeys;
 use safe_core::utils::{symmetric_decrypt, symmetric_encrypt, SymEncKey};
-use safe_core::{recovery, Client, CoreError, FutureExt, MDataInfo};
+use safe_core::{mutable_data, Client, CoreError, FutureExt, MDataInfo};
```
Yoga07: This renaming can be left unchanged. The `recovery` module contains a special set of functions which are wrappers over their corresponding Mutable Data APIs. These wrappers functions are written in a specific way to handle and recover from the failures that arise from the Mutable Data APIs over which they are wrapped. Hence the name `recovery`. _Only_ the authenticator makes use of this module, specifically on access_container, authentication and revocation operations so that MData updates may not fail crashing the authenticator. So this cannot be termed generally as a `mutable_data` module.

safe_core/src/client/client_handler.rs
```diff
@@ -0,0 +1,56 @@
...
+use futures::Future;
+use safe_nd::{AppPermissions, Error as SndError, PublicKey};
+
+const MAX_ATTEMPTS: usize = 10;
+
+///! Client Handler functionality
+
+/// Insert key to Client Handler.
+/// Covers the `InvalidSuccessor` error case (it should not fail if the key already exists).
+pub fn ins_auth_key(
```
Yoga07: This as well is a recovery function that wraps the `ins_auth_key` with a error handling wrapper. I think it's best to have all the recovery functions at one place, in the `recovery` module.

safe_core/src/client/mod.rs
```diff
@@ -8,13 +8,15 @@
...
+pub mod client_handler;
 /// Core client used for testing purposes.
 #[cfg(any(test, feature = "testing"))]
 pub mod core_client;
 /// `MDataInfo` utilities.
 pub mod mdata_info;
-/// Operations with recovery.
-pub mod recovery;
+/// Operations for mutable_data.
+pub mod mutable_data;
```
Yoga07: This can be reverted once the `recovery` module is brought back. As mentioned above, `recovery` module is not a generic mutable_data module :)

safe_core/src/client/client_handler.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::errors::CoreError;
+use crate::event_loop::CoreFuture;
+use crate::utils::FutureExt;
+use futures::future::{self, Loop};
+use futures::Future;
+use safe_nd::{AppPermissions, Error as SndError, PublicKey};
+
+const MAX_ATTEMPTS: usize = 10;
+
+///! Client Handler functionality
```
lionel1704: I don't feel that client_handler needs to be a separate module. It's not something that the _Client_ needs to be aware of.
If we add a module to handle all app key operations, then we might need to handle all the other operations as well (delete, list all keys)

safe_core/src/client/mod.rs
lionel1704: The `recovery` module is meant for SCL internals (eg. Authenticator logic) to recover from the errors that are likely to happen. While renaming it to mutable data does make sense, the original purpose of the module is a little different.

safe_authenticator/src/access_container.rs
joshuef: I think `recovery` is a pretty unhelpful name. It's not intuitive at all. Recovery of what? This is in effect a bunch of MD operations for the auth. So it should be named as such, IMO. (TBH, should the MD functions themselves not do this? Why do we have/need this wrapper?)

safe_authenticator/src/access_container.rs
```diff
@@ -26,6 +28,56 @@ use std::collections::HashMap;
 /// Key of the authenticator entry in the access container.
 pub const AUTHENTICATOR_ENTRY: &str = "authenticator";
 
+/// Updates containers permissions (adds a given key to the permissions set)
+#[allow(clippy::implicit_hasher)]
+pub fn update_container_perms(
```
lionel1704: This change makes a lot of sense! :+1:

safe_core/src/client/client_handler.rs
joshuef: Again, have to disagree here. I spent a long time reading around the code to figure out just what `ins_auth_key` was doing, and being part of quite a vague `recovery` module was part of this. The functionality should live where it's relevant. If we're wrapping something, then so be it. But I don't think it needs to live in a separate module.

safe_core/src/client/mod.rs
joshuef: Hmm, is that not functionality we should be baking in to the APIs then?

safe_core/src/client/client_handler.rs
joshuef: This is in core not client... The client handler exists as it's own node type. It doesn't make sense for this operation to be shoe-horned in to some other module either IMO. That just leads to confusion...

safe_core/src/client/mod.rs
joshuef: atm, i'm rolling with `recovery_wrapped_apis` in response to the above, though it feels overly verbose, i'm not sure what else to call it...

lionel1704: > is that not functionality we should be baking in to the APIs then?
Not necessarily. Apps should handle versions themselves. Otherwise I think it beats it's purpose. This is just the authenticator's way of handling incorrect versions.

lionel1704: Does `recoverable` or  make more sense?
We can document it as "Wrapper functions to the client API that can recover from expected network failures."

joshuef: `recoverable_apis` @lionel1704 ?

safe_core/src/client/mod.rs
```diff
@@ -13,8 +13,8 @@ pub mod account;
 pub mod core_client;
 /// `MDataInfo` utilities.
 pub mod mdata_info;
-/// Operations with recovery.
-pub mod recovery;
+/// Varias APIs wrapped to provide resiliance for common network operaionts.
```
lionel1704: *various *operations

safe_core/src/client/recoverable_apis.rs
```diff
@@ -18,10 +18,14 @@ use safe_nd::{
     MDataSeqEntries, MDataSeqEntryAction, MDataSeqEntryActions, MDataSeqValue, PublicKey,
     SeqMutableData,
 };
+// use safe_nd::{AppPermissions, Error as SndError, PublicKey};
```
lionel1704: Can be removed

safe_core/src/client/mod.rs
```diff
@@ -13,8 +13,8 @@ pub mod account;
...
+pub mod recoverable_apis;
```
lionel1704: I think it should be `recoverable_api` instead?

joshuef: Hmmmmmm. Not sure. `The API` is _all of them_. You can have a few APIs... But not a few API.

I still lean towards plural here, IMO.

lionel1704:
Happy to merge this once rebased :) 
There's one more thing too @joshuef 
The format for the commit messages that we follow are a little different. It'd be nice if you could reword the commits using interactive rebase.
Here are the [commit guidelines](https://github.com/autumnai/leaf/blob/master/CONTRIBUTING.md#git-commit-guidelines) that we follow.

PR #1133, opened 2020-01-06
chore/ffi: make safe_app FFI Error public

bochaco:


safe_app/src/lib.rs
```diff
@@ -95,7 +95,7 @@ mod errors;
 mod tests;
 
 use self::object_cache::ObjectCache;
-use crate::ffi::errors::{Error, Result as FfiResult};
+pub use crate::ffi::errors::{Error as FfiError, Result as FfiResult};
```
m-cat: Actually, these don't need to be publicly exported here, since they are accessible through the `ffi::errors` module already, which is public. I feel like FFI exports should be contained in the FFI module only, so we don't mix Native and FFI stuff and don't have to use the `Ffi` prefix here.

bochaco: yes, it makes sense @m-cat , I'll do that change

m-cat:
Looks good, just waiting on:

> Can you make the errors module pub in safe_app/src/lib.rs, since it is public in safe_auth?

Minor thing and mostly just for consistency 🙂

bochaco:
ah yeah, sorry, pushing it now

PR #1136, opened 2020-01-07
chore/all: add documentation for export strategy

### LINKED ISSUE #1065: Provide better documentation for our module and export strategy
As can be seen in [this PR](https://github.com/maidsafe/safe_client_libs/pull/1063), our export and re-export strategy in the `safe_auth` and `_app` libs are poorly documented or not documented at all, which leads to oversight such as not exporting FFI error codes to FFI callers. It also is not very clear to our contributors why our module and export structure is the way it is.

We should write documentation based on the comments I left in that PR. We can mostly copy those comments and dress them up a little, there is not much new stuff to write. We should leave at least some minimal documentation in the code, and we should create a wiki page with more complete documentation.

m-cat:
Closes #1065


nbaksalyar:
Superseded by #1144

PR #1138, opened 2020-01-07
chore/ci: remove --verbose flag from scripts

m-cat:
The --verbose flag is not needed in CI since we set `CARGO_TERM_VERBOSE`.

This makes the output less cluttered when running the scripts locally.


lionel1704:
This commit has been merged to master in #1137

PR #1139, opened 2020-01-08
chore/all: remove macro_use in safe_auth and safe_app

m-cat:


m-cat:
Waiting on an FFI Utils version change.

PR #1141, opened 2020-01-10
feat/examples: update and revive email.rs examples

### LINKED ISSUE #772: Restore deleted examples and use latest APIs
We previously deleted the files for examples that had been commented out, namely email.rs and email_stress_test.rs. These files should be restored from version control, uncommented and reworked so that they function correctly.

One issue here though is that we need to create an app client through registered_with_keys, i.e. the app needs to be authenticated first. So we either need to use system_uri in the examples, or run them on the mock-routing version only.

Also get rid of docopt and rustc-serialize for command line args parsing and completely move to clap + serde in the examples.

Yoga07:
Rebased on top of #1156 
- Updates and revives email.rs example - real-network only
- Also restores the email_stress_test.rs example which is yet to be updated

### DIFF
--- a/safe_app/Cargo.toml
+++ b/safe_app/Cargo.toml
@@ -61,3 +61,7 @@ name = "client_stress_test"
 [[example]]
 bench = false
 name = "self_authentication"
+
+[[example]]
+bench = false
+name = "email"
--- a/safe_app/examples/email.rs
+++ b/safe_app/examples/email.rs
@@ -1,0 +1,454 @@
+// Copyright 2016 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under (1) the MaidSafe.net
+// Commercial License, version 1.0 or later, or (2) The General Public License
+// (GPL), version 3, depending on which licence you accepted on initial access
+// to the Software (the "Licences").
+//
+// By contributing code to the SAFE Network Software, or to this project
+// generally, you agree to be bound by the terms of the MaidSafe Contributor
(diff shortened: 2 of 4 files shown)


safe_app/examples/email.rs
```diff
@@ -0,0 +1,451 @@
...
+// For explanation of lint checks, run `rustc -W help`.
+#![warn(
+    missing_docs,
+    trivial_casts,
+    trivial_numeric_casts,
+    unused_extern_crates,
+    unused_import_braces,
+    unused_qualifications,
+    unused_results
+)]
```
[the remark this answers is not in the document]
Yoga07: `safe_core`'s lint rules are used except for `#![deny(unsafe_code)]`(as we have unsafe ffi blocks in this example). And I think safe_app and safe_auth follow the same as safe_core's.

safe_app/Cargo.toml
```diff
@@ -61,3 +61,8 @@ name = "client_stress_test"
 [[example]]
 bench = false
 name = "self_authentication"
+required-features = ["mock-network"]
```
lionel1704: Why is this a required feature?

Yoga07: Ah, that shouldn't be there. Wrote it to test a few things. Removing it now. Thanks!

lionel1704:
The email example will make more sense once we have the "Sequence" and "Map" types implemented.
Closing this for now.

PR #1154, opened 2020-01-22
fix/core: send notifications via connection manager during network disconnection

### LINKED ISSUE #957: Send network events from the ConnectionManager
Currently, the `net_tx` passed down to the ConnectionManager is not being used. It should send proper notifications when we have a network disconnection.

It also needs to be simulated in the mock ConnectionManager.

Yoga07:
Resolves #957

### DIFF
--- a/safe_core/src/connection_manager.rs
+++ b/safe_core/src/connection_manager.rs
@@ -100,7 +100,8 @@ impl Inner {
             let _ = value.insert(fry!(ConnectionGroup::new(
                 self.config.clone(),
                 full_id,
-                connected_tx
+                connected_tx,
+                self.net_tx.clone(),
             )));
             Box::new(
                 connected_rx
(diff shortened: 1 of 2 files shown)


Yoga07:
>I don't think this approach is correct.

https://github.com/maidsafe/safe_client_libs/pull/1154/files#diff-4b03f71d31c8324f867afdd9ae56c9eaR602
That function is executed when there is a network connection failure, so, this is from where we want to send a notification to the client's event loop too right?

>The disconnect() functions disconnects from a particular section.

I agree that we don't want this:
https://github.com/maidsafe/safe_client_libs/pull/1154/files#diff-52efc5c5ea7af3699f9a5845ed2348fbR149
since `disconnect()` disconnects only from a section, and not from the network.

lionel1704:
> That function is executed when there is a network connection failure

Well not exactly. The quic-p2p error variant says it's a `ConnectionCancelled` event. That's not for a _internet access_ issue, is it?

Yoga07:
>That's not for a internet access issue, is it?

Connection get's cancelled when there is no internet connection, or if the node connected already refuses to connect I guess.

https://github.com/maidsafe/quic-p2p/blob/c8fc50a0270ea120ffc7d24f728a7f436e9ecc80/src/error.rs#L89
QuicP2P doc doesn't give much insight on this.

PR #1167, opened 2020-02-18
Update dependencies

lionel1704:
Update safe-nd and quic-p2p dependencies. This PR makes SCL compatible with the latest version of safe_vault and routing for Fleming.

### DIFF
--- a/safe_core/src/immutable_data.rs
+++ b/safe_core/src/immutable_data.rs
@@ -214,11 +214,11 @@ mod tests {
     }
 
     // Test creating and retrieving a 10mb idata.
-    #[cfg(not(debug_assertions))]
-    #[test]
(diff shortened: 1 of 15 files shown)


safe_core/src/immutable_data.rs
```diff
@@ -214,11 +214,11 @@ mod tests {
     }
 
     // Test creating and retrieving a 10mb idata.
-    #[cfg(not(debug_assertions))]
-    #[test]
-    fn create_and_retrieve_10mb() {
-        create_and_retrieve(10 * 1024 * 1024)
-    }
+    // #[cfg(not(debug_assertions))]
```
bochaco: Shouldn't we remove this if we don't want this test anymore? or enable it otherwise

PR #1168, opened 2020-02-19
Fix mobile builds

lionel1704:
This is a test PR to build the mobile libraries so that they can be tested against a section.
Do not merge this.


ravinderjangra:
Both native libs (Android, iOS) are working with the locally running single section vault.

lionel1704:
A separate branch has been created for this

PR #1169, opened 2020-02-24
add postition and index to get_value

ravinderjangra:


ravinderjangra:
I'll update in the morning and will try adding some tests as well.

PR #1174, opened 2020-03-03
Disable Android and iOS builds on CI

lionel1704:


lionel1704:
Merging without CI since the changes are only related to GH actions.

PR #1182, opened 2020-03-24
Bye bye safe_app ffi

joshuef:


safe_app/src/tests/mod.rs
```diff
@@ -11,15 +11,13 @@ mod append_only_data;
 mod coins;
 mod unpublished_mutable_data;
 
-use crate::ffi::test_utils::test_create_app_with_access;
 use crate::test_utils::{create_app, create_random_auth_req, gen_app_exchange_info};
-use crate::test_utils::{create_app_by_req, create_auth_req, create_auth_req_with_access};
+use crate::test_utils::{create_app_by_req, create_auth_req_with_access};
 use crate::{run, App, AppError};
-use ffi_utils::test_utils::call_1;
+// use ffi_utils::test_utils::call_1;
```
bochaco: Shall we simply remove it?

safe_app/src/object_cache.rs
```diff
@@ -25,6 +24,47 @@ use std::cell::{Cell, RefCell, RefMut};
...
+pub type SignPubKeyHandle = ObjectHandle;
+/// Disambiguating `ObjectHandle`
+pub type PubKeyHandle = ObjectHandle;
+/// Disambiguating `ObjectHandle`
+pub type SignSecKeyHandle = ObjectHandle;
+/// Disambiguating `ObjectHandle`
+pub type FileContextHandle = ObjectHandle;
+
+/// Holds context for file operations, depending on the mode.
+pub struct FileContext {
```
bochaco: I think this is for NFS emulation, shouldn't we remove the NFS completely an therefore this struct?

safe_app/src/object_cache.rs
```diff
@@ -280,23 +320,3 @@ impl<V> Store<V> {
         self.inner.borrow_mut().clear()
     }
 }
-
-#[cfg(test)]
```
bochaco: AFAIK, the object cache was only used by FFI, if we are removing FFI can we remove object cache too? otherwise, I'm just wondering why we are removing this test

safe_app/src/tests/mod.rs
```diff
@@ -73,24 +71,67 @@ fn refresh_access_info() {
 
 // Test fetching containers that an app has access to.
 #[test]
-#[allow(unsafe_code)]
+#[cfg(feature = "mock-network")]
+// #[allow(unsafe_code)]
```
bochaco: It seems this test still needs a refactoring? or why we are now running it only with mock?

safe_app/src/object_cache.rs
lionel1704: I second that. Object cache was used only for FFI. It can be removed.

safe_app/src/object_cache.rs
joshuef: I'd say removing NFS is a separate thing to FFI at this point.

Happy to do it, but in another PR I'd say?

safe_app/src/object_cache.rs
joshuef: Grand. Wasn't sure so I left it in. Will get it oooot

tests/src/lib.rs
```diff
@@ -21,8 +21,6 @@
     unused_results
 )]
 
-mod real_network;
```
lionel1704: Any particular reason why this is removed @joshuef ?

joshuef: everything in that file was ignored. It was a rabbit hole of refactoring tests away from ffi (which, i'm not sure was worthwhile, given how they were all ignored)

lionel1704: Alright. No worries. Let get this merged then :+1:

PR #1193, opened 2020-04-20
test/response_man: Ensure failed tests are recorgnised with response …

joshuef:
…manager

### DIFF
--- a/safe_core/src/connection_manager/response_manager.rs
+++ b/safe_core/src/connection_manager/response_manager.rs
@@ -153,6 +153,41 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    #[should_panic]
+    fn response_manager_failed_assert_is_caught_in_response() {
+        // This test is to ensure that our _other_ tests are runnning properly and would actually capture
+        // a failure in responses.
+
+        let response_threshold = 1;
+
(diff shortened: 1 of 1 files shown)


safe_core/src/connection_manager/response_manager.rs
```diff
@@ -153,6 +153,41 @@ mod tests {
...
+        let bad_immutable_data_response =
+            safe_nd::Response::GetIData(Ok(safe_nd::IData::from(bad_immutable_data)));
+
+        let _ = response_manager.await_responses(message_id, (sender_future, expected_responses));
+        let _ = response_manager.handle_response(message_id, response);
+
+        let _ = response_future
+            .map(move |i| {
+                // actual received response and bad response are different. So test should panic.
+                assert_eq!(&i, &bad_immutable_data_response);
```
lionel1704, line 186: If we expect this to panic then it's better to use `assert_ne!` to and remove `#[should_panic]`.
That way other panics will be caught.

joshuef: The point of the test is to check that a panic from within that future + map reaches outside.

Ie, that a failing assertion will result in a failed test. (As working on this, I was down a path where for _some time_ failing tests were not being recorded due to the setup).

So maybe I need to update the test name to make that clearer?

PR #1201, opened 2020-05-15
refactor/safe_core: break down the mock vault's request processor into diff mods

bochaco:


bochaco:
This will be covered with https://github.com/maidsafe/safe-client-libs/pull/1204

PR #1203, opened 2020-05-18
chore(ci): switch CI to generate dep graph in png_generator branch

S-Coyle:
We are seeing failures when trying to commit changes via CI to master branch because it is a protected branch - if we switch to a non-protected branch then commits should work as expected.


README.md
```diff
@@ -21,7 +21,7 @@ This is the project workspace. Please refer to individual members for details:
 
 Crate graph ([explanation](https://github.com/maidsafe/safe_client_libs/wiki#crate-graph)):
 
-![safe_app dependencies](safe-client-libs.png)
+[[https://github.com/maidsafe/safe_client_libs/blob/png_generator/safe-client-libs.png|alt=safe_app dependencies]]
```
bochaco, line 24: The image doesn't seem to exist yet, but apart from that, is it the correct format to render it? I don't see it properly rendered here: https://github.com/maidsafe/safe_client_libs/blob/70898620247251bac0f8b5e1c111357a16e6c3ce/README.md

S-Coyle: Thanks for the review! Yes the image doesn't exist yet at that URL location, hence why it's not showing yet. CI will generate the image on push to master, i.e. when this PR is merged, then it should display in the readme.

PR #1208, opened 2020-06-03
fix/auth-ffi: some safe_authenticator ffi functions were not returning when used with real network

bochaco:


bochaco:
Closing as FFI API in safe_client_libs will be deprecated so this won't be needed.

PR #1224, opened 2020-07-08
Gather code coverage stats

S-Coyle:
At the time of writing there are multiple failures for the SCL tests when run with the latest vault, cli and authd.

Ideally any failures would be investigated and fixed before this PR is merged, otherwise CI will fail for every PR until they are fixed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -8,6 +8,10 @@ env:
   RUST_BACKTRACE: 1
   # Deny all compiler warnings.
   RUSTFLAGS: "-D warnings"
+  # set the vault, cli and authd version to be downloaded and used in the E2E against real network job
+  SAFE_VAULT_VERSION: "latest"
+  SAFE_CLI_VERSION: "latest"
+  SAFE_AUTH_VERSION: "latest"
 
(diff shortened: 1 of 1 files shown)


S-Coyle:
This PR relies on there being a file name containing `latest` on S3 for safe-vault & safe-cli. I've added these manually for now. Separate PRs to auto generate each of these with every release are raised as per below, but don't need to be merged before this PR:
* [vault](https://github.com/maidsafe/safe-vault/pull/1021)
* [cli](https://github.com/maidsafe/safe-api/pull/592)

S-Coyle:
Note that I removed a commit from this PR which refactored the CI files, moving it to a separate PR #1232 to unblock it

S-Coyle:
Removed the `Don't merge` label now that the code coverage can be run with latest vault, cli and authd, plus the `latest` files required have been manually uploaded to S3.

Updated the description to state that as there are currently tests failing when attempting to gather code coverage, it may be best to fix these before this PR is merged. For discussion.

S-Coyle:
Closing as this PR has gone stale - can recreate in future once the repo is stable

PR #1229, opened 2020-07-30
Add CI tests against real network

S-Coyle:
This PR adds a scheduled test which will run each day:
- Start a section with the latest released node, downloaded from S3, and run the client tests (cargo test --release) on it.

It also adds e2e testing against a section with the `latest` released node to PR CI.


.github/workflows/vault_network_test-latest_release.yml
```diff
@@ -0,0 +1,76 @@
...
+  # Deny all compiler warnings.
+  RUSTFLAGS: "-D warnings"
+  # set the vault, cli and authd version to be downloaded and used
+  SAFE_VAULT_VERSION: "latest"
+  SAFE_CLI_VERSION: "latest"
+  SAFE_AUTH_VERSION: "latest"
+
+jobs:
+  network-test:
+    name: E2E against real baby
```
joshuef, line 23: lets just call this "against real network", as baby is more transient

S-Coyle: Updated in the 3 files

joshuef:
Looks like we're missing increasing the test startup interval: https://github.com/maidsafe/sn_node/blob/master/.github/workflows/pr.yml#L15

.github/workflows/node_network_test-latest_release.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-cache-${{ hashFiles('**/Cargo.lock') }}
+
+      # Run tests.
+      - name: Run the tests
+        run: cargo test --release
```
joshuef, line 66:
```suggestion
        run: cargo test --release --features=simulated-payouts
```

PR #1232, opened 2020-08-17
refactor(ci): Refactor of GHA CI files

S-Coyle:
Updated version of the checkout action being used to v2.
Tidied and condensed the caching.
Removed some unnecessary use of the cargo action, switching to
use the cargo command instead.

Split this out from #1224 as there has been significant updates to the CI files since then, plus the other commit in that PR is blocked and so was needlessly stopping this commit from being merged.


S-Coyle:
superseded by #1239, which is based off the latest master. Closing

PR #1239, opened 2020-08-25
refactor(ci): Refactor of GHA CI files

S-Coyle:
Updated version of the checkout action being used to v2.
Tidied and condensed the caching.
Removed some unnecessary use of the cargo action, switching to
use the cargo command instead.

Supersedes #1232

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - fleming
 
 env:
   # Run all cargo commands with --verbose.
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -122,46 +108,43 @@ jobs:
...
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-cache-${{ hashFiles('**/Cargo.lock') }}
+
+      # Test
+      - name: Immutable Data Tests
+        run: cargo test --manifest-path=safe_core/Cargo.toml --features=testing create_and_retrieve # id tests only
 
   # Test publish using --dry-run.
   # Only safe_core, as auth and app will fail if core was version-updated.
```
Yoga07, line 135: We don't have auth and app any more here. So should be good to have this line removed :)

.github/workflows/master.yml
```diff
@@ -164,31 +164,30 @@ jobs:
...
-      - uses: actions-rs/cargo@v1
-        with:
-          command: publish
-          args: --manifest-path=safe_authenticator/Cargo.toml
+
+      - name: Publish safe_authenticator
         if: steps.versioning.outputs.auth_version != steps.versioning.outputs.old_auth_version
-     
\ No newline at end of file
+        run: cargo publish --manifest-path=safe_authenticator/Cargo.toml
```
Yoga07, line 193: Same, no more safe_authenticator in SCL. Better to have it removed too?

Yoga07 (commented):
Just a couple of minor ones :)

.github/workflows/pr.yml
S-Coyle: Removed the `--manifest-path=safe_core/Cargo.toml` part of the test, thanks

.github/workflows/master.yml
S-Coyle: removed, thanks

Yoga07 (approved):

PR #1253, opened 2020-09-04
fix/conn: fix listening mechanism of Client

Yoga07:

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -38,8 +38,8 @@ use lru::LruCache;
 use quic_p2p::Config as QuicP2pConfig;
 use rand::thread_rng;
 use safe_nd::{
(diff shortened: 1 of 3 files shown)


src/client/mod.rs
```diff
@@ -38,8 +38,8 @@ use lru::LruCache;
 use quic_p2p::Config as QuicP2pConfig;
 use rand::thread_rng;
 use safe_nd::{
-    Blob, BlobAddress, ClientFullId, Cmd, Message, MessageId, Money, PublicId, PublicKey, Query,
-    QueryResponse, Sequence, SequenceAddress,
+    Blob, BlobAddress, ClientFullId, Cmd, Message, MessageId, Money, MsgEnvelope, PublicId,
```
joshuef, line 41: @Yoga07 it would be good to keep `MsgEnvelope` encapsulated in `connection_manager` IMO

PR #1255, opened 2020-09-07
chore(deps): update to reference renamed sn_data_types crate

S-Coyle:
Note that I had to update the `crdts` dep to "~4.2.0" to satisfy cargo. I assume this update is OK.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,9 +404,9 @@ dependencies = [
 
 [[package]]
 name = "crdts"
-version = "4.1.0"
+version = "4.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


S-Coyle:
Closing this PR. Seems that SCL is relying on an older version of sn_data_types - 0.11.2 which is from when it was named safe-nd. Attempting to update to use the new name of sn_data_types is resulting in a load of clippy errors because all versions of sn_data_types which have been released are not compatible. Errors such as:
```
error[E0432]: unresolved imports `sn_data_types::SequenceOwner`, `sn_data_types::SequencePrivUserPermissions`, `sn_data_types::SequencePubUserPermissions`, `sn_data_types::SequenceUserPermissions`
```

PR #1261, opened 2020-09-16
Client Event Listener setup

joshuef:

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -26,18 +26,15 @@ use tokio::task::JoinHandle;
 /// Simple map for correlating a response with votes from various elder responses.
 type VoteMap = HashMap<QueryResponse, usize>;
 
-#[derive(Clone)]
-pub struct ElderConnection {
-    pub connection: Arc<Mutex<Connection>>,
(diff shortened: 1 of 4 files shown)


src/connection_manager.rs
```diff
@@ -301,14 +361,21 @@ impl ConnectionManager {
...
-                self.elders.push(elder);
+                self.elders.push(ElderStream {
+                    send_stream,
+                    connection,
+                    listener: Arc::new(Mutex::new(listener)),
+                });
             }
 
+            // TODO: this will effectively stop driving futures after we get 2...
+            // We should still let all progress... just without blocking
```
Yoga07, line 378: Think we can increase/remove this as we are now spawning 7 elders?

Yoga07 (approved):

src/connection_manager.rs
joshuef: Not related to that. I'm not sure the loope here will guarantee 7 elders... it'll stop after > 2. Which could be 3 or 7....

But that's not related to this PR, so something to look at as we get 7 elders reliably (i still don't)

Yoga07 (approved):

PR #1271, opened 2020-10-08
Aggregate SignatureShares for AT2 transfers and multiple fixes

Yoga07:

### DIFF
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -302,8 +302,7 @@ mod tests {
     // 1. Create a client A and allocate 100 Money to it. (Clients start with 10 Money by default on simulated-farming)
     // 2. Get the balance and verify it.
     // 3. Create another client B with a wallet holding 10 Money on start.
-    // 4. Transfer 5 Money from client A to client B and verify the new balance.
-    // 5. Try to do a coin transfer without enough funds, it should return `InsufficientBalance`
+    // 4. Transfer 5 Money from client A to client B and verify the new balances.
     #[tokio::test]
     #[cfg(feature = "simulated-payouts")]
     pub async fn balance_transfers_between_clients() -> Result<(), ClientError> {
@@ -319,7 +318,6 @@ mod tests {
         let balance = client.get_balance().await?;
         assert_eq!(balance, Money::from_str("110")?); // 10 coins added automatically w/ farming sim on client init.
 
-        let orig_balance = client.get_balance().await?;
         let _ = client.send_money(wallet1, Money::from_str("5.0")?).await?;
 
         // Assert sender is debited.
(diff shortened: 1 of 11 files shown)


src/client/transfer_actor/balance_management.rs
```diff
@@ -232,63 +232,66 @@ mod tests {
...
+        let (_, random_pk) = shared_box::gen_bls_keypair();
 
-        let mut client = Client::new(Some(secret_key)).await?;
+        let mut client = Client::new(None).await?;
 
-        let res = client
-            .send_money(PublicKey::Bls(sk2.public_key()), Money::from_str("0")?)
+        // Send 0 Money to a random PK.
+        client
+            .send_money(PublicKey::Bls(random_pk), Money::from_str("0")?)
```
joshuef, line 279: @Yoga07 this should fail no?

src/client/transfer_actor/balance_management.rs
```diff
@@ -303,30 +306,36 @@ mod tests {
...
-                .checked_sub(Money::from_str("5.0")?)
-                .ok_or_else(|| ClientError::from("Invalid checked sub in test"))?,
+            calculate_new_balance(balance, None, Some(Money::from_str("5.0")?))
         );
 
+        // Assert that the receiver has been credited.
+        let mut receiving_bal = receiving_client.get_balance().await?;
+        assert_eq!(receiving_bal, Money::from_str("15.0")?);
+
+        // Try transferring money exceeding our balance.
```
joshuef, line 324: @Yoga07 this should be another test IMO. `cannot_transfer_in_excess_of_funds`

src/connection_manager.rs
```diff
@@ -39,6 +41,23 @@ struct ElderStream {
     socket_addr: SocketAddr,
 }
 
+struct DebitProofAggregator {
```
joshuef, line 44: I'm not sure this should be handled a) in connection_manager, or b) in the client... Isn't this already handled by transfers? Or if not.. .should it be handled there?

src/connection_manager.rs
```diff
@@ -513,7 +536,7 @@ impl ConnectionManager {
 
             // TODO: this will effectively stop driving futures after we get 2...
             // We should still let all progress... just without blocking
-            if self.elders.len() > 5 {
+            if self.elders.len() >= 7 {
```
joshuef, line 539: We may not be able to connect to all 7... do we need all 7? 

This is an ongoing thing we need to look at... what is acceptable for us here?

src/connection_manager.rs
```diff
@@ -553,12 +576,51 @@ impl ConnectionManager {
...
                             Message::Event {
-                                event: _,
-                                correlation_id: _,
+                                event,
+                                correlation_id,
                                 ..
                             } => {
                                 // TODO: Send full event or error out to desired listener.
                                 warn!("Event received {:?}", envelope.message);
+                                if let Event::TransferValidated { event, .. } = event {
```
joshuef, line 585: Again, I'm not sure this is necessary vs just passing in to the actor. That should spit out a debitProof when there's enough sigs

joshuef (changes requested):

src/connection_manager.rs
Yoga07: Section spin-up has become a lot more reliable with the latest routing changes and we are now consistently having 7 Elders starting good. So it would be ideal for the client to start connecting to 7 nodes I believe :)

src/client/transfer_actor/balance_management.rs
Yoga07: Will push in the upcoming PR :+1:

lionel1704 (dismissed):

src/client/transfer_actor/balance_management.rs
Yoga07: We've been allowing zero-value transfers so far, but with that fixed in maidsafe/sn_transfers#41, changes have been made to reflect the same :+1:

src/connection_manager.rs
```diff
@@ -509,7 +510,7 @@ impl ConnectionManager {
 
             // TODO: this will effectively stop driving futures after we get 2...
             // We should still let all progress... just without blocking
-            if self.elders.len() > 5 {
+            if self.elders.len() >= 7 {
```
lionel1704, line 513: Will this ever be greater than 7?

src/client/transfer_actor/balance_management.rs
```diff
@@ -307,30 +318,36 @@ mod tests {
...
         let res = client.send_money(wallet1, Money::from_str("5000")?).await;
```
joshuef, line 337: This should still be another test @Yoga07

src/client/transfer_actor/mod.rs
```diff
@@ -174,24 +173,60 @@ impl Client {
...
+            .connection_manager
+            .lock()
+            .await
+            .send_query(&message)
+            .await?;
+
+        match res {
+            QueryResponse::GetStoreCost(cost) => cost.map_err(ClientError::DataError),
+            _ => Err(ClientError::from(format!(
+                "Unexpected response when retrieving account history {:?}",
```
joshuef, line 200:
```suggestion
                "Unexpected response when retrieving store core {:?}",
```

lionel1704 (approved):
Just a trivial comment!

src/connection_manager.rs
joshuef: Should be ideal... but what if we're not in ideal conditions? for a client what is _enough_? 3 elders? 5? 7?

If we cannot get 7, i'd say 5 should be sufficient for most ops. The elders will comm between themselves no?

Or are we saying if we cannot connect to 7, there's an issue worth failing over?

src/utils/mod.rs
```diff
@@ -184,6 +185,11 @@ pub fn bin_data_format(data: &[u8]) -> String {
     )
 }
 
+/// Wrap a data Cmd.
+pub fn wrap_data_cmd(cmd: DataCmd, payment: DebitAgreementProof) -> Cmd {
+    Cmd::Data { cmd, payment }
```
joshuef, line 190: IMO this util wrapper func doesn't add thaaat  much. We loose more in terms of `use::util::wrap_data_cmd` than just having this implemented at each point...

We loose clarity, and it's the same (or more) LOC, i think.

joshuef (changes requested):

src/connection_manager.rs
Yoga07: In the future maybe? When we move away from static `ELDER_SIZE` in routing. But yea, it's a long way ahead.

joshuef: If we ever have flexi elder numbers, could be. 

I guess the arg for 7 is that anything less and we cannot trust elders... You could have 2 under DOS by 3 elders, and then they have majority over what happens... Still, impact would only be about GETs, (with any pay ops needing sig) but could be a pain.

So aye, actually 7 is good I guess :+1:

joshuef (approved):

PR #1288, opened 2020-11-24
feat(err_listener): implement CmdError listener and fix map data tests

Yoga07:

### DIFF
--- a/src/client/map_apis.rs
+++ b/src/client/map_apis.rs
@@ -654,6 +654,8 @@ pub mod exported_tests {
     use crate::utils::test_utils::gen_bls_keypair;
     use sn_data_types::{Error as SndError, MapAction, MapKind, Money};
     use std::str::FromStr;
+    use std::thread::sleep;
+    use std::time::Duration;
     use xor_name::XorName;
 
     // 1. Create unseq. map with some entries and perms and put it on the network
@@ -675,7 +677,7 @@ pub mod exported_tests {
         let address = client
             .store_unseq_map(name, tag, owner, Some(entries.clone()), Some(permissions))
             .await?;
-        println!("Put unseq. Map successfully");
+        sleep(Duration::from_secs(1));
 
         let version = client
             .get_map_version(MapAddress::Unseq { name, tag })
@@ -721,6 +723,7 @@ pub mod exported_tests {
         let address = client
             .store_seq_map(name, tag, owner, Some(entries.clone()), Some(permissions))
             .await?;
+        sleep(Duration::from_secs(1));
         println!("Put seq. Map successfully");
 
         let fetched_entries = client.list_seq_map_entries(name, tag).await?;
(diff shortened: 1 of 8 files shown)


src/client/map_apis.rs
```diff
@@ -775,7 +778,9 @@ pub mod exported_tests {
         let owner = client.public_key().await;
 
         let address = client.store_unseq_map(name, tag, owner, None, None).await?;
+
         client.delete_map(mapref).await?;
+        sleep(Duration::from_secs(2));
```
joshuef, line 783: We want to be removing sleeps.

Can we loop over the get until it fails? We can put a limit on the global test timeout then instead to stop this running forever

src/client/mod.rs
```diff
@@ -158,6 +164,7 @@ impl Client {
             simulated_farming_payout_dot,
             blob_cache: Arc::new(Mutex::new(LruCache::new(IMMUT_DATA_CACHE_SIZE))),
             sequence_cache: Arc::new(Mutex::new(LruCache::new(SEQUENCE_CRDT_REPLICA_SIZE))),
+            notif_receiver: Arc::new(Mutex::new(notif_receiver)),
```
joshuef, line 167: can we name this `notification_receiver` and sender? so much easier to read IMO

src/client/mod.rs
```diff
@@ -275,18 +282,37 @@ impl Client {
 
         self.apply_write_payment_to_local_actor(payment_proof).await
     }
+
+    /// Assert if all 5 Elders are returning the same errors we expect.
+    #[cfg(feature = "simulated-payouts")]
+    pub(crate) async fn expect_error(&mut self, err: ClientError) {
+        for _ in 0..ELDER_SIZE {
```
joshuef, line 289: Elder size may not be 5... could be 3 also now. Which is fine, but maybe just for the comment above... (Also if this is run in node tests, w/ `chaos` we may be dropping more messages which is something to consider here. though not sure what the answer is yet)

src/connection_manager.rs
```diff
@@ -50,11 +51,16 @@ pub struct ConnectionManager {
     elders: Vec<ElderStream>,
     endpoint: Arc<Mutex<Endpoint>>,
     pending_transfer_validations: Arc<Mutex<HashMap<MessageId, TransferValidationSender>>>,
+    notif_sender: UnboundedSender<ClientError>,
```
joshuef, line 54: lets rename this too please

joshuef (changes requested):

src/client/sequence_apis.rs
```diff
@@ -1250,15 +1251,18 @@ pub mod exported_tests {
         let address = client
             .store_private_sequence(None, name, tag, owner, perms)
             .await?;
-
+        sleep(std::time::Duration::from_secs(2));
```
joshuef, line 1254: Ah, we'll need to get these sleeps out too

joshuef (dismissed):

joshuef (approved):

joshuef (approved):

PR #1295, opened 2020-12-01
Fix blob implementation and refactor the API and tests.

lionel1704:


src/client/blob_apis.rs
```diff
@@ -44,58 +52,63 @@ impl Client {
...
+    /// // Let's use an existing client, with a pre-existing balance to be used for write payments.
+    /// let mut client = Client::new(None).await?;
+    /// # let initial_balance = Money::from_str("100")?; client.trigger_simulated_farming_payout(initial_balance).await?;
+    /// let data = b"some data".to_vec();
+    /// // grab the random blob from the network
+    /// let _address = client.create_public_blob(&data).await?;
+    ///
+    /// # let balance_after_write = client.get_local_balance().await; assert_ne!(initial_balance, balance_after_write); Ok(()) } ); }
+    /// ```
+    pub async fn create_public_blob(&mut self, data: &Vec<u8>) -> Result<BlobAddress, ClientError> {
```
joshuef, line 104: hmmmm, why `create` over `store` here? would be good to stay consistent across the APIs

src/client/blob_apis.rs
```diff
@@ -233,63 +299,44 @@ impl Client {
...
-    async fn pack(
-        &mut self,
-        public_key: PublicKey,
-        mut value: Vec<u8>,
-        published: bool,
-    ) -> Result<Blob, ClientError> {
-        // This blob storage is used to perform write operations to the network, via self encryptor
-        let blob_storage = BlobStorage::new(self.clone(), published);
-
+    /// This function takes the "Root data map" and returns a Blob that is acceptable by the network
```
joshuef, line 302: I think we can drop "This function" from the docs. It's redundant

joshuef (dismissed):
Couple small things, otherwise lgtm!

src/client/blob_apis.rs
lionel1704: Used create here because it's not exactly, "Here's a blob, store it". But rather, "Here's some data, put it in a blob for me".
I can change it if required, not a biggie.

joshuef: Yeh, I see it fits well for create... 
At this layer, during the refactor I was trying to avoid such abstractions (putting data vs the data type). That should really be the at the API layer IMO, and just `store(<Blob>)` here

bochaco: Isn't the data passed in already a blob? from API's perspective it's already a blob, isn't it?

bochaco (approved):

PR #1296, opened 2020-12-01
feat: add configurable bootstrap to client

joshuef:

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -105,12 +105,15 @@ impl Client {
     ///
     /// # #[tokio::main] async fn main() { let _: Result<(), ClientError> = futures::executor::block_on( async {
     ///
-    /// let mut client = Client::new(None).await?;
+    /// let mut client = Client::new(None, None).await?;
     /// // Now for example you can perform read operations:
     /// let _some_balance = client.get_balance().await?;
     /// # Ok(()) } ); }
     /// ```
-    pub async fn new(optional_keypair: Option<Arc<Keypair>>) -> Result<Self, ClientError> {
+    pub async fn new(
+        optional_keypair: Option<Arc<Keypair>>,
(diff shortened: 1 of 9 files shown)


src/client/mod.rs
```diff
@@ -309,8 +317,14 @@ pub async fn attempt_bootstrap(
     qp2p_config: &QuicP2pConfig,
     keypair: Arc<Keypair>,
     notification_sender: UnboundedSender<ClientError>,
+    bootstrap_nodes: Option<HashSet<SocketAddr>>,
 ) -> Result<ConnectionManager, ClientError> {
     let mut attempts: u32 = 0;
+    let mut qp2p_config = qp2p_config.clone();
+
+    if let Some(contacts) = bootstrap_nodes {
+        qp2p_config.hard_coded_contacts = contacts;
```
lionel1704, line 326: Is this meant to replace all the contacts in the qp2p_config ? Perhaps this should be extend ?

lionel1704: If we change this to extend() then we can remove the Option as well. So just pass Default::default() instead of None.

joshuef: Hmm, I'm not sure. I'd guess if you're passing in something you may want to override? (what if trying to override a broken network as default eg)

joshuef: >we can remove the Option as well. So just pass Default::default() instead of None.
Where do you mean sorry?

bochaco: I think we should keep it simple and just allow for overriding, and therefore we need the `Option<...>`. Although, ar we ok with having a `HashSet` as in the API's args? I always tend to use the most primitive data types for API args...

joshuef: I've just rolled with what was existing for BootstrapConfig here. I did read up on the diff w/ vec, and it was preventing duplicates. which could be useful here

bochaco (approved):

PR #1297, opened 2020-12-03
tests: improve tests, and update qp2p

joshuef:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ log = "~0.4.8"
 log4rs = { version = "~0.8.3", features = ["toml_format"] }
 lru = "~0.5.1"
 miscreant = { version = "~0.4.2", features = ["soft-aes"] }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ log = "~0.4.8"
 log4rs = { version = "~0.8.3", features = ["toml_format"] }
 lru = "~0.5.1"
 miscreant = { version = "~0.4.2", features = ["soft-aes"] }
-qp2p = "~0.9.1"
+qp2p = "~0.9.2"
```
S-Coyle, line 30: Should this be ~0.9.3?

joshuef: wouldnt actually matter i think, as cargo would choose latest patch v. But i have updated anywho

S-Coyle (approved):

S-Coyle (approved):
👍

PR #1299, opened 2020-12-07
feat: update lseq data type

joshuef:

### DIFF
--- a/src/client/sequence_apis.rs
+++ b/src/client/sequence_apis.rs
@@ -968,13 +968,21 @@ pub mod exported_tests {
             .store_private_sequence(None, name, tag, owner, perms)
             .await?;
 
-        let balance_before_delete = client.get_balance().await?;
-        client.delete_sequence(sequence_address).await?;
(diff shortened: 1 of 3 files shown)


src/client/sequence_apis.rs
```diff
@@ -968,13 +968,21 @@ pub mod exported_tests {
             .store_private_sequence(None, name, tag, owner, perms)
             .await?;
 
-        let balance_before_delete = client.get_balance().await?;
+        let mut balance_before_delete = client.get_balance().await?;
+
+        while balance_before_delete == Money::from_str("0")?
```
Yoga07, line 973: Could this actually be a case? We can't write to the network with 0 bal.

joshuef: We could get 0 back from _some_ elder. That doesn't mean the write hasn't gone through.

And equally, we could do the delete, and still be reported 0 from some elder / combo.

Yoga07 (approved):

PR #1319, opened 2020-12-29
feat(errors): use thiserror for error creation

joshuef:


src/client/sequence_apis.rs
```diff
@@ -670,10 +665,10 @@ impl Client {
         // TODO: do we need to query with some specific PK?
         let perms = match sequence
             .permissions(SequenceUser::Key(user), None)
-            .map_err(ClientError::from)?
+            .map_err(Error::from)?
```
bochaco, line 668: Just curious if we can remove this mapping and let it do the conversion automatically?

Cargo.toml
```diff
@@ -27,8 +27,13 @@ lru = "~0.5.1"
 qp2p = "~0.9.3"
 rand = "~0.7.3"
 regex = "1.3.1"
-sn_data_types = "~0.11.40"
-sn_transfers = "~0.2.5"
+anyhow="1.0.36"
```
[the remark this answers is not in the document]
joshuef: Atm it can't due to exporting tests for node. As/when we use compiled CLI tests for node, this can all be moved back into test module and dev deps 👍

src/client/sequence_apis.rs
joshuef: no, removing it altogether throws the error type we're expecting for other arms. Just `map_err(Error)` doesn't work, and `into` should, but we'd have to match out the error first anywho, so I think this is simplest

bochaco: ah right, i was assuming we had the errors in the Error type with `#[from]`

bochaco (approved):

bochaco (approved):

PR #1327, opened 2021-01-07
Fix onboarding

joshuef:
- updates to remove 'stream' listening as nodes no longer store streams for responses.
- all messages are now caught on `IncomingMessages` (from qp2p), and parsed / forwarded from there.


src/connection_manager.rs
```diff
@@ -541,72 +568,91 @@ impl ConnectionManager {
...
         let pending_transfer_validations = Arc::clone(&self.pending_transfer_validations);
         let notifier = self.notification_sender.clone();
 
+        let pending_queries = self.pending_query_responses.clone();
+
         // Spawn a thread for all the connections
         let handle = tokio::spawn(async move {
-            info!("Listening for incoming connections started");
-
             // this is recv stream used to send challenge response. Send
```
Yoga07, line 586: Can remove the outdated comment

Yoga07 (commented):
Just a few general comments. Looks good otherwise :+1:

src/connection_manager.rs
```diff
@@ -382,31 +402,49 @@ impl ConnectionManager {
...
-
-                // Obtain the addresses of the Elders
-                let elders_addrs = elders.into_iter().map(|(_xor_name, ci)| ci).collect();
-                Ok(elders_addrs)
+        match conn.send_bi(msg).await {
+            Ok(_) => {
+                if let Some(message) = incoming_messages.next().await {
+                    match message {
+                        Qp2pMessage::BiStream { bytes, .. }
+                        | Qp2pMessage::UniStream { bytes, .. } => {
```
Yoga07, line 417: Can nodes be connected to the client via `UniStream`? Think we should return an error here.

bochaco: I'm not sure, but once they have the connection created I think both sides can create any type of streams?

src/connection_manager.rs
```diff
@@ -267,9 +276,20 @@ impl ConnectionManager {
                         if *counter > threshold {
                             trace!("Enough votes to be above response threshold");
 
-                            winner = (Some(response), *counter);
+                            winner = (Some(response.clone()), *counter);
                             has_elected_a_response = true;
                         }
+
+                        // short circuit GetHistory w/ simulated payouts, as we will _never_ have two responses which are the same
```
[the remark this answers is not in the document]
joshuef: Initial history during `SimulatedPayouts` is generated at each elder. Thus, each history is different so we never get to "one true history". 

I may not strictly be necessary... I was seeing delays and suspectes this was it, but I think it's the qp2p issue in the end. Still... could make sense for testnets where we'll never arrive at a "correct" history

src/connection_manager.rs
joshuef: We're getting uni messages back at the mo when we do `routing.send_to_client` at nodes. What qp2p is doing under the hood I'm not sure. Sending a uni back on the original bistream i guess?

src/connection_manager.rs
```diff
@@ -354,10 +373,10 @@ impl ConnectionManager {
 
         if number_of_responses > threshold {
             trace!("No clear response above the threshold, so choosing most popular response with: {:?} votes: {:?}", most_popular_response.1, most_popular_response.0);
-
-            *has_elected_a_response = true;
         }
 
+        *has_elected_a_response = true;
```
bochaco, line 378: Since this is a boolean, I don't see a point in passing it as a `&mut` but instead just return the new value in a pair/tuple. At the same time it seems this is always set to true by this function? if so there is no point in even returning it

joshuef: ahhh, yeh. super true. Unneeded now it's moved out of there 👍

src/connection_manager.rs
```diff
@@ -167,38 +201,64 @@ impl ConnectionManager {
...
-                                    Err(Error::ReceivingQuery)
-                                }
+                    // TODO: we need to remove the msg_id from
+                    // pending_query_responses upon any failure below
+                    match connection.send_bi(msg_bytes_clone).await {
+                        Ok((send_stream, _)) => {
+                            send_stream.finish().await?;
+
+                            // TODO: receive response here.
+                            result = match receiver.recv().await {
```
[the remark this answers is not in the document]
joshuef: We only get/want one on this channel. The threshold handling is below here. First we get our response tasks set up (each receives 1 response). And then from those, we vote/get to threshold (the looop from ln 300)

src/connection_manager.rs
```diff
@@ -234,8 +285,9 @@ impl ConnectionManager {
         let mut vote_map = VoteMap::default();
         let mut received_errors = 0;
 
-        // TODO: make threshold dynamic based upon known elders
-        let threshold: usize = (self.elders.len() as f32 / 2_f32).ceil() as usize;
+        // 2/3 of known elders
+        // let threshold: usize = (self.elders.len() as f32 / 2_f32).ceil() as usize;
+        let threshold: usize = 1;
```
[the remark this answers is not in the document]
joshuef: Ach good spot 👍 That's a debugging hangover

src/client/transfer_actor/write_apis.rs
```diff
@@ -40,10 +41,11 @@ pub mod exported_tests {
 
         let initial_actor = Client::new(Some(keypair), None).await?;
         match initial_actor.pay_and_write_sequence_to_network(data).await {
-            Err(Error::NetworkDataError(e)) => {
-                assert!(e
-                    .to_string()
-                    .contains("Could not get history for key PublicKey"));
+            Err(Error::Transfer(TransfersError::InsufficientBalance)) => {
+                // Ok(())
+                // assert!(e
```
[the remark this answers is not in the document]
joshuef: Actually just `Ok` should be enough. The error message comes from the specific error itself, so the above match will do

Yoga07 (approved):

PR #1328, opened 2021-01-08
Add get_sequence_at_version method

iancoleman:
This removes the need to call get_sequence_range(address, (index, index+1))

This change is a prerequisite to removing the use of `get_sequence_range` when fetch a single sequence entry (which is in [this part of the source code](https://github.com/maidsafe/sn_api/blob/67b746f607501511c38fe752f64119a12985ab72/sn_api/src/api/app/safe_client.rs#L491-L495)).

The response is consistent with `get_sequence_last_entry`

Not sure if fetching from too high a version should give the latest, or should throw an error. Considering `in_range` allows versions higher than the number available (not explicitly but as a side effect), I went with an equivalent concept here. Returning the actual version allows the user to know if they've requested a version too high and deal with it in their own way. I imagine it's most useful to the client to return the latest data and the latest version, rather than an error and the latest version.

---

**Example code in docs**

I did not verify the example code works, I copied and modified from `get_sequence_last_entry`. Is there a standard way to check the code example?

**Tests**

I tried to run the test using

```
cargo test --release --target x86_64-unknown-linux-musl append_to_sequence_test
```

but get an error

```
Failed to bootstrap to the network: Connection lost
```

Any tips how to get the tests to work?

**Cargo.lock**

Also not sure if I should check in the changes to `Cargo.lock` or not. Any tips?

### DIFF
--- a/src/client/sequence_apis.rs
+++ b/src/client/sequence_apis.rs
@@ -422,6 +422,63 @@ impl Client {
         }
     }
 
+    /// Get Sequence Data from the Network at a specific version
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # extern crate tokio; use sn_client::Error; use std::str::FromStr;
+    /// use sn_client::Client;
+    /// use sn_data_types::{Keypair, PublicKey, Money, SequencePrivatePermissions};
+    /// use std::collections::BTreeMap;
+    /// use xor_name::XorName;
+    /// use rand::rngs::OsRng;
+    /// # #[tokio::main] async fn main() { let _: Result<(), Error> = futures::executor::block_on( async {
+    /// // Let's use an existing client, with a pre-existing balance to be used for write payments.
+    /// let id = std::sync::Arc::new(Keypair::new_ed25519(&mut OsRng));
+
+    /// let mut client = Client::new(Some(id), None).await?;
+    /// # let initial_balance = Money::from_str("100")?; client.trigger_simulated_farming_payout(initial_balance).await?;
+    /// let name = XorName::random();
+    /// let tag = 10;
+    /// let owner = client.public_key().await;
+    /// let mut perms = BTreeMap::<PublicKey, SequencePrivatePermissions>::new();
+    ///
+    /// // Set the access permissions
+    /// let _ = perms.insert(
+    ///    owner,
+    ///    SequencePrivatePermissions::new(true, true, true),
+    /// );
+    ///
+    /// // The returned address can then be used to `append` data to.
+    /// let address = client.store_private_sequence(None, name, tag, owner, perms).await?;
+    /// client.append_to_sequence(address, b"New Entry Value".to_vec()).await?;
+    /// client.append_to_sequence(address, b"Another New Entry Value".to_vec()).await?;
+    ///
+    /// // Now we can retrieve the alst entry in the sequence:
+    /// let entry_v1 = client.get_sequence_entry(address, 1).await?;
+    ///
+    /// assert_eq!(entry_v1, b"Another New Entry Value".to_vec());
+    /// # Ok(()) } ); }
+    /// ```
+    pub async fn get_sequence_entry(
+        &self,
+        address: SequenceAddress,
+        index_from_start: u64,
+    ) -> Result<SequenceEntry, Error> {
+        trace!(
+            "Get entry at index {:?} from Sequence Data {:?}",
+            index_from_start,
+            address.name()
+        );
+
+        let sequence = self.get_sequence(address).await?;
+        let index = SequenceIndex::FromStart(index_from_start);
+        match sequence.get(index, None)? {
+            Some(entry) => Ok(entry.to_vec()),
+            None => Err(Error::from(sn_data_types::Error::NoSuchEntry)),
+        }
+    }
+
     /// Get a set of Entries for the requested range from a Sequence.
     ///
     /// # Examples
@@ -1252,7 +1309,7 @@ pub mod exported_tests {
         }
     }
 
-    pub async fn append_to_sequence_test() -> Result<(), Error> {
+    pub async fn append_to_sequence_test() -> anyhow::Result<()> {
         let name = XorName(rand::random());
         let tag = 10;
         let client = Client::new(None, None).await?;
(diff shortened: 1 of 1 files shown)


dirvine:
Nice one Ian, I will let @bochaco  chime in here (he is on ODD day today). The tests are currently a bit fragile with some changes we are making, also the qp2p usage. That is currently being worked on. So this should all be more concrete next week and this PR then will likely pass the tests (at least no connection issues)

src/client/sequence_apis.rs
```diff
@@ -422,6 +422,70 @@ impl Client {
...
+    /// // Now we can retrieve the alst entry in the sequence:
+    /// let (_version, entry_v1) = client.get_sequence_at_version(address, 1).await?;
+    ///
+    /// assert_eq!(entry_v1, b"Another New Entry Value".to_vec());
+    /// ```
+    pub async fn get_sequence_at_version(
+        &self,
+        address: SequenceAddress,
+        version: u64,
+    ) -> Result<(u64, SequenceEntry), Error> {
```
bochaco, line 466: > Not sure if fetching from too high a version should give the latest, or should throw an error. Considering in_range allows versions higher than the number available (not explicitly but as a side effect),

I'm imagining that perhaps a user using this function wants a specific version, rather than "give me this version or latest", thus perhaps it's better to simply return an error if the index is invalid? otherwise, assuming the normal use case is trying to fetch a specific version the user will need to call this function and also check is the one requested and not a higher one.

bochaco (commented):
To run the tests you need a section running, thus launch a local section with master sn-node codebase. We've removed all type of mocks from all crates/layers.

bochaco:
> Also not sure if I should check in the changes to `Cargo.lock` or not. Any tips?

I think we shouldn't, so it should be added to the `.gitignore` file. Feel free to do so in this PR if you will or just ignore it.

src/client/sequence_apis.rs
iancoleman: Yes I think the error is ok, but I feel we should try to reduce the need to make another request where possible. I'd rather the client do the (almost zero) work of comparing the requested version with the response version, than have them put additional load on the network by repeating the request with the corrected (presumably latest) version.

What would a user be least surprised by? An error? Or the latest version with a correction on the number? Probably an error since the data doesn't exist, like http 404 response.

What do you think is a suitable error to return? Errors (to my understanding) are just strings, right?

Using a too-high version with `in_range` does not return an error ([source](https://github.com/maidsafe/sn_data_types/blob/b0796d1a4baf9f53d68e07425eb20791651da382/src/sequence/seq_crdt.rs#L382-L395)). Maybe it should? Just trying to be as consistent as possible.

I feel it's better for network health to return the latest version, but can see how an error would be most expected. Honestly I'm not sure which way to go on this one.

bochaco: The reason why I think we should return an error because the function is meant to return a specific version, if the user was looking for latest then it it already has `get_sequence_last_entry` function. So not sure it's of any use to return latest if the one you were requesting is not there, here I'm assuming/imagining that most users will simply fail with an error since they were looking for a specific version and it doesn't exist, rather than looking for one or latest. I'm with you about consistency anyway, that's the most important thing.

> Using a too-high version with in_range does not return an error (source). Maybe it should?

I think e return an error from the client since [this](https://github.com/maidsafe/sn_data_types/blob/b0796d1a4baf9f53d68e07425eb20791651da382/src/sequence/seq_crdt.rs#L385) should cause the call to return `None`, and we match that to an error: https://github.com/maidsafe/sn_client/blob/master/src/client/sequence_apis.rs#L480. Although, we can argue if the function in data-types should return an error in that case rather than `None`, it sounds like it should.

> What do you think is a suitable error to return? Errors (to my understanding) are just strings, right?

No, they are not strings, they contain a description but they can be identified by the user to act upon it accordingly, these are existing ones in sn_client: https://github.com/maidsafe/sn_client/blob/master/src/errors.rs#L21

bochaco: I think getting @joshuef input here will be helpful for deciding

joshuef: I think naming might be off... `get_sequence_at_entry` is perhaps truer to the seq impl. `version` _could_ be what a sequence is used for... but that's not inherent here.

-----

>  we shouldn't be generating an sn_data_types::Error from the sn_client crate as we do there.
That's making a client error, chainging from DT error. There error is in the DT no? So I _think_ it makes sense to be sending that back up?

----

@iancoleman you can verify the doc tests work with `cargo test --doc`, what will at least verify compilation of the example code.

----

Re Error or not to error... we have `get_last_entry` as gab says, if you only want latest. If you're going for a specific entry an error seems most intuitive if it's not existing. 

if it can return last entry you have more scope for mixed responses from out of sync nodes. If each storing node doesn't have the same last entry, you could get an earlier 'last' entry... 

(This is an issue with `get_last_entry` response management too. but I imagine that has to be explicitly dealt with for those responses.... ie. err on higherIndex even when it's a less common response. For `get_entry_at_index` that seems superfluous to me).

So I _think_ an error is what you'd expect.

joshuef: I think https://github.com/maidsafe/sn_client/pull/1328/files#diff-054f27c0e3e484d166bd1b31aad5eee4f9b5de9d6114f3957dffcf83c01bc7c1R485 is the correct error to be returning.

iancoleman:
Changed:

* Rename `get_sequence_at_version` to `get_sequence_entry`
* Rename input argument from `version` to `index_from_start`
* Return the entry if it exists, otherwise return an error.
    * Return only the entry, not the tuple `(version, entry)`
    * Return an error if the requested index is too large, instead of the latest entry

iancoleman:
> To run the tests you need a section running, thus launch a local section with master sn-node codebase. We've removed all type of mocks from all crates/layers.

I was having trouble with this but found I was missing the `--features=simulated-payouts` flag. This is the command to run tests that worked for me (with a network already running):

```
cargo test --release --features=simulated-payouts --target x86_64-unknown-linux-musl
```

bochaco (approved):
Nice! it looks good to me

joshuef:
Merged, thanks @iancoleman !

PR #1340, opened 2021-01-21
chore(refactor): Remove some locks and simplify connection management

joshuef:
- remove listener locks
- remove connect_to_elder func to simplify retries there.

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -11,7 +11,7 @@ use bincode::{deserialize, serialize};
 use bytes::Bytes;
 use futures::{
     future::{join_all, select_all},
-    lock::{self, Mutex},
+    lock::Mutex,
 };
 use log::{debug, error, info, trace, warn};
 use qp2p::{
@@ -231,7 +231,14 @@ impl ConnectionManager {
 
             let endpoint = self.endpoint.clone().ok_or(Error::NotBootstrapped)?;
             let endpoint = endpoint.lock().await;
-            let (connection, _) = endpoint.connect_to(&elder.socket_addr).await?;
+            let (connection, incoming) = endpoint.connect_to(&elder.socket_addr).await?;
+
+            if let Some(_) = incoming {
+                warn!(
+                    "No listener established for elder connection {:?}",
+                    elder.socket_addr
+                );
+            }
 
             let task_handle = tokio::spawn(async move {
                 // Retry queries that failed for connection issues
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -15,10 +15,9 @@ use futures::{
 };
 use log::{debug, error, info, trace, warn};
 use qp2p::{
-    self, Config as QuicP2pConfig, Connection, Endpoint, IncomingMessages, Message as Qp2pMessage,
-    QuicP2p,
+    self, Config as QuicP2pConfig, Endpoint, IncomingMessages, Message as Qp2pMessage, QuicP2p,
 };
-use sn_data_types::{HandshakeRequest, HandshakeResponse, Keypair, PublicKey, TransferValidated};
+use sn_data_types::{HandshakeRequest, HandshakeResponse, Keypair, TransferValidated};
 use sn_messaging::{Event, Message, MessageId, MsgEnvelope, MsgSender, QueryResponse};
 use std::{collections::HashMap, net::SocketAddr, sync::Arc};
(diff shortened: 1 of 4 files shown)


src/connection_manager.rs
```diff
@@ -211,13 +223,15 @@ impl ConnectionManager {
...
             let socket_addr = elder.socket_addr;
-            // let endpoint = self.endpoint.clone();
             // Create a new stream here to not have to worry about filtering replies
             let msg_id = msg.id();
-            let connection = elder.connection.clone();
 
             let pending_query_responses = self.pending_query_responses.clone();
 
+            let endpoint = self.endpoint.lock().await;
+            let (connection, _) = endpoint.connect_to(&elder.socket_addr).await?;
```
lionel1704, line 232: Looks like we expect this to return the _existing_ connection from the connection pool in qp2p.
 Can we enforce it somehow? Eg. check if the incoming_messages returned here is `None`
 
 Or just add a comment saying that this is the existing connection.

src/connection_manager.rs
```diff
@@ -59,14 +56,15 @@ pub struct ConnectionManager {
...
+    pub async fn new(
         mut config: QuicP2pConfig,
         keypair: Arc<Keypair>,
         notification_sender: UnboundedSender<Error>,
     ) -> Result<Self, Error> {
         config.port = Some(0); // Make sure we always use a random port for client connections.
         let qp2p = QuicP2p::with_config(Some(config), Default::default(), false)?;
-        let endpoint = qp2p.new_endpoint()?;
+        let (endpoint, conn, _) = qp2p.bootstrap().await?;
+        conn.close();
```
lionel1704, line 67: Why are we bootstrapping and then closing the connection here?

src/connection_manager.rs
Yoga07: >Can we enforce it somehow?

Should we though? It should be okay(maybe even ideal) for the client to re-connect right? (and update `IncomingConnections` accordingly if it's a new connection)

src/connection_manager.rs
Yoga07: Same Q.

src/connection_manager.rs
lionel1704: Yea that's the follow up :sweat_smile: 
If it's a new connection then we just have to make sure to start listening on the new incoming_messages...

src/connection_manager.rs
```diff
@@ -168,24 +175,29 @@ impl ConnectionManager {
...
-            .lock()
-            .await
-            .insert(msg_id, sender);
+        {
+            let _ = self
+                .pending_transfer_validations
+                .lock()
+                .await
+                .insert(msg_id, sender);
+        }
```
bochaco, line 184: I'm not sure we need the brackets, isn't it the case that the lock is dropped at the end of the whole statement since it's creating a temporary object which is therefore dropped at the end of the statement?

src/connection_manager.rs
joshuef: Aye ill clean that up

src/connection_manager.rs
```diff
@@ -226,7 +240,6 @@ impl ConnectionManager {
 
                 while !done_trying {
                     let msg_bytes_clone = msg_bytes_clone.clone();
-                    let connection = connection.lock().await;
 
                     let (sender, mut receiver) = channel::<Result<QueryResponse, Error>>(7);
                     {
```
bochaco, line 245: Same doubt about the need of the brackets/scope here.

src/connection_manager.rs
```diff
@@ -109,12 +109,19 @@ impl ConnectionManager {
         let mut tasks = Vec::default();
         for elder in &self.elders {
             let msg_bytes_clone = msg_bytes.clone();
-            let connection = Arc::clone(&elder.connection);
+            let (connection, incoming) = endpoint.connect_to(&elder.socket_addr).await?;
+
+            if let Some(incoming_messages) = incoming {
+                let socket_addr = elder.socket_addr;
+                error!("no listener exists for elder: {:?}", socket_addr);
```
bochaco, line 116: This means we are reconnecting to the Elder, presumably we lost a previous connection with it, and if that's correct our listener task should be already dead, so perhaps instead of logging an error we can actually do `elder.listener = self.listen_to_incoming_messages(incoming_messages, socket_addr).await?;`

Yoga07 (approved):

PR #1351, opened 2021-01-29
Rewards flow

oetyng:
Updates for reward flow changes across sn_* repos.

### DIFF
--- a/src/client/sequence_apis.rs
+++ b/src/client/sequence_apis.rs
@@ -1025,7 +1025,7 @@ pub mod exported_tests {
         while balance_before_delete == Money::from_str("0")?
             || balance_before_delete == Money::from_str("10")?
         {
-            delay_for(Duration::from_millis(500)).await;
+            delay_for(Duration::from_millis(200)).await;
             balance_before_delete = client.get_balance().await?;
         }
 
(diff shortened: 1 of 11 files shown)


src/crypto.rs
```diff
@@ -106,7 +106,7 @@ pub mod shared_box {
     }
 
     /// Generate new random public/secret keypair.
-    pub fn gen_bls_keypair() -> (threshold_crypto::SecretKey, threshold_crypto::PublicKey) {
+    pub fn gen_ed_keypair() -> (threshold_crypto::SecretKey, threshold_crypto::PublicKey) {
```
Yoga07, line 109: It is still returning a BLS key pair.

src/client/sequence_apis.rs
```diff
@@ -1111,6 +1118,8 @@ pub mod exported_tests {
         let mut seq_res = client.get_sequence(address).await;
 
         while seq_res.is_err() {
+            delay_for(Duration::from_millis(500)).await;
+            delay_for(Duration::from_millis(500)).await;
```
Yoga07, line 1122: Could just be a 1 sec delay? Applies for other similar places too

joshuef: aye that's a mistake

joshuef (approved):

PR #1360, opened 2021-02-04
feat: remove logging implementation

b-zee:
Libraries should not setup a logging implementation.

This should be left to the applications or binaries that use the
library.

This will make sure applications do not get unexpected output on stdout
or stderr.


b-zee:
I've encountered the same issue with some testing myself. I can hardly see any side effects, looking at the code changes, though. Really curious to see if someone else can spot something.

dirvine:
Scanning the changes I don't see how this fails a test. I Will look deeper later. Perhaps @bochaco  might lend a hand. I know he's busy though

bochaco:
Ah no, don't worry about that test failure right now, it's because we still have the sn-node issues since very recent changes with rewards.

bochaco (approved):
We'll need to have tests to initialise the logger as it's not done at the moment.

PR #1361, opened 2021-02-06
feat(connMgr): read multiple contacts and not just one from node_connection_info.config

bochaco:


lionel1704:
We should make sure that sn_node is also changed to store the configuration as a list instead of a single socket adress.

bochaco:
Yes, but I'll actually go for the agreed change to remove all this logic of reading contacts list from sn_client and only receive it as an optional arg, as well as receiving the path for the qp2p config file.

PR #1369, opened 2021-02-09
feat(config): read config file from an optionally provided path

bochaco:


src/utils/test_utils/tokens.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Generates a random BLS secret and public keypair.
+pub fn gen_ed_keypair() -> Keypair {
+    let mut rng = rand::thread_rng();
+    Keypair::new_ed25519(&mut rng)
+}
+
+/// Helper function to calculate the total cost of expenditure by adding number of mutations and
+/// amount of transferred coins if any.
+pub fn calculate_new_balance(
```
Yoga07, line 21: Think we can remove this function as `COST_OF_PUT` is obsolete now and `StoreCost` needs to be calculated dynamically.

Yoga07 (approved):

Yoga07 (approved):

PR #1375, opened 2021-02-12
Updates for new messaging w/ infra update messages

joshuef:
- listens for both client messages and infrastrucure in listener
- simplify bootstrap so we can call that without recursion (while receiving redirects eg)
- handle infrastructure messaging


src/client/mod.rs
```diff
@@ -99,7 +98,7 @@ impl Client {
     /// ```
     pub async fn new(
         optional_keypair: Option<Keypair>,
-        config_file_path: Option<&Path>,
+        config_file_path: Option<&'static Path>,
```
bochaco, line 101: Why do we need `'static` ?

src/connection_manager.rs
```diff
@@ -56,18 +65,22 @@ pub struct ConnectionManager {
     pending_transfer_validations: Arc<Mutex<HashMap<MessageId, TransferValidationSender>>>,
     pending_query_responses: Arc<Mutex<HashMap<(SocketAddr, MessageId), QueryResponseSender>>>,
     notification_sender: UnboundedSender<Error>,
+    // receive the pk set when calling bootstrap func
+    key_set_sender: Option<Sender<Result<ReplicaPublicKeySet, Error>>>,
+    config_file_path: Option<&'static Path>,
```
bochaco, line 70: I think `'static` will be problematic/non-thread-safe, we should be able to remove that and simply keep an instance of `PathBuf` here.

src/connection_manager.rs
```diff
@@ -597,97 +602,189 @@ impl ConnectionManager {
...
-                                    .await
-                                    .get_mut(&correlation_id)
-                                {
-                                    debug!("Cmd Error was received, sending on channel to caller");
-                                    let _ = sender.send(Err(Error::from(error.clone()))).await;
-                                } else {
-                                    warn!("No sender subscribing and listening for errors relating to message {:?}. Error returned is: {:?}", correlation_id, error)
-                                }
+                let message_type = WireMsg::deserialize(message)?;
+                warn!("Message received at listener from {:?}", &src);
```
bochaco, line 614: I think this log should be `info`/`debug` level instead of `warn`, to me a warning is when it's not an error and it's not causing any problems, but we want to signal that it could bring issues/side-effects to the normal flow/functioning of the lib/crate.

src/connection_manager.rs
```diff
@@ -597,97 +602,189 @@ impl ConnectionManager {
...
+                self.handle_infrastructure_info_update(error).await
+            }
+            InfrastructureMessage::GetSectionResponse(GetSectionResponse::Redirect(
+                addresses,
+            )) => {
+                trace!("GetSectionResponse::Redirect, trying with provided elders");
+                let config = Config::new(self.config_file_path, Some(addresses.iter().cloned().collect())).qp2p;
+
+                // Continually try and bootstrap against new elders while we're getting rediret
+                let _ = self.bootstrap_and_handshake(config).await?;
```
bochaco, line 648: Isn't this creating a recursivity? if so, perhaps to remove it we can send the new addresses back on the `key_set_sender` so the `bootstrap` function does the loop to retry N times??

src/connection_manager.rs
```diff
@@ -597,97 +602,189 @@ impl ConnectionManager {
...
+                .send(Ok(pk_set))
+                .await
+                .map_err(|_| Error::CouldNotSaveReplicaPkSet)?;
+            // let's wipe that
+            self.key_set_sender = None;
+        }
+
+        // clear existing elder lsit.
+        self.elders = Default::default();
+        self.connect_to_elders(elders_addrs).await
```
bochaco, line 717: What if we send the elders addresses back on `key_set_sender` to the `bootstrap` function so it can connect to elders, I'm just thinking if that'd make the flow more easy to follow since the bootstrap would do the retries if it has been redirected to other section, and at the then of that loop we'll see this call to connect to the elders once it successfully bootstrapped to a section.....?.....

src/client/mod.rs
joshuef: atm, cos we're saving this on the struct. I was thinking we may not need that. I think the whole qp2p reinstantiation is bunk, and I just need to manually config to send off infra queries after the inital qp2p

src/connection_manager.rs
joshuef: it's not at the mo. That just triggers sending of a message, which will be received.

It's repeated (will continue until we get to our section), but that's necessary afaik. (As it could be n+1... times... we don't know...). 

IMO this is different to "can we bootstrap / connect to _something_" vs, "connect to our section". Which could take longer (and may warrant a timeout at some point?)

src/connection_manager.rs
joshuef: Hmmmm. Right. Coould be. But we need to be able to do this outwith of that bootstrap func, on split etc.

PR #1377, opened 2021-02-15
Recursively delete all blobs

kanav99:
Fixes https://github.com/maidsafe/self_encryption/issues/252

Deletes all the blobs until the root blob is found. Previously only the outermost blob was deleted.


lionel1704 (changes requested):
Hi @kanav99 I think you are missing some parts of this. Let me try to elaborate what is happening.

1. Given a blob address, we first fetch the raw data.
2. This raw data is definitely a data map.
3. Once this raw data is fetched, this blob has to be deleted.
4. Deserialize the raw data into `DataMap`.
5. 
  ### Case 1:
  - `DataMap::Child`
  -  Read raw data using `read_using_data_map`
  - We need to add a `delete_using_data_map` function that will use the data map and delete all the chunks listed in the map.
  - This raw data is again definitely a `DataMap`
  - Repeat from step 3
  
  ### Case 2:
  - `DataMap::Root`
  - This means that after calling `read_using_data_map` we would get the raw data that was sent by the user.
  - We don't need to fetch the raw data, so we can just call `delete_using_data_map` and delete all the chunks from the network.
  
 Please let me know if you have any questions

src/client/blob_storage.rs
```diff
@@ -57,6 +57,34 @@ impl Storage for BlobStorage {
...
+        }
+
+        let name = {
+            let mut temp = [0_u8; XOR_NAME_LEN];
+            temp.clone_from_slice(name);
+            XorName(temp)
+        };
+
+        let address = if self.published {
+            // Should raise error though
```
lionel1704, line 76: This comment is right. An error can be returned here.

kanav99: Done!

lionel1704 (dismissed):
Great work! :tada:

lionel1704 (approved):

PR #1378, opened 2021-02-15
Update to the latest qp2p API

lionel1704:

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -14,26 +14,27 @@ use futures::{
     lock::Mutex,
 };
 use log::{debug, error, info, trace, warn};
-use qp2p::{
-    self, Config as QuicP2pConfig, Endpoint, IncomingMessages, Message as Qp2pMessage, QuicP2p,
-};
+use qp2p::{self, Config as QuicP2pConfig, Endpoint, IncomingMessages, QuicP2p};
 use sn_data_types::{HandshakeRequest, Keypair, TransferValidated};
 use sn_messaging::{
     client::{Event, Message, MessageId, MsgEnvelope, MsgSender, QueryResponse},
     infrastructure::{GetSectionResponse, Query},
 };
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{HashMap, HashSet},
     net::SocketAddr,
     sync::Arc,
+    time::Duration,
 };
 use tokio::{
     sync::mpsc::{channel, Sender, UnboundedSender},
     task::JoinHandle,
+    time::timeout,
 };
 use xor_name::XorName;
 
 static NUMBER_OF_RETRIES: usize = 3;
+static RESPONSE_WAIT_TIME: u64 = 30;
 pub static STANDARD_ELDERS_COUNT: usize = 5;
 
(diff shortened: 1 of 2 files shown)


joshuef (approved):

src/connection_manager.rs
```diff
@@ -43,17 +41,15 @@ type VoteMap = HashMap<[u8; 32], (QueryResponse, usize)>;
 type TransferValidationSender = Sender<Result<TransferValidated, Error>>;
 type QueryResponseSender = Sender<Result<QueryResponse, Error>>;
 
-type ElderConnectionMap = BTreeMap<SocketAddr, Arc<NetworkListenerHandle>>;
+type ElderConnectionMap = BTreeSet<SocketAddr>;
```
bochaco, line 44: Not a big deal I think, but since we are using `HashSet` in this and other crates for the bootstrapping/hard-coded contacts list, perhaps also use it for this here?

lionel1704: Sure. I can make that change. I'm going to do a rebase now, will include this change as well.

src/connection_manager.rs
```diff
@@ -618,106 +596,97 @@ impl ConnectionManager {
...
                             }
                         }
                     }
+                    Err(_error) => {
+                        error!("Could not deserialize MessageEnvelope");
+                    }
                 }
             }
-
+            error!("IncomingMessages listener is closing now");
```
bochaco, line 685: I don't think this should be an `error` level log but a `warn`? otherwise we wouldn't be returning `Ok(())`

src/connection_manager.rs
```diff
@@ -618,106 +596,97 @@ impl ConnectionManager {
...
+
+                                let _ = notifier.send(Error::from(error));
+                            }
+                            msg => {
+                                warn!("another message type received {:?}", msg);
                             }
                         }
                     }
+                    Err(_error) => {
+                        error!("Could not deserialize MessageEnvelope");
```
bochaco, line 681: Same comment about `error` level log as below, it doesn't seem an error but just a scenario being handled which we my just wanna warn about with `warn` or simply `info`

src/connection_manager.rs
```diff
@@ -463,53 +449,52 @@ impl ConnectionManager {
...
+        endpoint.send_message(msg, &bootstrapped_peer).await?;
+
+        if let Some((_src, message)) = incoming_messages.next().await {
+            match Query::from(message) {
+                Ok(Query::GetSectionResponse(GetSectionResponse::Redirect(
+                    addresses,
+                ))) => {
+                    trace!("GetSectionResponse::Redirect, trying again");
+                    // TODO: initialise `hard_coded_contacts` with received `elders`.
+                    Err(Error::UnexpectedMessageOnJoin(format!("Client should re-bootstrap with a new set of Elders, but it's not yet supported: {:?}",addresses)))
```
bochaco, line 476: Note this has been fixed/implemented recently, so this probably was missed when rebasing, this was implemented with:
```
 trace!("GetSectionResponse::Redirect, trying with provided elders");
 Ok(addresses)
```

bochaco (changes requested):

joshuef (approved):

PR #1387, opened 2021-02-22
Refactored state handling

oetyng:
Refactors state handling of connection_manager.rs.
Updated sn_messaging and client bootstrap flows.


src/connection_manager.rs
```diff
@@ -347,6 +347,7 @@ impl ConnectionManager {
                     received_errors,
                     &mut has_elected_a_response,
                 );
+                has_elected_a_response = true;
```
joshuef, line 351: this is not necessary.

It's handled by the above func:https://github.com/maidsafe/sn_client/blob/7a133d66947837265f6eb9fefedb82c9f77d04f2/src/connection_manager.rs#L415-L418

joshuef (changes requested):

src/connection_manager.rs
joshuef:
```suggestion
               
```

joshuef (approved):

PR #1390, opened 2021-02-24
chore(tokio): upgrade qp2p to v0.10.1 and tokio to v1.3.0

bochaco:


lionel1704:
I've updated this PR with a commit from #1397.
I believe that it supersedes #1395 as well

dirvine (approved):

PR #1400, opened 2021-03-09
Fix force wait for keys

### LINKED ISSUE #1404: Client does not time out when looking for network on its own port
If the client binds its own qp2p port to the default (12000) and it's looking for a network on the default port (12000), it seems like the client does not time out while looking for the network (bootstrapping).

I've tried this with the `QuicP2pConfig.idle_timeout_msec` set to 5000 (5 seconds). Providing a hard coded contact with a different port (e.g. 12001), the client 'successfully' times out.

joshuef:
Resolves #1404

### DIFF
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -82,11 +82,7 @@ impl Client {
         let msg_contents = Query::Transfer(TransferQuery::GetBalance(public_key));
 
         let message = self.create_query_message(msg_contents).await?;
-        let endpoint = self.session.endpoint()?.clone();
-        let elders = self.session.elders.iter().cloned().collect();
-        let pending_queries = self.session.pending_queries.clone();
-        let res =
-            ConnectionManager::send_query(&message, endpoint, elders, pending_queries).await?;
+        let res = ConnectionManager::send_query(&message, &self.session).await?;
 
         match res {
             QueryResponse::GetBalance(balance) => balance.map_err(Error::from),
@@ -186,9 +182,7 @@ impl Client {
             transfer_proof
         );
 
-        let endpoint = self.session.endpoint()?.clone();
-        let elders = self.session.elders.iter().cloned().collect();
-        let _ = ConnectionManager::send_cmd(&message, endpoint, elders).await?;
+        let _ = ConnectionManager::send_cmd(&message, &self.session).await?;
 
         let mut actor = self.transfer_actor.lock().await;
         // First register with local actor, then reply.
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -81,7 +81,7 @@ impl Client {
 
         let msg_contents = Query::Transfer(TransferQuery::GetBalance(public_key));
 
-        let message = self.create_query_message(msg_contents)?;
+        let message = self.create_query_message(msg_contents).await?;
         let endpoint = self.session.endpoint()?.clone();
         let elders = self.session.elders.iter().cloned().collect();
         let pending_queries = self.session.pending_queries.clone();
@@ -140,17 +140,21 @@ impl Client {
         // first make sure our balance  history is up to date
         self.get_history().await?;
 
-        info!(
-            "Our actor balance at send: {:?}",
-            self.transfer_actor.lock().await.balance()
-        );
-
-        let initiated = self
-            .transfer_actor
-            .lock()
-            .await
-            .transfer(amount, to, "asdf".to_string())?
-            .ok_or(Error::NoTransferGenerated)?;
+        {
+            info!(
+                "Our actor balance at send: {:?}",
+                self.transfer_actor.lock().await.balance()
+            );
+        }
+        let initiated;
+        {
+            initiated = self
+                .transfer_actor
+                .lock()
+                .await
+                .transfer(amount, to, "asdf".to_string())?
+                .ok_or(Error::NoTransferGenerated)?;
+        }
 
         let signed_transfer = SignedTransfer {
             debit: initiated.signed_debit,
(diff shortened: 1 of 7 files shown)


src/connection_manager.rs
```diff
@@ -160,19 +173,33 @@ impl ConnectionManager {
     pub async fn send_transfer_validation(
         msg: &Message,
         sender: Sender<Result<TransferValidated, Error>>,
-        endpoint: Endpoint,
-        elders: Vec<SocketAddr>,
-        pending_transfers: PendingTransferValidations,
+        session: &Session,
+        // endpoint: Endpoint,
+        // elders: Vec<SocketAddr>,
+        // pending_transfers: PendingTransferValidations,
```
Yoga07, line 179: Can remove these.

Yoga07 (dismissed):

Cargo.toml
```diff
@@ -29,7 +29,7 @@ rmp-serde = "~0.15.4"
 self_encryption = "~0.20.0"
 serde_json = "1.0.62"
 sn_data_types = "~0.15.0"
-sn_messaging = "~6.0.0"
+sn_messaging = { git = "https://github.com/oetyng/sn_messaging", branch = "split-section-funds" }
```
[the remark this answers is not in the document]
joshuef: ahhh, aye right. _this_ was the blocker.

src/client/transfer_actor/balance_management.rs
```diff
@@ -140,17 +136,21 @@ impl Client {
...
-            .await
-            .transfer(amount, to, "asdf".to_string())?
-            .ok_or(Error::NoTransferGenerated)?;
+        {
+            info!(
+                "Our actor balance at send: {:?}",
+                self.transfer_actor.lock().await.balance()
+            );
+        }
+        let initiated;
```
lionel1704, line 145: Can't this be
```suggestion
        let initiated = { ... };
```

src/connection_manager.rs
```diff
@@ -57,47 +57,61 @@ pub struct ConnectionManager {}
...
-                Err(e) => Err(e.into()),
-            }
-        } else {
-            trace!("incoming_messages.next().await was None");
-            Err(Error::NotBootstrapped)
+        // bootstrap is not complete until we have pk set...
+        while !we_have_keyset {
+            use tokio::time::{delay_for, Duration};
+            delay_for(Duration::from_millis(500)).await;
+            we_have_keyset = session.section_key_set.lock().await.is_some();
```
lionel1704, line 91: I still think we can prevent this looping here

src/connection_manager.rs
```diff
@@ -57,47 +57,61 @@ pub struct ConnectionManager {}
...
+        // elders: Vec<SocketAddr>,
     ) -> Result<(), Error> {
         let msg_id = msg.id();
+        let endpoint = session.endpoint()?.clone();
+
+        let elders: Vec<SocketAddr>;
+
+        {
+            elders = session.elders.lock().await.iter().cloned().collect();
+        }
```
lionel1704, line 111:
```suggestion
        let elders = {
             session.elders.lock().await.iter().cloned().collect();
        }
```
Perhaps even without the braces. It should remain the same...

src/connection_manager.rs
```diff
@@ -160,19 +173,33 @@ impl ConnectionManager {
...
             "Sending transfer validation command {:?} w/ id: {:?}",
             msg,
             msg.id()
         );
+        let endpoint = session.endpoint()?.clone();
+        let elders: Vec<SocketAddr>;
+
+        {
+            elders = session.elders.lock().await.iter().cloned().collect();
+        }
```
lionel1704, line 191: Here too

src/connection_manager.rs
```diff
@@ -201,12 +228,16 @@ impl ConnectionManager {
...
-        elders: Vec<SocketAddr>,
-        pending_queries: PendingQueryResponses,
-    ) -> Result<QueryResponse, Error> {
+    pub async fn send_query(msg: &Message, session: &Session) -> Result<QueryResponse, Error> {
+        let endpoint = session.endpoint()?.clone();
+        let elders: Vec<SocketAddr>;
+
+        {
+            elders = session.elders.lock().await.iter().cloned().collect();
+        }
```
lionel1704, line 237: One more.

src/connection_manager.rs
```diff
@@ -422,44 +453,65 @@ impl ConnectionManager {
...
+    ) -> Result<Session, Error> {
+        if session.is_connecting_to_new_elders {
+            debug!("Already attempting elder connections, dropping get_section call until that is complete.");
+            return Ok(session)
+        }
+        let elders: Vec<SocketAddr>;
+        
+        {
+            elders = session.elders.lock().await.iter().cloned().collect();
+        }
```
lionel1704, line 470: One more

src/connection_manager.rs
```diff
@@ -422,44 +453,65 @@ impl ConnectionManager {
...
         let mut tasks = Vec::default();
 
         let endpoint = session.endpoint()?;
         let msg = session.bootstrap_cmd().await?;
-        let peers = session.elders.clone();
+
+        let peers;
+        {
+            peers = session.elders.lock().await.clone();
+        }
```
lionel1704, line 513: Here too

src/connection_manager.rs
```diff
@@ -592,18 +667,44 @@ impl ConnectionManager {
         mut session: Session,
         info: &SectionInfo,
     ) -> Result<Session, Error> {
+        let original_elders ;
+
+        {
+            original_elders = session.elders.lock().await.clone();
+        }
```
lionel1704, line 674: Here too

src/client/transfer_actor/balance_management.rs
joshuef: why though?

src/connection_manager.rs
joshuef: We need to, as we cannot guarantee a first repsonse is for our section. So we want to hit the standard looping code.

We dont _have_ to have a delay here... but seems better than just hammering things.

lionel1704 (dismissed):

lionel1704 (approved):

PR #1409, opened 2021-03-18
chore: dep updates and changes for split

joshuef:

### DIFF
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -363,9 +363,7 @@ mod tests {
         let mut balance = client.get_balance().await?;
 
         while balance != Token::from_str("110")? {
-
             println!("FOUND BALANCE: {:?}", balance);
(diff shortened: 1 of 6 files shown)


src/client/transfer_actor/balance_management.rs
```diff
@@ -363,29 +363,44 @@ mod tests {
         let mut balance = client.get_balance().await?;
 
         while balance != Token::from_str("110")? {
+
+            println!("FOUND BALANCE: {:?}", balance);
```
lionel1704, line 367: Probably best to remove `println!`s and use log::* instead?

joshuef: aye, that should no be there.

lionel1704 (dismissed):
Just one small comment about `println!`

lionel1704 (dismissed):

lionel1704 (approved):

PR #1429, opened 2021-04-02
fix: create listeners before sending queries

joshuef:


src/connection_manager.rs
```diff
@@ -209,6 +209,7 @@ impl Session {
         let endpoint = self.endpoint()?.clone();
         let elders: Vec<SocketAddr> = self.connected_elders.lock().await.keys().cloned().collect();
 
+        let msg = msg.clone();
```
Yoga07, line 212: Is this clone required?

joshuef: no!

Yoga07 (approved):

PR #1432, opened 2021-04-04
Feat resend recent messages on infra updates

joshuef:
- adds small cache for sending messages (currently set to be 100 messages).
- add to cache on send
- remove oldest if we're over the limit
- re send messags on infra update messages received in response to an outgoing msg (after updating our session)


S-Coyle:
Closing as repo is being archived - to be raised against https://github.com/maidsafe/safe_network if still required.

PR #1445, opened 2021-04-08
chore: fix node configuration information file path

Dylan-DPC:
the relative path of the configuration file should be a path instead of a string so that it works on platforms like Windows which uses `\`  for separation. Creating a `Path` before joining it, solves the cross-platform issue.


joshuef:
I'm wondering if this will sort the brittle test and re can un-ingore: https://github.com/maidsafe/sn_client/blob/master/src/config_handler.rs#L98

lionel1704 (approved):

PR #1446, opened 2021-04-12
chore(log): log at INFO level the store cost reported by the network …

bochaco:
…before a write operation is attempted

### DIFF
--- a/src/client/transfer_actor/mod.rs
+++ b/src/client/transfer_actor/mod.rs
@@ -198,12 +198,16 @@ impl Client {
         let section_key = self.session.section_key().await?;
 
         let cost_of_put = self.get_store_cost(bytes).await?;
+        info!(
+            "Current store cost reported by the network: {}",
+            cost_of_put
+        );
 
         let initiated = self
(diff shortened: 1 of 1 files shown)


Yoga07 (dismissed):

Yoga07 (approved):

src/client/transfer_actor/mod.rs
```diff
@@ -198,12 +198,16 @@ impl Client {
         let section_key = self.session.section_key().await?;
 
         let cost_of_put = self.get_store_cost(bytes).await?;
```
happybeing, line 200: Can we have something equivalent in the node log? 

If so it would be good to have both the bytes and the cost in a single log message, so `vdash` can monitor and display a chart of the rate. For example:

```
info!(
   "Current store cost for section: {} for {} bytes",
   cost_of_put, bytes
);
```


Also, is `cost_of_put

oetyng (commented):
Yes I agree here. The store cost requires bytes as well to be helpful.

PR #1447, opened 2021-04-12
Rebootstrap with SectionInfo::Redirect instead of connecting to all peers

lionel1704:
When SectionInfo::Redirect is called we can re-boostrap instead of connecting to all the provided peers and sending the bootstrap request.
In the testnet for a client redirect was sent 8 times which meant 8 x 7 connections which were eventually just dropped.

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -398,10 +398,10 @@ impl Session {
     // Otherwise we use all session nodes
     async fn send_get_section_query(&self, initial_peer: Option<SocketAddr>) -> Result<(), Error> {
         if self.is_connecting_to_new_elders {
-            debug!("Already attempting elder connections, not sending section query until that is complete.");
+            // This should ideally be unreachable code. Leaving it while this is a WIP
+            error!("Already attempting elder connections, not sending section query until that is complete.");
             return Ok(());
         }
-        let elders: Vec<SocketAddr> = self.all_known_elders.lock().await.keys().cloned().collect();
 
         // 1. We query the network for section info.
         trace!("Querying for section info from bootstrapped node...");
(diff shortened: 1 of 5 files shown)


S-Coyle (dismissed):

src/connection_manager.rs
```diff
@@ -583,24 +585,13 @@ impl Session {
...
-
-                    *known_elders = temp_elders;
-                }
+                trace!("GetSectionResponse::Redirect, reboostrapping with provided peers");
                 // Disconnect from peer that sent us the redirect, connect to the new elders provided and
                 // request the section info again.
                 self.disconnect_from_peers(vec![src])?;
-                self.connect_to_elders().await?;
-                self.send_get_section_query(None).await?;
+                let endpoint = self.endpoint()?.clone();
```
bochaco, line 592: You probably don't need to clone it as all you need is a ref ?

src/connection_manager.rs
```diff
@@ -667,7 +658,8 @@ impl Session {
 
         if original_known_elders != received_elders {
             debug!("Connecting to new elders");
-            let new_elder_addresses = received_elders.keys().collect::<BTreeSet<_>>();
+            let new_elder_addresses = received_elders.keys().cloned().collect::<BTreeSet<_>>();
+            let updated_contacts = new_elder_addresses.iter().cloned().collect::<Vec<_>>();
```
bochaco, line 662: Can't you collect directly from `received_elders.keys()` ?

src/connection_manager.rs
lionel1704: The next statement requires qp2p to be `&mut` and since `self.endpoint()` is an immutable borrow, we cannot borrow mutably until the reference is dropped.

src/connection_manager.rs
lionel1704: We can. `new_elder_addresses` was already there and used elsewhere, hence I just reused it.

Yoga07 (approved):

PR #1473, opened 2021-04-28
refactor(connMgr): send query messages to a subset of 3 Elders, and simply accept the very first response

bochaco:


dirvine (approved):

joshuef:
Are we not opening up clients to receive potentially bad info here... accepting first of only 3? (and not best of majority?). (I also think regardless of 3 accepting first only could make the client more likely to get out of date info, no?)

(I know this is all a trade off...)

dirvine:
As it's a query then it's cool. For ease let's use numbers. So group == 7 supermajority == 5. For queries and forwarding messages, we want to make sure we have at least 1 honest node. So we select 3. This means in worst case 2 are bad and do try to give bad (i.e. outdated mutable) but we know 1 is honest. As data is signed etc. we just take the most current one.

joshuef:
But here we're just accepting whatever we get first.
Aha, I see @bochaco is all over what I'm saying actually: https://github.com/maidsafe/sn_client/pull/1473/files#diff-6557813f6025ab0361832490790e354a3e68b8eb595d81f7785c5cd60ae63028R352-R361

Aye, i think 'most current' doesn't necessarily mean it's valid. So we'd need to re-impl a vote-map or something similar at some point.

Neat we can do that w/ chunks though. I was not aware of that.

PR #1507, opened 2021-05-17
feat(session): make Sessions agnostic of the keypair used to sign each individual client message

bochaco:
BREAKING CHANGE: removing the EndUser registration step from the boostrapping stage


src/client/mod.rs
```diff
@@ -8,36 +8,32 @@
...
 use std::{
     path::Path,
     str::FromStr,
     {collections::HashSet, net::SocketAddr, sync::Arc},
 };
-use tokio::sync::mpsc::{Receiver, Sender};
+use tokio::sync::mpsc::Receiver;
+
+// Number of attempts to make when trying to bootstrap to the network
+const NUM_OF_BOOTSTRAPPING_ATTEMPTS: u8 = 1;
```
Yoga07, line 36: We can increase the bootstrap time.

Yoga07 (approved):

src/client/mod.rs
bochaco: yeah, good catch! i'll send a PR to fix it, thanks!

PR #1534, opened 2021-06-07
Small improvements to client

joshuef:
- RwLock where possible
- return errors when history not as expected


src/client/transfer_actor/balance_management.rs
```diff
@@ -133,36 +133,14 @@ impl Client {
     ) -> Result<(u64, PublicKey), Error> {
         info!("Sending token");
 
-        // first make sure our balance  history is up to date
+        // first make sure our balance  history is up to date with the network (and vica versa)
         self.get_history().await?;
```
Yoga07, line 137: We can remove this `get_history` here as we'll be calling it once again in the method `get_validation_cmd_and_transfer_from_actor` that is called below.

src/connections/messaging.rs
```diff
@@ -285,7 +304,12 @@ impl Session {
         // We send the same message to all Elders concurrently
         let mut tasks = FuturesUnordered::new();
         let (sender, mut receiver) = channel::<QueryResponse>(7);
-        let _ = pending_queries.lock().await.insert(msg_id, sender);
+
+        let mut pending_queries = pending_queries.write().await;
+        // let _ = pending_queries.write().await.insert(msg_id, sender);
```
Yoga07, line 309: Can remove this

src/client/transfer_actor/mod.rs
```diff
@@ -138,7 +144,32 @@ impl Client {
...
+            let received_debit_count = history.debits.len();
+
+            debug!(
+                "Received debit history len {:?}, our debit history len : {:?}",
+                received_debit_count,
+                our_history.debits.len()
+            );
+
+            for (_i, debit) in our_history.debits.iter().enumerate() {
+                debug!("Updating elders with some known debits");
```
Yoga07, line 165: Not sure if we are doing it correctly here, because both credits and debits are sequential in our AT2 mechanism with the counter ticking for each event. So sending only the debits might error out at the network, as there could've been credits in between these debits and the network would wait until the next counter is seen.

src/connections/listeners.rs
```diff
@@ -229,23 +229,28 @@ impl Session {
...
                 response,
                 correlation_id,
                 ..
             } => {
-                trace!(
-                    "Query response (correlation id: {}): {:?}",
-                    correlation_id,
-                    response
-                );
+                // trace!(
```
Yoga07, line 243: Can uncomment these?

src/client/transfer_actor/mod.rs
joshuef: only debits are sequential. Credits not.

Yoga07 (approved):

src/connections/messaging.rs
```diff
@@ -285,7 +303,11 @@ impl Session {
         // We send the same message to all Elders concurrently
         let mut tasks = FuturesUnordered::new();
         let (sender, mut receiver) = channel::<QueryResponse>(7);
-        let _ = pending_queries.lock().await.insert(msg_id, sender);
+
+        let mut pending_queries = self.pending_queries.write().await;
+        let _ = pending_queries.insert(msg_id, sender);
+
+        drop(pending_queries);
```
bochaco, line 310: I think if you do it all in one statement as it's done in previous code, the lock is dropped right after the statement finished, which is what you are trying to do

joshuef: aye, you're right. This is leftover from some debugging; there's no difference at all there. Will change that uppp 👍

joshuef:
Closing. I think there may be a better/simpler way forward

PR #1535, opened 2021-06-08
refactor(elder-comms): choose random elders when sending queries

lionel1704:
BREAKING CHANGE: this updates the sn_messaging dependency to v0.31.0


dirvine:
This will cause amplification of messages over hops which risks swamping the network in bad ways that won't be seen in early tests. We should not do this.

joshuef:
@dirvine I think the intention was that this is coupled with a deterministic (msgid based) forwarding at nodes: https://github.com/maidsafe/sn_node/pull/1557/files#diff-f430193dc828f57746e99607374f8d0db1664a704ab972172c2812312b6fa954R164-R167

So any random elders the client chooses, will always forward on to the same three elders in the dst section.

So i dont think there's any amplification there.

(but clients could also send to _any_ node anyway, so I dont _think_ this necessarily breaks anything) ?

lionel1704:
> Caching has a chance of working
Valid point about caching. (We don't have a cache mechanism in place at the moment)

As Josh mentioned, we wouldn't be amplifying the messages since each of the three nodes will use the message ID generated at the client to choose 3 elders to forward the request to.

The randomness at the client section helps us in this situation:
When the client section matches the prefix for the requested data, if we choose the same elders for a piece of data and one(or more) of the adults are not contactable then only those Elders will propose that node as offline. To reach a super-majority we have to either:
- Send a swarm message to all Elders to check and vote on the Adult's reach-ability status
- Choose random Elders so when the request is repeated it will choose a different set of Elders which will help us reach super-majority.

dirvine:
>Send a swarm message to all Elders to check and vote on the Adult's reach-ability status
This is the way, accuse a node of being uncontactable and get it kicked out. The random approach is very flakey IMO and relies on a lottery type approach. We need to find uncontactable nodes and kill them, then on top we have responsiveness checks which are also vital.

I would say here we are decisive and act immediately on unresponsive or unreachable Adults and do so via consensus of elders. Not by chance, If that makes sense.

PR #1540, opened 2021-06-09
Concurrency updates

Yoga07:

### DIFF
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -37,7 +37,7 @@ impl Client {
     /// ```
     pub async fn get_local_balance(&self) -> Token {
         info!("Retrieving actor's local balance.");
-        self.transfer_actor.lock().await.balance()
+        self.transfer_actor.read().await.balance()
     }
 
     /// Handle a validation event.
@@ -51,7 +51,7 @@ impl Client {
             Event::TransferValidated { event, .. } => event,
             _ => return Err(Error::UnexpectedTransferEvent(event)),
         };
-        let mut actor = self.transfer_actor.lock().await;
+        let mut actor = self.transfer_actor.write().await;
         let transfer_validation = match actor.receive(validation) {
             Ok(Some(validation)) => validation,
             Ok(None) => return Ok(None),
@@ -138,12 +138,12 @@ impl Client {
 
         info!(
             "Our actor balance at send: {:?}",
-            self.transfer_actor.lock().await.balance()
+            self.transfer_actor.read().await.balance()
         );
 
         let initiated = self
             .transfer_actor
-            .lock()
+            .read()
             .await
             .transfer(amount, to, "".to_string())?
             .ok_or(Error::NoTransferGenerated)?;
@@ -156,7 +156,7 @@ impl Client {
         let cmd = Cmd::Transfer(TransferCmd::ValidateTransfer(signed_transfer.clone()));
 
         self.transfer_actor
-            .lock()
+            .write()
             .await
             .apply(ActorEvent::TransferInitiated(TransferInitiated {
                 signed_debit: signed_transfer.debit.clone(),
@@ -176,7 +176,7 @@ impl Client {
 
         self.send_cmd(cmd).await?;
 
-        let mut actor = self.transfer_actor.lock().await;
+        let mut actor = self.transfer_actor.write().await;
         // First register with local actor, then reply.
         let register_event = actor
             .register(transfer_proof)?
--- a/src/client/transfer_actor/balance_management.rs
+++ b/src/client/transfer_actor/balance_management.rs
@@ -136,17 +136,18 @@ impl Client {
         // first make sure our balance  history is up to date
         self.get_history().await?;
 
+        let temp_actor = self.transfer_actor.read().await;
         info!(
             "Our actor balance at send: {:?}",
-            self.transfer_actor.read().await.balance()
+            temp_actor.balance()
         );
 
-        let initiated = self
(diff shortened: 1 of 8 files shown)


src/client/transfer_actor/mod.rs
```diff
@@ -283,7 +283,7 @@ impl Client {
                 Some(event) => match event {
                     Ok(transfer_validated) => {
                         response_count += 1;
-                        let mut actor = self.transfer_actor.lock().await;
+                        let mut actor = self.transfer_actor.write().await;
```
bochaco, line 286: Any reason why we don't get write lock in L292 instead of here?

src/connections/messaging.rs
```diff
@@ -45,14 +45,32 @@ impl Session {
 
         self.endpoint = Some(endpoint.clone());
         let mut bootstrap_nodes = endpoint
+            .clone()
             .bootstrap_nodes()
             .to_vec()
             .into_iter()
             .collect::<BTreeSet<_>>();
-        let connected_elders = self.connected_elders.clone();
+        // let connected_elders = self.connected_elders.clone();
```
bochaco, line 53: it seems we can remove this comment

src/client/transfer_actor/balance_management.rs
```diff
@@ -136,32 +136,32 @@ impl Client {
...
-            }))?;
+        let mut actor_w_write_lock = self.transfer_actor.write().await;
+
+        actor_w_write_lock.apply(ActorEvent::TransferInitiated(TransferInitiated {
+            signed_debit: signed_transfer.debit.clone(),
+            signed_credit: signed_transfer.credit.clone(),
+        }))?;
+
+        // drop write lock before we await validation
+        drop(actor_w_write_lock);
```
bochaco, line 164: I think we can do all this in a single statement, I'm wary about having them like this since somebody could easily remove the drop statement causing some issues.

src/client/transfer_actor/mod.rs
bochaco: I mean get read lock here, and only when we find out we need to write we get the write lock in L292

src/client/transfer_actor/balance_management.rs
Yoga07: I was thinking if we could do these within `{ }` blocks. That way they would automatically be dropped once the block is exited, we don't have to drop them manually.

src/connections/messaging.rs
```diff
@@ -285,7 +303,13 @@ impl Session {
...
         let mut tasks = FuturesUnordered::new();
         let (sender, mut receiver) = channel::<QueryResponse>(7);
-        let _ = pending_queries.lock().await.insert(msg_id, sender);
+
+        let pending_queries2 = pending_queries.clone();
+        let _ = tokio::spawn(async move {
+            // Remove the response sender
+            trace!("Removing channel for {:?}", msg_id);
+            let _ = pending_queries2.write().await.insert(msg_id, sender);
+        });
```
bochaco, line 312: Do we really need to spawn a task a single insert op? also the comment is saying it's removing rather than adding

src/connections/messaging.rs
```diff
@@ -402,9 +426,11 @@ impl Session {
             msg_id, response
         );
 
-        // Remove the response sender
-        trace!("Removing channel for {:?}", msg_id);
-        let _ = pending_queries.lock().await.remove(&msg_id);
+        let _ = tokio::spawn(async move {
+            // Remove the response sender
+            trace!("Removing channel for {:?}", msg_id);
+            let _ = pending_queries.clone().write().await.remove(&msg_id);
```
bochaco, line 432: Do we really need to spawn a task a single insert op?

src/client/transfer_actor/mod.rs
joshuef: nope!

src/client/transfer_actor/balance_management.rs
joshuef: Yeh we probably totally could. But I wanted to hard verify this cos things are so hard to debug here. (and i didnt get round to cleanup yet)

src/connections/messaging.rs
joshuef: No, as above

joshuef: seemingly, without this things hang

src/client/transfer_actor/balance_management.rs
```diff
@@ -136,15 +136,10 @@ impl Client {
         // first make sure our balance  history is up to date
         self.get_history().await?;
 
-        info!(
-            "Our actor balance at send: {:?}",
-            self.transfer_actor.lock().await.balance()
-        );
+        let temp_actor = self.transfer_actor.read().await;
```
Yoga07, line 139: Do we not want to release the lock here before trying to write at L#153? i.e. doing this read lock in a separate `{ }` block. Think we might deadlock here 🤔

src/client/transfer_actor/balance_management.rs
```diff
@@ -176,7 +170,7 @@ impl Client {
 
         self.send_cmd(cmd).await?;
 
-        let mut actor = self.transfer_actor.lock().await;
+        let mut actor = self.transfer_actor.write().await;
```
Yoga07, line 173: Can we use the `actor_w_write_lock` directly?

src/client/transfer_actor/balance_management.rs
joshuef: just having them as a one liner statement is fine it seems

S-Coyle (approved):

PR #1554, opened 2021-06-18
fix:  put improvement

joshuef:
- adds large put test (100mb)
- removes get_history from funcs that dont need it; we'll assume one client at a time per PK for now
- increases payment buffer


oetyng (dismissed):

S-Coyle:
Closing as repo is being archived - to be raised against https://github.com/maidsafe/safe_network if still required.

